PR #7, opened 2022-08-06
first pass at determining if jobs have descendents, fix zindex

taylordowns2000:
Hey @stuartc and @amberrignell , I have:
1. fixed the zIndex thing but it feels hacky
2. determined if a job should be an "output" job (i.e., if it has no downstream jobs)
3. identified a possible place that we could then use the data from (2) to add a node for the "plus button"
4. talked to the ReactFlow team and found out that they're HAPPY to have us use all the pro features (including shapes and hiding the attribution) without paying anything. They're really swell.
5. Hid the attribution so it doesn't pop up over our job editor.

@amberrignell , if you want to take a crack at the plus button, please go for it. also have a look at the `TODO`s i've scattered around. this needs a cleanup before merging, but i'd like to merge before Wednesday

Also for consideration... this was created very much as a proof of concept. As we start to make decisions about how things look there is a lot of DRYing out that needs doing in the node and edge generation and formatting.


stuartc (approved):

stuartc:
Thats super exciting @taylordowns2000 about the ReactFlow feedback! Excited to hear what/how we get the pro features here - especially the layout stuff since it now only would make our existing code way simpler - it would cut our bundle size down _significantly_.

stuartc:
@taylordowns2000 the plus button is done for now... did a bit of cleaning up etc. Could you give this a look and merge if you're happy.

PR #10, opened 2022-08-18
Ignore node modules in flatten

josephjclark:
For some reason, when we call [https://unpkg.com/@openfn/language-common@1.7.3/?meta](https://unpkg.com/@openfn/language-common@1.7.3/?meta) for metadata, the resulting `files` list includes type definitions from node modules.

I presume this is something "wrong" with the 1.7.3 package of language-common.

In any case, what this means is that when our shiny new Monaco editor looks for the d.ts for an older module, it pulls down a lot of d.ts files which is a) kinda slow and b) could result in incorrect typings.

While we could work around this in the Lightning repo, it seems sensible to me to add a flag to ignore  node_modules paths when running `flattenFiles`. I've defaulted this to `false`, so that it doesn't break anythhing, but maybe we should default that to `true`.

`fetchDTS` is, so far as I can tell, not used by anyone else, so I've taken the liberty to defaulting that to ignoring node modules.

### DIFF
--- a/packages/compiler/src/package-fs.ts
+++ b/packages/compiler/src/package-fs.ts
@@ -18,21 +18,24 @@ interface PackageListing {
   files?: PackageListing[];
 }
 
-function* flattenFiles(
+export function* flattenFiles(
   listing: PackageListing,
+  ignoreNodeModules = false,
   path: string = ""
 ): Generator<string> {
   if (listing.type == "directory") {
     for (let i = 0; i < listing.files!.length; i++) {
       const f = listing.files![i];
-      yield* flattenFiles(f);
+      if (!ignoreNodeModules || !f.path.startsWith('/node_modules/')) {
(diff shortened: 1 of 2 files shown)


stuartc (approved):
Nice! Yeah our old (or current if we ignore `language-common@v2.0.0-rc2`) build process uses `bundledDependencies` on a lot of the adaptors which results in all those deps being included. Back in the day we wanted modules to be as tarball-able as possible.

From the patterns layed out in the newer `language-common` and `language-http` I want to try and all of our modules to use `esbuild` or at least bundle a sensible amount of deps in a single file - bringing us in line with this decade and still making dependency version drift and conflict as little as a problem as possible.

While I don't think supporting older versions of adaptors is a priority for me, I mean if they do work in some way with Monaco/diagnostics then win! But if they don't :shrug:, we ship new versions with dts files and move forward.

Anyway, ignoring bundled deps is :100: to me! :+1:

PR #11, opened 2022-08-24
Runtime and Compiler v2

josephjclark:
The next generation runtime engine.

To make it so:
```
npm install -g @openfn/cli
openfn --help
openfn --test
```
## Merge Checklist

What do we need to do before merging this to main?

* [ ] Update Lightning to use the new describe-package (instead of `compiler`)
* [x] Check the workflow-diagram tests and un-skip (I need help with this!)
* [x] ~~Port describe-package tests over to ava ?~~ Nah
* [x] Tests passing
* [x] Update the top level readme
* [x] Work out what to do with the top level package.json and examples  folder
* [x] Spin out unfinished items below into new issues (anything spun out has been struckthrough here)

Likely blockers with real-world jobs

* [x] Immutable state (we should make state mutable for now, or at least hide immutability behind a flag)

## Runtime

Create a new runtime which accepts a job as a pipeline of functions (as a string representation of an esm module which exports an array of functions), and runs them through the execute reducer.
```
import { readFile } from 'node:fs/promises';
import run from '@openfn/runtime';

const job = await readFile('expression.js', 'utf8'); // compiled expression .ejs
const initialState = {};
const { data } = await run(source, initialState);
```
The job is loaded into a sandbox environment using experimental node:vm module loaders. This is more to load the source properly and control the execution environment, it's not a security thing. Seems to work great.

The runtime manager may want to run the whole runtime in a vm2 context for added security.

- [x] Load the job ESM code as an actual executable module (and run it)
- [x] Allow environment globals like `console` to be overridden inside the job
- [x] Don't import execute, use a native implementation in this package (even if it's identical)
- [x] Automated Logging and log level controls
- [ ] ~~run should return a promise which is also an event emitter~~
- [ ] ~~Emit events as operations start and return (current status: running operation 4: fetch from salesforce)~~
- [x] State is mutable by default, but can be made immutable if a flag is passed

## Compiler

The new compiler will be quite different from the old because it won't need to do stuff like inject the execute reducer or ensure there's an environment for language adaptors. The runtime handles all that stuff now.

I am still using recast, acorn and ast-types to do the heavy lifting (although tbh I'm not sure why we're favouring acorn over esprima!)

- [x] Set up some patterns for visitors and unit tests
- [x] Create a transformer to ensure a valid `export default []`
- [x] Create a transformer to move top-level call expressions into the default exports
- [x] Get the top-level compile function working and tested (with examples)
- [ ] ~~Create a validator to catch naughties in code (like errant import/export statements)~~
- [ ] ~~Provide a validation API with error reporting (closely related to the validator I hope!)~~
- [ ] ~~Compile from typescript (!)~~
- [ ] ~~Add wrappers for global state - #17~~
- [x] Add imports for a given adaptor

## Devtools /  CLI

Probably installed as a global to work on adaptors, or run straight out of kit when working on the runtime and compiler

- [x] Write a new devtools CLI which loads a jobfile and executes it with the runtime
- [x] Write output to disk
- [x] Use strong conventions to run from a single folder, but provide overrides
- [x] Ensure the runtime module loader knows where to find modules (OPENFN_MODULE_HOME)
- [x] Allow individual module (adaptor) paths to be passed in
- [x] Run in a nested process so that consumers don't have to pass in experimental flags
- [x] create a bin stub
- [x] publish to npm as a pre-built package (Stu did this)
- [x] Add compile only task (not really happy about how this is set up tbh)
- [x] Use peer dependencies rather than building stuff in

## Runtime Manager

The runtime manager /service provides an API to run jobs in worker threads.

There's an API to support and report on thread management, and there's a little web server which stays alive and receives post requests to run jobs. The server#'s just a demo client of the manager API really (but it'll be a fun toy to build out!).

- [x] Use a worker pool to execute jobs in thread
- [x] Accept jobs as strings and compile them (with a cache)
- [x] Track which threads are running which jobs, plus some usage stats
- [x] Work out a way to unit test the thread pool
(remaining steps moved to #52)

## Ava Notes

I've disabled type checking in ava's unit tests because:
1) It slows them down
2) There's absolutely no reason to type check here
3) If ava does encounter a type error anywhere in its test or source compilation, it'll hang and timeout without reporting anything (I have lost hours to this).
4) I for one am happy to relax typing rules while developing against tests (it's often useful to have unreferenced variables and imports, for example!)

Type errors will still get caught by the build process in CI. We could consider adding a `test:tsc` script as well if we want seperate CLI/local test runner for typings. At the moment I have no appetite for this.

Also worth nothing that ava works nicely in monorepos - just drop a config in the top level and ava will find it from down in the packages. Now it's easy for all packages to share a standard test definition. Neat!

## Naming Jobs

When running a job, we ideally want to give it a useful name. Probably found in the preceeding comments. But how will the runtime know the name?

I think we need like an Operation function, which explicitly declares an operation with a name and a handler.
```
export default [
  Operation("do the thing", fn(() => { ... })
]
```

Alternatively we could ad arguments to every adaptor function to do this stuff.

#### the state global

I want to remove the state global. The currnet `core.execute` function loads state into the 'global' context. This doesn't work very well with my new vision of "immutable" state, where we clone the state object and pass it into each operation.

I would like to:

* Remove global state from the runtime environment
* In the compiler, wrap every operation in a `(state) => x` wrapper. Maybe only if a flag is provided, maybe only if we detect a global state reference inside that operation.

This means old code can be made to work if it's compiled properly, while new code can be encouraged to adopt better practice.

## General TODOs

* [ ] Think a bit more about source maps. Runtime needs to be able report errors against source map positions. Source maps could be inline, disk space isn't a priority.
* [x]  Create a logger service (TODO create an issue to track this)
* [ ] #19


josephjclark:
Running into quite a big issue in the CLI (devtools).

The linker in `@openfn/runtime` is failing to resolve imported modules properly - ie, if you import `@openfn/language-common` in your job, and load that job through the CLI, the CLI will fail to load the module and throw an error (out of the runtime stack).

If I use `path.resolve` and write out the path manually to `node_modules`, it's fine.

The issue lies somewhere in module resolution, and I can see from a trace with `NODE_DEBUG=module` that it's resolving the modules from `packages/runtime/dist`. That's a bit of a surprise as I reckon it should be resolving from `packages/cli`  🤔

Even more surprising, `@open/language-common` is actually available on that path anyway. So it should be silently running.

Modules like `ava` and `rollup` are resolving OK from within the job. Haven't worked out yet what's special about them.

The problem is consistent with the `tsm` loader and when running the built CLI from `dist` with `node`. 

I think `pnpm` might be interfering with module resolution here. Everything is symlinked an I think it's triggering module resolution to run from the wrong root. This would be fine if running `@openfn/cli` as a globally installed package (I think), but a big use-case for the CLI is to run it straight out of kit while working on the compiler or runtime.

I will spend more time on this when I get back - just wanted to document the struggle here.

stuartc:
@josephjclark I just went through this PR, this is really great - it's feeling like a very solid foundation.

The CLI appears to expect a folder? Or a file called `job.js`, for a first pass would it not be easier/clearer to expect the exact file instead of a path that contains files with a naming pattern?

> TODO what if stdout and output path are set?

I think in this case, then `console.log` and other things are redirected to `STDERR`. And the resulting state is sent to `STDOUT` and a file. It's a good question because what I just proposed is a pretty niche interface.

Perhaps we make the option mutually exclusive for now, you pick just one.
But this does lead me to think we might be better off logging to `STDERR` if you have `--stdout` enabled.

> Security thoughts: the process inherits the node command arguments
> (it has to for experimental modules to work)
> Is this a concern? If secrets are passed in they could be visible
> The sandbox should help

I'm wasn't immediately concerned about this, like I couldn't think of what might be passed in. I think what _is_ super important is code accessing `process`, including importing it from `node:process`, we're far more concerned with someone accessing `process.env` or adding a callback to `process.on`. 

We will need to test how the sandboxing can deal with this. I think that's the only handle we really have, I don't know how we could 'knee-cap' the worker without making life very hard for everyone - we should focus on the sandboxing and test that. 

As an aside, it wouldn't be unreasonable to pass the process a 'cookie' (Erlang terminology) or key/secret used to trust other nodes in the cluster. Basically I do think there could/would be a requirement to have workers carry _some_ kind of temporary and constant trust mechanism for communicating with the backend cluster. 

You'll have to walk me through that module resolution issue you outlined above, I'd like to understand it better.

josephjclark:
@stuartc Thanks!

The CLI takes a file or folder and uses conventions to work out the rest (where everything can be overridden if you like). That seems the simplest thing (and takes exact path(s)). If you think there's an easier way an example would help!

Re `process`: we can prevent imports of `node:process` via the runtime linker (also the validator in the compiler can catch simple imports in advance). Passing a whitelist of allowed modules names, with a sensisble default, is a good solution here. Meanwhile the `node:vm` sandboxing ensure there's no `process` variable available to the job scope. I've done a bit of sniffing around here but let me double check and maybe add some unit tests to prove the point.

Passing a trusted key to different nodes is a good idea. I don't know a lot about clustering, never really done it (certainly never done anything interesting with it). We'll have to look at this a bit further downstream I think.

josephjclark:
TODO: how can we make devtools safer when running customer data through it?

A good step would be to allow state.json an even credentials to be fetched from a URL (rather than a file) and never written to disk.

Another would be to allow devtools to manage things like credentials and state. You can add credentials and devtools will store them somewhere secure (maybe local but maybe not), and will provide automation on clearing those eg every day). Maybe you can define a project, set a delete timeout, set credentials, etc. That project will be saved into an encrypyed local db. Devtools needs a password to decrpyt the credentials. The project can have named and default state and credentials. So a lot of the stuff that lightning does would be done on your local machine by devtools.

What about sensitive logging?

stuartc:
@josephjclark Aaah! Yeah local dep hell, a special place.

This is far from a good solution, but I have previously (like years ago); made a little bash script to build, pack and install into a tmp folder in the _reverse_ order, if my memory serves correctly `npm` won't try and install if the version already exists.

You can also kinda recreate a global install (without polluting your local global installs) with a `npm install --global-style ...` flag. This _may_ not work because it doesn't flatten dependencies (they go in the direct deps' `node_modules`)

I'm gonna test this out, both as a pre-deploy check and see if we can make this easier on all of us.

stuartc:
@josephjclark works my side, published!

Btw I used `changeset` to do the publishing, not sure how I didn't think to check it can handle that... No more pnpm filtering needed. I updated the README publish command to reflect what I did.

josephjclark:
What do we need to do here before merging to main?

* [ ] Need a :+1: from Stu (although I'm hoping that's a formality at this point!)
* [ ] Need to check the workflow-diagram tests now that they're convert to ava
* [ ] Are we happy with ava instead of mocha and chai?  I have no strong opinion, i just want to be consistent. Ava seems fine at the moment, I have no motivation to change!
* [ ] If so, we need to port describe-package tests over to ava
* [ ] Update lightning to use the new describe-package (instead of `compiler`)
* [ ] Update the top level readme

josephjclark:
@stuartc If and when you have time, we can build and publish a new version from the latest commit. Recommend you build the local package and give it a sanity check with `openfn --test` and `openfn --test --log=debug` or something.

There's a changeset for the new version (it's basically just the logger)

josephjclark:
I've just made the runtime state MUTABLE, so that it's compatible with legacy jobs. That''ll make the migration to lightning easier.

There's a flag in the runtime (and CLI) to force immutable state.

packages/workflow-diagram/test/layout.test.ts
```diff
@@ -0,0 +1,50 @@
+import test from 'ava';
+
+import { ProjectSpace } from '../dist/types';
+import { toElkNode, toFlow, doLayout } from '../src/layout/index';
+import { FlowElkNode, FlowNodeEdges } from '../src/layout/types';
+import { getFixture } from './helpers';
+
+test.skip('toElkNode should convert a project space to a workflow', async (t) => {
```
josephjclark, line 8: I've had to skip this test because the returned tree doesn't have any children, which is a bit worrying.

This will need to be resolved before merging.

packages/workflow-diagram/test/fixtures/single-workflow-nodeedges.json
```diff
@@ -309,7 +309,7 @@
       "labelBgStyle": {
         "fill": "#f3f4f6"
       },
-      "className": "dashed-edge"
+      "className": ""
```
josephjclark, line 312: Surprised at this diff - I don't expect any changes here. I can't see where dashedEdge is actually assigned so this feels correct, but I'm concerned about these changes in general

josephjclark:
So I've been through and updated everything that  I think needs doing. The examples now work, the readme is updated.

We do need to update Lightning to use the new `describe-package`, as after merging we can't update the old package (at least not easily). But it should be OK to merge before we do this. Perhaps it needs an issue over on the Lightning repo.

@stuartc  I think you need to give this a quick once-over - check the top level readme and package json, plus check my previous comment re describe-package tests. It's probably fine and we could merge and fix later.

Subject to that, I think we can merge...

PR #12, opened 2022-08-25
tsconfig refactor

josephjclark:
Really small thing but I wanted to spin it out of the v2 work I'm doing.

As I'm creating more packages I'm duplicating a lot of tsconfigs - I jut wanted to take a breather and abstract out the main compiler rules into the top level and have each package extend it.

Packages are technically free to add their own rules if they want, but we should be looking to apply a consistent set of standards across everything in this repo.


stuartc (approved):
Wonderful, been wanting to do this. :+1:

PR #26, opened 2022-09-30
New logger stuff

josephjclark:
Implementation of @openfn/logger

* [x] A new logger service which looks like Console, supporting log, info, debug, warn, error and success level logging
* [x] Pluggable emitter layer so you can redirect the output (to an array for unit tests or http for remote logging)
* [x] Icons are assigned to log levels for clear and concise output 
* [x] Loggers can be named and have priority levels assigned (ie ignore all debug level logging)
* [x] CLI creates named loggers for each component (cli, runtime, compiler, job)
* [x] `state.configuration` objects will obfuscated in logs (this is very crude)
* [x] Runtime logs how long each operation and the whole pipeline took
* [x] Pre-canned silent logger which just swallows logs(useful for testing and defaults)
* [x] Mock logger with an inspection API for testing
* [x] Mucho unit testo

There's also some general typing and refactoring lumped in here, mostly things that annoyed me while working.

image


josephjclark:
Reminders for me for next week:

* The logger is now reasonably well implemented in the CLI (and pushed into the compiler). CLI tests are passing but there's some skipped tests which use the logger. I need to enrich the mock logger API to report what its output has been (like I'm doing in the logger/mock tests
* Compiler tests are all failing - probably just assumptions around the new logger api
* Remove any references to `silent` in the CLI
* Integrate the new logger in the runtime (don't worry about the event emitter stuff, I think that's orthogonal - the runtime will log AND emit)

PR #33, opened 2022-10-07
#24 remove workflow box and update highest workflow box

### LINKED ISSUE #24: Remove workflow box and update highest workflow box to include a name
### User story
As a user that is looking at a workflow for the first time, I would like to easily identify the name of a workflow and understand when it gets triggered. 

### Details
remove the “workflow” wrapper node
change the “webhook” node to have the workflow name and describe how the workflow gets triggered. (make it bigger so that it is obviously different to the rest of the nodes in a workflow)

workflow

### Implementation notes

1. clone `OpenFn/kit`
2. read `README.md`
4. start the flow example with `pnpm run -C examples/flow start`
5. cd into `/packages/workflow-diagram` and start a build/watch with `pnpm run watch`
6. take a look at the example which is now running in your browser on http://127.0.0.1:8080/
7. take a look at how this example gets "fed" by a project description at `examples/flow/src/index.jsx`
8. head over to `packages/workflow-diagram/src/layout/index.ts` and see how the example gets built.
9. To start work, you _might_ approach this by removing the wrapper node entirely and _modifying_ the trigger node so that it displays information about both the workflow (name) and the trigger (details).

wanecode:
## Related issue
#24 

Fixes #24 


stuartc (approved):
@wanecode this looks fine, I made a few changes.

I dropped the old code since there was no other code calling them anymore. And renamed the files to reflect the naming of the new Nodes you added.

Lets try not adding named exports when the original default export isn't being used anymore. Usually (not always) default exports are a hint that this module has a single purpose. Since your stuff supersedes what was there you can replace the exports.

I see you copied the design exactly, we should have mentioned pretty much any picture uploaded is a mockup. I have tweaked the css classes to make the TriggerWorkflowNode match the JobNode.

PR #54, opened 2022-11-01
Fix globals

josephjclark:
This PR contains a fix where the compiler would try and auto-import globals like Date or Promise from the adaptor.

The compiler shouldn't do this for globals, so the fix is simply to expand the list of globals being tracked.

This PR contains a changeset. I've also run `changeset version` and commited it in just to track what's changed since the last release.


josephjclark:
@stuartc  this doesn't really need a review, but when you have time could  you run a release? Obviously runtime-manager shouldn't be published yet.

PR #56, opened 2022-11-03
Autoinstall

josephjclark:
This PR adds autoinstall functionality to the the new runtime and CLI.

The heavy lifting of actually installing a module to a repo, as well as fetching it out again, is handled by the runtime.

The CLI is responsible for the strategy of when modules should be installed.

Note that the CLI now uses the new repo to lookup type definitions before passing them in to the compiler (TODO: is this stuff unit tested?)

Major changes introduced here:

* The CLI now uses subcommands, like `execute`, `compile` and `test`.
* There's a new subcommand called `repo`, with `install clean list` and `pwd`
* The runtime is now aware of a thing called a repo, which is a private npm repository with pre-installed npm modules. It also includes an API to install new modules (but has no autoinstall of its own)
* The CLI will autoinstall an adaptor if the `-i` flag is passed. If no version is provided the latest available on npm will be loaded.
* The OPENFN_MODULES_HOME env var is gone, and basically replaced with OPENFN_REPO_DIR. It's not quite the same thing, but it's so easy to install adaptors now I don't mind the breaking change.

TODOs

* [x] Allow the runtime to install and load modules from a repo (an arbitrary folder)
* [x] Allow the CLI to install the provided adaptor (using a runtime api) if the the `-i` flag is passed
* [x] Refactor a `repo` subcommand into the CLI with `repo install/clean/list/pwd` commands
* [x] Refactor `modulesHome` into `repoHome` in the CLI
* [x] Simplify and properly unit test module path lookups in `cli/src/compile/compiler.ts` - there's one gnarly function there which gives very little feedback and yet is absolutely critical to auto-imports.

## Docs

To auto install a language adaptor, pass the `-i` flag (use info logs to track what's happening)

```
openfn job.js --log info -i -a common
```

To explicitly install a number of adaptors, use `repo install`
```
openfn repo install -a common http dhis2
```

To see whats installed:

```
openfn repo list
```

To clear out the repo:

```
openfn repo clean
```

This is safe to call redundantly - the runtime will check that a matching adaptor is installed before doing anything.

Note that short-form and long-form adaptor names work so long as the `-a` flag is in the command.


josephjclark:
One change coming in on this branch is that the CLI will no longer call out to unpkg to preload type definitions for a module. So far as I can tell this is absolutely fine.

At the moment, we have logic which will auto-insert an import statement for a job, importing adaptor functions into the scope.

There is logic to lookup the adaptor's exported functions and only add those to the import statement. So if we know what `language-common` exports, we only add those to the `import` statement. This means we don't try and do something stupid like `import { Date } from '@openfn/language-common'`.

However, if we don't have any definitions for an adaptor, we actually just dumbly import any undeclared variables - which in practice, works just as well.

The intelligence to call out to `unpkg` is  staying in the compiler's util `preloadAdaptorExports`, I'm just decoupling it from the CLI. Because now the runtime maintains its own repo of packages, and if the package we want isn't available locally, the runtime will fail anyway.

I hope this makes sense when I re-read it tomorrow :thinking:

josephjclark:
Alright, I think this is ready for merge. I'd appreciate a quick test from @stuartc before we release and invite more folks to look at it!

stuartc (approved):
This is great @josephjclark! A few questions/suggestions:

- If you're finding `npm view <...> version` a bit flaky, perhaps adding the `--json` flag would be helpful. Since the result is JSON.parse-able even with the extra characters..
- Does `repo install` have to take a `-a`? Is that just to keep the args handling easier? It would be cleaner if we didn't need a flag for the adaptor list since there aren't any other flags for install.
- It would be nice if we used `defaultRepoPath` in the Opts as a default, so `repo pwd` doesn't return `undefined`.
- Would `which` or `show` be a better command for `pwd`? Since pwd makes me thing of "my shells pwd"?

I'm going to merge this anyway, but would like your thoughts on the above. :D

josephjclark:
Sorry for the late reply on this!

* I don't think `npm view` is a long-term solution but it let me move forward. `--json` is a good tip though, thanks!
* The `-a` in install just enables a shorthand, expanding adaptor names from `http` to `@openfn/language-http` . It enables a shorthand. Maybe this should be default as adaptors are all we'll be installing for a while, but  I'm trying to keep the api forward looking.  If it's proving to confusing I'll address it, although most users will be doing `openfn execute -ia` anyway
* Yes, we should use `defaultRepoPath` in pwd! Whoops.
* Yeah `pwd` isn't ideal, although tbh I don't think `which` or `show` are much clearer. Open to ideas on this. Also if you do `openfn repo list` it'll show the working dir anyway. Maybe we should just do `openfn repo`?

PR #66, opened 2022-11-09
Transform project space

### LINKED ISSUE #60: Transform projectSpace JSON data for presentation concerns
### User story
As a user looking at the workflows diagrams I want to see the title `Untitled` for untitled workflows and a description for `webhook` and / or `cron` workflows.

### Details
For `webhook` the description should be like this `when data is received at https://demo.openfn.org/i/34f843bd-eb87-4833-b32a-905139534d5a` and for `cron` the description should be like `"at 7:34pm every Tuesday"`

elias-ba:
CircleCI failed to execute the command `corepack enable && corepack prepare pnpm@7.14.2 --activate`. This seems to be a common Circle CI problem and seems to be related to switching between `~/bin` and /usr/bin` folders in CI. I had to to force CI to use `/usr/bin` by adding `sudo` to that command in the config file of Circle CI to solve the issue. We may probably need to investigate more about this but for now it works.

## Related issue

Fixes #60 


packages/workflow-diagram/src/index.tsx
```diff
@@ -28,7 +29,16 @@ const WorkflowDiagram: React.FC<{
...
+      const jobs = addDescription(projectSpace.jobs);
+      const workflows = renameUntitledWorkflows(projectSpace.workflows);
+
+      const transformedProjectSpace = {
+        ...projectSpace,
+        jobs,
+        workflows,
+      } as ProjectSpace;
+
+      Store.setProjectSpace(transformedProjectSpace);
```
stuartc, line 41: Instead of changing the ProjectSpace, I think the description should be computed in the Node.

Have a look at `layout/factories.ts`, you can attach things to the `__flowProps__` objects and handle them in the components. This way we aren't breaking the contract of the ProjectSpace type by manipulating it (adding description) and we open up the pathway for avoiding re-renders on the individual nodes (_we don't currently do anything about this now, but when/if we do this pattern helps_).

packages/workflow-diagram/src/types.ts
```diff
@@ -2,12 +2,14 @@ import { Edge, Node } from "react-flow-renderer";
 
 export type CronTrigger = {
   type: "cron";
+  expression: string;
```
stuartc, line 5: Could you change this to `cron_expression` and `url` to `webhook_url`.

While these fields look like they are their own interfaces, they are types that are unioned: `export type Trigger = CronTrigger | WebhookTrigger | FlowTrigger;`

So it's important we use more specific names, I suggest sticking to the same name we use on Lightning, i.e. `cron_expression` etc. 

This change will need to be done in https://github.com/OpenFn/Lightning/pull/412 as well.

stuartc (changes requested):
Good work so far. Please have a look at my comments.

By the way, the flow example renders a blank diagram on my side, the `nodes` prop is `[]` - not sure whats going on there.

packages/workflow-diagram/src/types.ts
elias-ba: Done. Instead of snake case naming (`cron_expression` and `webhook_url`) I did chose camel case naming (`cronExpression` and `webhookUrl`) to respect our naming in Kit. I have also updated the Lightning side (see [here](https://github.com/OpenFn/Lightning/pull/412/commits/96fb03510c4f7d19f32bd4710dd2a4d7bf5de7fb))

packages/workflow-diagram/src/index.tsx
elias-ba: Done.

packages/workflow-diagram/src/layout/factories.ts
```diff
@@ -56,10 +74,10 @@ export function triggerNodeFactory(job: Job, workflow: Workflow): FlowElkNode {
     __flowProps__: {
       data: {
         label: TriggerLabels[job.trigger.type],
-        description: job.trigger.description,
-        workflow,
+        description: generateDescription(job.trigger),
+        workflow: renameWorkflowForUntitled(workflow),
```
stuartc, line 78: Ah man! So close! What I intended was that the actual Job/Workflow/Trigger object is passed in and the `generateDescription` would be in `TriggerWorkflowNode.tsx` for example. If we pass in the exact object we used to decide where the node goes in the tree we will be able to do all sorts of ah-hoc rendering in the component.

In the near future we might allow us to inject the nodes just before using the library, meaning we might have a moment when we have this specific nodes inside Lightning.

stuartc (changes requested):

packages/workflow-diagram/src/layout/factories.ts
elias-ba: Hey @stuartc this is done now. Thanks for the catch. Seems much better now. Give it another round of review and let me know.

elias-ba: Done

elias-ba:
Hey @stuartc this is done now. Thanks for the catch. Seems much better now. Give it another round of review and let me know.

stuartc (approved):

Issue #70, opened 2022-11-14
CLI: running openfn repo list before the repo is inited causes an exception


josephjclark:
![image](https://user-images.githubusercontent.com/7052509/201616078-f5439dce-4e75-4587-8a9b-2f58872bde67.png)

PR #74, opened 2022-11-15
Cli fixes & Improvements

### LINKED ISSUE #67: Mtuchi feedback
Issues found in cli from pairing with Mtuchi:

* CLI help says you can use "trace" level, which is wrong
* The runtime logger is printing details in the config object - shouldn't the logger be  obfuscating values?
* The no exprts found warning is probably a bit too severe given that it really doesn't matter...
* `~` in adaptor paths isn't working

josephjclark:
Sundry fixes here:

* `openfn repo list` won't throw if the repo is uninitialised
* Better handling of the default repo path
* Fixed docs for logger (removed reference to trace)
* Better debug logging in the linker and repo
* De-emphasise the "exports could not be found" warning as it doesn't matter and looks too scary
* Globals are fine and working as I expect (if a package isn't in the repo then a general `import()` will load the global).
I think tilde in path is fine - works for me (TM) in job path, adaptor path and state path. There are until tests for this in `cli/test/commands/test/ts` (see `run a trivial job from a folder: openfn ~/openfn/jobs/the-question`)
* When a specific adaptor version is passed in, the runtime now actually uses it

Close #67 and #70

### DIFF
--- a/packages/runtime/src/modules/linker.ts
+++ b/packages/runtime/src/modules/linker.ts
@@ -3,10 +3,15 @@
  * The tricky bit is this MUST load all linked libraries in the context of the parent module
  * https://nodejs.org/api/html#modulelinklinker
  */
+import { createMockLogger, Logger } from '@openfn/logger';
 import vm, { Module, SyntheticModule, Context } from './experimental-vm';
 import { getModulePath } from './repo';
 
+const defaultLogger = createMockLogger();
+
 export type LinkerOptions = {
+  log?: Logger;
+
   // paths to modules: '@openfn/language-common': './path/to/common.js'
   modulePaths?: Record<string, string>;
 
@@ -15,6 +20,7 @@ export type LinkerOptions = {
 
   whitelist?: RegExp[]; // whitelist packages which the linker allows to be imported
 
+  // TODO deprecated - use the logger instead
   trace?: boolean; // log module lookup information
 };
 
@@ -25,10 +31,10 @@ export type Linker = (
 ) => Promise<Module>;
 
 const linker: Linker = async (specifier, context, options = {}) => {
-  const { whitelist, trace } = options;
-  if (trace) {
-    console.log(`[linker] loading module ${specifier}`);
-  }
+  const { whitelist } = options;
+  const log = options.log || defaultLogger;
+  log.debug(`[linker] loading module ${specifier}`);
+
   if (whitelist && !whitelist.find((r) => r.exec(specifier))) {
     throw new Error(`Error: module blacklisted: ${specifier}`);
   }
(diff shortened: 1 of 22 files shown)


josephjclark:
Tests pass while that naughty one on `workflow-diagram` is skipped. We should probably sort the test before merging though.

josephjclark:
Hey @mtuchi, do you fancy running some QA on this? It would be super helpful.

You could re-try the stuff we did yesterday, running different versions of `mssql`. You should be able to tell from `debug` logs which version is being used - you should check this though and verify. You may also want to check the attached issues and see if everything looks alright.

Interested in particular in the use of `~` in adaptor paths. It seems fine to me!

This isn't released yet so you should:
* Remove your existing CLI `npm remove -g @open/cli`
* Checkout this branch
* Build: `pnpm i && pnpm build`
* Install the local CLI globally: `cd packages/cli && npm install -g
* Test the global install `openfn test`
* Run your tests
* Remove the local build when you're done to prevent confusion: `npm remove -g @open/cli`

packages/workflow-diagram/test/layout.test.ts
```diff
@@ -5,7 +5,7 @@ import { toElkNode, toFlow, doLayout } from '../src/layout/index';
 import { FlowElkNode, FlowNodeEdges } from '../src/layout/types';
 import { getFixture } from './helpers';
 
-test('toElkNode should convert a project space to a workflow', async (t) => {
+test.skip('toElkNode should convert a project space to a workflow', async (t) => {
```
stuartc, line 8: This is fixed in `main` now

packages/runtime/src/modules/linker.ts
```diff
@@ -77,31 +91,38 @@ const linker: Linker = async (specifier, context, options = {}) => {
...
+    // TODO: should we log WHY this path was used? Maybe this is enough for now
+    log.debug(`[linker] Loading module ${specifier} from ${path}`);
     try {
       return import(path);
     } catch (e) {
       if (options.trace) {
-        console.warn(
-          `[linker] Failed to load module ${specifier} from ${path}`
-        );
+        log.debug(`[linker] Failed to load module ${specifier} from ${path}`);
```
stuartc, line 118: I'm curious about the shift from `warn` to `debug` (not the use of the logger), surely with these kinds of situation we should be logging above `debug`?

stuartc (approved):
Looks good, just two comments - but yeah lemme know when you want me to merge and release.

josephjclark:
Whoops - I forgot to check why the logger is printing config stuff (I'm sure it shouldn't be!)

packages/runtime/src/modules/linker.ts
josephjclark: It's because it's OK for this to "fail" and think a warning is too scary at this stage. if the module can't be found in the repo it'll do a regular import and pull in the local or global package, if found.

BUT I need to remove the outer `if (trace)`, because trace has been deprecated.  I managed to miss that somehow. And actually I should go ahead and remove trace properly.

There is another question here. Now that we prefer to load modules from the repo, I think we should probably disable the `import(specifier)` fallback. That will load globals or packages installed in the runtime, but otherwise fail. It's not very traceable if it goes wrong (or right).

But that's probably outside the scope of this, PR, so I'll spin out an issue

josephjclark:
Turns out the logger wasn't sanitising a state object if it didn't have `configuration` AND `data` keys. The rules have been relaxed and unit tests added. Sorted.

packages/runtime/src/modules/linker.ts
josephjclark: Ok I've removed trace fully (thanks for the catch) and raised #78

josephjclark:
@mtuchi has given this a :+1: , so @stuartc You can release and merge when you're ready!

Issue #84, opened 2022-11-17
Tailwind: what's the correct way to use styles in library components?


josephjclark:
The components we create in `kit` use embedded CSS to style themselves.

This causes problems in Lightning because tailwind utilities are defined several  times in conflicting CSS, resulting in weird stuff.

I have suggested a quick workaround that keeps us moving forward, but this needs more thought.

Seems to be that if I have a tailwind component with a bunch of util classes, it should "inherit" those classes from the parent app. So if I use `text-s` as 56px in my parent app, the component should use the same. Because of how Tailwind works, this approach (if its even correct) requires the parent app build to process the utility classes used by the components in node_modules.

I suspect the answer may be to create a Tailwind plugin for the component. I think this allows us to declare and contribute styles used by the library component to the parenting app, so it can generate the correct CSS. But this needs more research.

This problem has actually been on my mind all week, although until now it had felt a bit academic. I haven't been able to Google a good solution yet, but I will keep looking. The Tailwind UI library probably hold the answers (which is what clued me into plugins)

stuartc:
I've just made a change to `adaptor-docs`, doing what your initial hunch is - produce an `index.css` file that the consumer decides how and when to import.

So in the case of Lightning since we also use tailwind we _don't_ import it - we point our tailwind config to go and sniff around the module to determine what styles should be added to it's css.

It is curious that this isn't an issue with `workflow-diagram` since it's the exact same initial setup. I can see it judiciously injecting CSS into the page - so I think we should take the same approach with that package as well.

PR #85, opened 2022-11-17
Fix tailwind styles

josephjclark:
This is a suggested quick fix to the styling issues in Lightning.

* use the `prefix` option in Tailwind config to prefix all styles
* Sadly that means I've had to manually prefix every utility class used in the  components
* Remove base styles from the generated CSS (on the assumption that Lightning is using it)

This breaks the example a little bit but I don't think I mind at this stage as I see this as a temporary solution. If it ends up being permanent then, when I've stopped vomiting, I'll add tailwind to the example page and it should all work.

I've raised #84 to track the real issue, which I'll address as a priority this afternoon/next week.

Also, unrelated, apparently I didn't push my last commit to the main PR. Bit worrying. So I've included that, which updates the `export` utility script.

Oh, there's a little bit of tidying up of some unused stuff here as well.


packages/adaptor-docs/tailwind.config.cjs
```diff
@@ -1,5 +1,6 @@
 // This is duplicated and lightly edited from Lightning
 module.exports = {
+  prefix: 'ad--',
```
josephjclark, line 3: There's quiet a noisy diff here (sorry!) but this is basically the fix

packages/adaptor-docs/src/DocsPanel.tsx
```diff
@@ -23,9 +23,9 @@ const DocsPanel = ({ specifier, onInsert }: DocsPanelProps) => {
   
   const { name, version, functions } = pkg as PackageDescription;
   return (
-    <div className="block m-2">
-      <h1 className="h1 text-lg font-bold text-secondary-700 mb-2">{name} v{version}</h1>
-      <div className="text-sm mb-4">Operations available for use by this adaptor are listed below.</div>
+    <div className="ad--block ad--m-2">
```
josephjclark, line 26: This is also part of the actual fix - repeated across both components in the app.

stuartc:
I'd prefer to try the approach of not bundling the css (via injection) for now - assuming we don't run into any issues with the approach here https://github.com/OpenFn/kit/pull/87.

If we want to really get into prefixing in order to have truly isolated styling then this is the way to go - and if we do I'd like to make sure we have a rock solid way to both minify and dedup.

Long story short, I'm starting to think components shouldn't be injecting styles - consumers can but not our deps? Your thoughts?

josephjclark:
I will leave this open until next week because I still need to move the export utility across, so it'll remind me of that. Then I'll shut it down.

I do agree that the component shouldn't be injecting their own CSS. Especially tailwind components because they're all going to be injecting the same stuff.

I'll reserve more comment until I get a chance to do more research next week :)

PR #92, opened 2022-11-22
Fix exports from common

### LINKED ISSUE #86: describe-package: export/from seems to export symbols as "unknown"
Here's a weird thing.

I was just looking at the docs for `primero` and `describe-package` isn't reporting the exports from common.

In `Adaptor.js` we do this:

```
export {
  alterState,
 . ..
} from '@openfn/language-common';
```

Looking at the output of `getSymbols()`, typescript seems to recognise the names of all these exports as `unknown`. It doesn't seem to pull out any useful information at all.

I wonder if that's because there's no definition of `common` available in the TS project. In fact that's probably it: even if it was able to scrape the names of these exports, it wouldn't be able to report any type information.

So, `describe-package` may have to somehow pull the definitions of any dependencies into the Project in order to properly report types. Because this doesn't just apply to language adaptors - `http` exports `axios`, for example...

josephjclark:
This PR ensures that adaptor documentation includes functions exported from language-common.

Basically `describe-package` was failing to describe anything at all for functions exported from another package - ie, the way language-primero exports most of  language-common. This is because the local typescript context only has definitions for the top package (primero) and not its dependencies (common).

This is a quick fix which always loads common alongside whichever adaptor is being described, which should fix most if not all exports for the forseeable future.

Ultimately, the fix for this is to dynamically import whatever is needed. But that has to be a server-side operation, not something that will run in the client. So I'm putting it off for another day.

Note that there is still some chaos in `describe-package` itself as  I organise things around. It'll be improving in smaller PRs throughout the week.

## Change summary

* describe-package now loads common as well as the target DTS file
* Added a drop-down in examples/adaptor docs allowing two different packages to be explored (including Primero, which is complete thanks to this fix but will be broken in older versions of describe-package)
* A bit more unit test coverage (but still not very good I'm afraid)
* `adaptor-docs/useDocs` will set its cached documentation to null whenever the specifier prop changes (so that it'll correctly show load state)
* Only export/describe public functions

## Related issue

Fixes #86

### DIFF
--- a/packages/describe-package/src/api.ts
+++ b/packages/describe-package/src/api.ts
@@ -43,8 +43,6 @@ export type ParameterDescription = {
   type: string; // this is a human-readble string I think. Should we also store a machine readable type?
 };
 
-// I think we can get the API down to just these functions
-
 export const describePackage = async (
   specifier: string,
   _options: Options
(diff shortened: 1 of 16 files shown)


packages/describe-package/src/api.ts
```diff
@@ -43,23 +44,29 @@ export type ParameterDescription = {
...
 
-  const files = await fetchDTSListing(specifier);
+  // Include language-common in the project model
+  // (I don't expect this to be permanent)
+  const common = await fetchDTSListing('@openfn/language-common');
+  for await (const fileName of common) {
+    const f = await fetchFile(`@openfn/language-common${fileName}`);
+    // Flatten the paths or else there's trouble
+    // TODO need to better understand this at some stage
+    const relativeFileName = fileName.split('/').pop();
```
josephjclark, line 61: This is a bit janky - but common's paths are nested in `types`, so you have paths like `/types/Adaptor.d.ts` coming out of the package.json.

This basically breaks the tracking in the TS project. Flattening the paths to simple file names seems to work better.

I would like a better solution but I'll sort it out when we do proper dynamic dts dependency tracking.

stuartc (approved):

stuartc:
@josephjclark this cool merge yet? I see a couple of commits since the review request?

josephjclark:
@stuartc Yeah this is good to go! Something crept up out of a chat with Taylor but no biggie

PR #95, opened 2022-11-22
Build stuff

josephjclark:
Sorry, another PR!

This one:
* Restores the tsup build in adaptor-docs (I haven't added a changeset as it shouldn't be public facing)
* Restores the top-level export utility script.

Just tidying up really.

Testing this locally, Lightning doesn't seem to build my component's CSS properly, so the component is all over the place. This happens with the latest deployed version too, and I note that demo.openfn.org is fine. So I think it's just my setup?

Before merging, you should:

* `pnpm -C packages/adaptor-docs build`
* `pnpm export adaptor-docs`
* Check that the built component in Lightning looks OK (the copy/add buttons should be right-aligned and the window shouldn't stretch to fill all the space).

If this is all OK then there's nothing to worry about and we should merge the PR. I'll fix my local issues separately.


stuartc:
@josephjclark when you say all over the place, like is the CSS being built properly? Does the `dist/index.css` file get created on build?
I had to add the `external: true` option to `postcss` on the rollup config to get it to not inject stuff - I can't remember what esbuilds defaults are for css imports.

As a hack - we _could_ try and point lightnings tailwind at the installed modules dist directory to slurp up any styles used.

I'll take a look at this in a bit, just wanted to get a sense of how confident you are it's your system..

josephjclark:
@stuartc I am confident that this is fine and the build is emitting all the right things.

I just can't really test properly as my Lightning seems to have a separate issue.

Ignoring this PR, if I go to Lighting/main and reset everything (`adaptor-docs` 0.0.3 installed) and start the server, there are missing styles in the docs panel. I see those styles in `assets/node_modules/@openfn/adaptor-docs`, but the tailwind build seems to not incorporate them. The tailwind config file seems correct.

So it's something funky at my end, which I'm going to have to go and debug next as I'm about to work on an issue at the Lightning end

josephjclark:
Just pushed an update 'cause it turns out I missed a few things :blush:

packages/describe-package/package.json
```diff
@@ -64,7 +61,6 @@
   },
   "dependencies": {
     "@typescript/vfs": "^1.3.5",
```
stuartc, line 63: How many of these do we need in `dependencies`, sorry some of them might have been from me.. Can you do a quick pass to see which ones are absolutely required for runtime.

stuartc (changes requested):
@josephjclark this looks good dude! One note on the deps for `describe-package` and thats it - will merge after your feedback on that.

packages/describe-package/package.json
josephjclark: Yeah good questions.

I've raised an issue to remove `cross-fetch` entirely when we can safely upgrade engines to node 18.

Over in #100 I've actually made a few changes to the bundle. I'll review this over on that branch so I'm not stepping on my own toes...

josephjclark: So for what it's worth I think most of these should still be considered dependencies.

I've moved `tsup` because it's clearly a dev dep, and `threads` because it's bundled into the web worker build. The rest I think are appropriately `dependencies` (at least for now)

stuartc: To be sure, this good to go then?

josephjclark: This is good to go :+1:

PR #97, opened 2022-11-23
describe-package: Don't describe beta

### LINKED ISSUE #80: describe-package: handle namespaces in export descriptions
Some packages export a `beta` namespace.

At the moment, `describe-package` presents the beta namespaced functions as a flat list alongside the main functions - which is a) confusing and b) can lead to duplication (for example, `common` exports two `each` functions).

We basically need to de-flatten the exported list (or flag the namespace).

Probably we need to utilise the new `package` structure which is returned and `namespaces` to the `functions` properties, or something.

josephjclark:
A simple fix to exclude beta from the documentation exported by `desrcribe-package`.

This is the dumb solution which is fine for now.

What we should REALLY be doing is parsing the index and tracing the actual exports. This would let us properly trace export aliases and present a more accurate picture of what the adaptor exports (not just what code happens to be lying around in the folder).

I've updated the internal documentation to represent what describe package now does:
* Only describe functions marked as @public in the jsdoc
* Ignore the beta.d.ts file

This closes #80

You can test it works by starting up `examples/adaptor-docs` and looking at the dev console. No more duplicate key errors.

### DIFF
--- a/examples/adaptor-docs/tsup.config.js
+++ b/examples/adaptor-docs/tsup.config.js
@@ -1,48 +1,54 @@
 import { defineConfig } from 'tsup';
 import Koa from 'koa';
 import serve from 'koa-static';
-import websockify from 'koa-websocket'
+import websockify from 'koa-websocket';
 
 import copyStaticFiles from 'esbuild-copy-static-files';
 
 let app;
-let listeners = []
+let listeners = [];
 
 const onSuccess = () => {
   if (!app) {
     app = new websockify(new Koa());
-    app.use(serve('./dist'))
-    console.log('Server running at localhost:1234')
-    app.listen(1234)
-    
+    app.use(serve('./dist'));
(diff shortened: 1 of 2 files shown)


examples/adaptor-docs/tsup.config.js
```diff
@@ -1,48 +1,54 @@
...
-  watch: ['.', '../../packages/adaptor-docs/dist'],
-  noExternal: ["@openfn/adaptor-docs", "react"],
-  esbuildPlugins: [copyStaticFiles({
-    src: './static/',
-    dest: './dist/'
-  })],
+  watch: [
+    '.',
+    '../../packages/adaptor-docs/dist',
+    '../../packages/describe-package/dist',
```
josephjclark, line 40: The diff here is mostly just prettier (idk why it didn't run before) - what's actually changed is I've added `describe-package` to the watch list here.

Issue #98, opened 2022-11-23
CLI is broken on main


josephjclark:
Something has broken on main - running the CLI locally throws this:

![image](https://user-images.githubusercontent.com/7052509/203548097-6b92b0ce-96c4-48de-9238-8fbad90e86e4.png)

This must be a result of the last merge, probably something in describe package.

I don't think any of this has been released but I just need to work this out and bring the packages back into line before the next release.

josephjclark:
Hmm, looks like `describe-package` 0.0.9 has an issue in that it exports Project as a type, not a class (or whatever it is). That's a simple fix but the compiler, at least, will fall over if it tries to use it.

PR #99, opened 2022-11-23
CLI: Fix compiler globals

josephjclark:
This PR contains a small fix for the CLI/compiler, allowing extra globals to be supported in job code.

Basically when we  compile code and generate an import statement, we sometimes (wrongly) add globals to the list of imports. It's the same problem as back in #54 (back then I was thinking about node globals, today 'm thinking about browser  globals).

I've expanded the list of supported globals to include everything on [MDN's list](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects). I'm sure something else will creep through at some stage.

## Relrases

I hope I've done this right: I've run `changeset version` for the CLI and compiler, which should be released ASAP. There are some unreleased changesets hanging around which should not be released yet.

There's a bit of complication because main is slightly broken - see #98. I've rolled back the compiler's `describe-package` dependency and added a couple of simple integration tests (which fail on main). So the new CLI and compiler releases from here should be stable.

To release, you'll need build everything and just publish `cli` and `compiler`.


josephjclark:
@stuartc When you're ready tomorrow this is a small fix to the CLI. All versioned up, just needs a release.

packages/compiler/src/transforms/add-imports.ts
```diff
@@ -15,29 +15,80 @@ import type { Transformer } from '../transform';
...
   'clearTimeout',
   'console',
+  'DataView',
   'Date',
+  'decodeURI',
+  'decodeURIComponent',
+  'encodeURI',
+  'encodeURIComponent',
   'Error',
+  'eval',
```
stuartc, line 42: I'm a little concerned about adding `eval` here, obviously I don't want to break any adaptors - but if we allow `eval` we have to ensure it is sufficiently isolated or proxied. Like that vm2 vuln. where the calling context is available via the Error prototype; if you think we should allow it and handle the sandboxing in a more holistic manner in the near future I'm ok with that.

stuartc (commented):
@josephjclark this looks good, just would like your thoughts on `eval`.

packages/compiler/src/transforms/add-imports.ts
josephjclark: I'll push a commit in a few minutes with some unit tests, but globals like `Set` are available, while `eval` is disabled and `process` not in scope.

I have raised an issue #104  to review and improve unit testing around this.

josephjclark: Ok - this is good to go! See the latest commit for a couple of comforting unit tests.

Re release - I suggest we merge this  into main, then release from the `release-next` branch and do it all in one go.

josephjclark: .. apart from the failing test of course

josephjclark: Ok, all done!

PR #102, opened 2022-11-23
Move Plus Icon In Job Nodes

elias-ba:
## Related issue

Partially Fixes [#396](https://github.com/OpenFn/Lightning/issues/396)


taylordowns2000 (changes requested):
@elias-ba , would you adjust the plus node so that it sits above the edges in a z-index sense? see how in this image the edges come through the plus node? it doesn't look very clean that way.
image

elias-ba:
@taylordowns2000 thanks for the catch. I tried to solve it but I think it's a little bit more tricky that it seems. This one happens only when the Job Node has operations. When the Job Node has operations, it's div has a `z-index` of -1. I can fix your catch by changing that, but then we will lose the ability to see the links in the different operations of the Job Node. What do you suggest ? cc @stuartc

stuartc:
@taylordowns2000 the z-index issue with nested operations is really tricky to solve. The main issue is that all the lines are rendered on one z layer/stack, and all nodes that 'are parents' get given a z-index of `-1`; which also resets their position in the stack - meaning no matter what z-index you use inside that div, from the outside something that is like `z-index: 1` will still win.

There are a number of approaches we can take, none of them are straightforward.. This isn't an issue when not showing operations btw. We can tackle this again when we add operations.

stuartc:
@elias-ba I've rebased this branch and made a small tweak, could you add a changeset version to this branch and I'll happily merge; thanks man - this looks good!

stuartc (approved):

PR #110, opened 2022-11-25
Tighten up output and support circular structures

### LINKED ISSUE #106: CLI: exception thrown when trying to serialize non-serializable state
From @mtuchi:
```
 OpenFn-WCS-Projects git:(main) ✗ openfn kobo_monitoring/initiateGetSubmissions.js -a http=/home/openfn/Workspace/adaptors/packages/http -s tmp/state.json -o tmp/output.json
[CLI] ✔ Loaded state from tmp/state.json
[CLI] ✔ Compiled job from kobo_monitoring/initiateGetSubmissions.js
[JOB] ℹ Next round of getFormDataFromKobo metadata {
  "trigger": "Get kobotoolbox form data",
  "metadata": {
    "uuidIndex": 0,
    "totalFormsUuid": 57,
    "getSubmission": true
  }
}
[R/T] ✔ Operation 1 complete in 0ms
POST request succeeded with 200 ✓
[R/T] ✔ Operation 2 complete in 1.457s
[CLI] ✔ Writing output to tmp/output.json
file:///home/openfn/.asdf/installs/nodejs/18.12.0/lib/node_modules/@openfn/cli/dist/process/runner.js:353
    await writeFile(options.outputPath, JSON.stringify(result, null, 4));
                                             ^

TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'ClientRequest'
    |     property 'socket' -> object with constructor 'TLSSocket'
    --- property '_httpMessage' closes the circle
    at JSON.stringify (<anonymous>)
    at executeHandler (file:///home/openfn/.asdf/installs/nodejs/18.12.0/lib/node_modules/@openfn/cli/dist/process/runner.js:353:46)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v18.12.0
```

This is expected but not something I've dealt with yet.

It's common practice in jobs right now to write functions and stuff to state. When writing that state to disk (or I guess stdout), than acn causes problems. In this case the structure is circular. It may have even come from the adaptor.

I think the solution is for the sanitizer to detect function and class types and either delete them or generate a nice string. Worth checking out engine deals with this today.

Related: when we serialize configuration to disk, should it be sanitized? Even better, should we chop config out of the resulting data? In Lightning the config data is held in a sensitive credential and shouldn't be written to disk where just anyone can copy it out of the result data.

josephjclark:
Three fixes here related to what we write to `output.json`

1) Support circular structures thanks to `fast-safe-stringify` (fixes #106 which started all this)
2) Introduce a "strict mode" for state output, which only prints state.data to the output file
3) Don't serialize `configuration`

The second part in particular is worth some discussion.

## Strict output

By default the CLI will output in strict mode. This means we only print the `data` key of state. Anything else written to the state object will be ignored.

This avoids us serialising junk which is needed at runtime but shouldn't be returned at the end. Like request objects from the HTTP adaptor (which is the trigger for all this).

In non strict mode (`openfn job.js --no-strict-output`) we'll try and print the whole state object (except config, see below). We should be able to handle complex objects should anyone try to serialize a database connection or something.

## Config output

We now never serialize the `configuration` object in the output.

This means that in lightning, sensitive credential data is never printed out. We could take the approach of of sanitizing the output (scrubbing the values), but to what end? The configuration object is input-only and shouldn't IMO be shown in the output.

## Other ideas

Maybe we can ignore all keys starting with `_` ? Just at the top or nested? In strict mode or always?

## Issues

Closes #106

### DIFF
--- a/.changeset/famous-chicken-marry.md
+++ b/.changeset/famous-chicken-marry.md
@@ -1,0 +1,7 @@
+---
+'@openfn/cli': patch
+---
+
+Support circular structures in JSON output
+Introduce strict output (by default) which only serializes data
+Never serialize configuration to output
--- a/.changeset/rude-shrimps-beg.md
+++ b/.changeset/rude-shrimps-beg.md
@@ -1,0 +1,5 @@
+---
+'@openfn/logger': patch
+---
+
+Allow null and undefined to be logged
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -48,6 +48,7 @@
     "@openfn/compiler": "workspace:^0.0.16",
     "@openfn/logger": "workspace:^0.0.6",
     "@openfn/runtime": "workspace:^0.0.11",
+    "fast-safe-stringify": "^2.1.1",
     "rimraf": "^3.0.2",
     "treeify": "^1.1.0",
     "yargs": "^17.5.1"
--- a/packages/cli/src/cli.ts
+++ b/packages/cli/src/cli.ts
@@ -5,9 +5,10 @@ import { repo as repoCommand, install as installCommand } from './repo/command';
 import executeCommand from './execute/command';
 import compileCommand from './compile/command';
 import testCommand from './test/command';
+import { Opts } from './commands';
 
 export const cmd = yargs(hideBin(process.argv))
-  .command(executeCommand as yargs.CommandModule<{}>)
+  .command(executeCommand)
   .command(compileCommand)
   .command(installCommand) // allow install to run from the top as well as repo
   .command(repoCommand)
@@ -19,4 +20,4 @@ export const cmd = yargs(hideBin(process.argv))
     array: true,
   })
   .alias('v', 'version')
-  .help();
+  .help() as yargs.Argv<Opts>;
--- a/packages/cli/src/commands.ts
+++ b/packages/cli/src/commands.ts
@@ -24,6 +24,7 @@ export type Opts = {
   packages?: string[];
   statePath?: string;
   stateStdin?: string;
+  noStrictOutput?: boolean; // defaults to false
 };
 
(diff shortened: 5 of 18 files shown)


josephjclark:
@taylordowns2000 I'm gonna ask for your input on this one - not so much on the code (although you're welcome to dig in) but the approach outlined in the description. Let me know what you think.

taylordowns2000:
super interesting @josephjclark . i love not writing configuration and we could start _coaching_ people to only use `data` between runs, but at the time of writing we'd only be able to introduce the new CLI to `platform-app` in `--no-strict-output` because so many flows rely on jobs that leave things in other, custom parts of state.... most notably customers often create a `state.cursor` key which holds data on the last record the fetched the last time the job ran, or something like that.

stuartc (approved):
@josephjclark this looks and sounds good to me... 

I like that for now, it's the default and we can opt out.
Regarding not serialising `configuration` I think is great, we are going to have in Lightning (like in platform) scrubbers for configuration as well (and I see no reason to _not_ do them even with this).

One thing I personally like is the dropping of everything other than `data`. But I'll have to refer to @taylordowns2000 on that. I have no idea how many clients might lean on that (albeit probably bad practice).

taylordowns2000:
@stuartc , what do you think of this? 

> super interesting @josephjclark . i love not writing `configuration` and we could start _coaching_ people to only use `data` between runs, but at the time of writing we'd only be able to introduce the new CLI to `platform-app` in `--no-strict-output` because so many flows rely on jobs that leave things in other, custom parts of state.... most notably customers often create a `state.cursor` key which holds data on the last record the fetched the last time the job ran, or something like that.

i think we both commented at the same time :-)

PR #100, opened 2022-11-23
Next release

josephjclark:
This branch contains a bunch of fixes for the next set of releases.

I am suspicious of the adaptor-docs build here. It seems to work fine, but I've removed a bunch of build stuff to get it working on the web in the first place. I mean I love deleting code, and I love it when stuff works. But it took a while to get this working in the first place and it's a bit suss that I can just delete it...

Suggest we release from here then merge into main.

## Changes
This PR includes the following new releases:

* adaptor-docs 0.0.5
  * Collapsible sections
  * Show loading when the specifier prop changes
* describe-package 0.0.10
  * Don't describe beta
  * Only describe public functions
  * Include exports from language-common
* CLI 0.0.17
  * Use latest compiler dependency
* Compiler 0.0.17
  * updated dependency on describe package
  * Extended globals list

## QA

I would recommend a little bit of QA

* Build locally
* Run tests locally
* Install the CLI globally (`pnpm pack:local` and `npm install -g dist/openfn-cli-0.0.17``and run `openfn test` to ensure it's working
* Load adaptor docs into Lightning (`pnpm export describe-package && pnpm export adaptor-docs`) - docs should load collapsed

## Related PRs

Change tracking:

* [x] Build stuff #95 
* [x] Fix build #98
* [x] Dependency updates
* [x] Hide beta from docs #97
* [x] Collapsible docs #94

When this is released we can action #96

Also: we need a better release process. After this, I'll start creating `release/` branches with sets of related releases on them. Should make things easier to organise.

PR #111, opened 2022-11-25
Release cli@0.0.18

josephjclark:
This PR is to track the 0.0.18 release of the CLI (plus its dependencies)

Changes:
* [x] #100
* [x] #109
* [x] #108
* [x] use `jsdelivr` #112

I have run `changeset version` so version numbers are accurate.

If merging jsdelivr as well, we'll need to run this again to update `describe-package`


josephjclark:
@stuartc I think this is ready for release on Monday. We may want to merge #112 as well, in which case we need to run changeset version again.

PR #119, opened 2022-12-02
drop wf diagram

taylordowns2000:
This is the other half of moving workflow-diagram to Lightning: https://github.com/OpenFn/Lightning/pull/486


josephjclark (approved):
Looks good to me, does what it says on the tin. I suggest we hold off the merge until everything is OK on the other side.

PR #123, opened 2022-12-08
cli: enable adaptors to be loaded from the monorepo

### LINKED ISSUE #55: CLI: Be monorepo aware
Now that we have a monorepo, we should make devtools aware of it.

If you're developing a language adaptor, you'll be writing it in the monorepo. You want to be able to use the new CLI to test that adaptor against jobs.

If you give the CLI a path to the monorepo, any adaptor imports should load straight from the monorepo.

This can be used a couple of ways:

* pass in a `monorepo=path/to/monorepo` flag (or `-m path/to/monotro`)
* Set an env var and pass a `-M` flag to the CLI (makes monorepo use easy but still won't load from the monorepo by default)

Some thoughts:
* We should load from the build, not the source (this also ensures that .d.ts files are there)
* Logging should make it very clear when adaptors have been loaded out of the monorepo (success level logging)
* If using the monorepo all version stuff is ignored - no autoinstalls or anything, and any version directives in job code will be ignored

josephjclark:
This PR makes the CLI monorepo aware. Closes #55

I need this while working on magic function stuff because I want to run my local adaptors, rather than autoinstalling them. So it make sense to spin it out into its own feature.

This will be really helpful for testing `adaptors` with the CLI.

Here are the monorepo rules
* Pass the `-m` (`--use-adaptor-monorepo`) flag to use the monorepo
* The monorepo path will be looked up from the OPENFN_ADAPTORS_REPO env var. Setting the env var for your system seems much cleaner than passing a path each time
* Success-level logging will tell you when the monorepo is being used
* When loading from the monorepo, version specifiers are ignored (with a warning)
* Autoinstall won't run if the monorepo is configured and enabled (but a warning will be shown if -i is passed)

Info level logging should make it clear where things are being loaded from, in case it isn't clear if the monorepo is actually being used.

Note that this won't do anything to build the adaptor before running. If you're working on an adaptor you'll have to run `build:watch` .

Other notes:
* If `-m` is passed and there's no env var, or the env var doesn't point to `adaptors` (or at least something that looks like adaptors), the process will exit with a helpful message (error code 9)


packages/cli/src/util/use-adaptors-repo.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (adaptor.match('=')) {
+    // Should do nothing if a path is already provided
+    return adaptor;
+  }
+
+  const { name, version } = getNameAndVersion(adaptor);
+  if (version) {
+    // version numbers are ignored with warning
+    log.warn(
+      `Warning: Ignoring version specifier on ${adaptor} as loading from the adaptors monorepo`
```
josephjclark, line 30: Actually, if you ask for a specific version, probably we should use that version and ignore the monorepo. Like `openfn execute -a http@1.0.0` should lookup the 1.0.0 version in the repo, not whatever is in the monorepo.

josephjclark:
I am worried that automatic monorepo support is going to be annoying. Once the env var is set, there's a risk that you don't really know which adaptor is being used. Yes we can add logging but logs are easy to miss. Someone like Mtuchi will both write job and adaptor code and it could really get confusing.

So actually I'm gonna drop the env var and just use on a a `-m` flag. It's the least risky thing. If users really really want an env var I'll add it later.

josephjclark:
Hi @mtuchi, when you get a chance to you think you could run some QA on this, make sure it works in your environment? Thanks!

josephjclark:
This has cleared Mutchi's QA.

I think I'd like to merge adaptor versions first and release that as the next version, then put this and JSON logging and stuff as the release after that.

josephjclark:
@stuartc Could you give this a quick review and merge today?

stuartc (approved):

PR #125, opened 2022-12-08
Magic Function Demos

josephjclark:
## Overview

This PR adds:
* A magic function demo playground in `examples`
* A `metadata` command to the CLI

You can read more about magic functions in the [documentation](https://docs.google.com/document/d/1fjRq6YNw5fExNlH0wkDL8MHdhhBcYioLKAIOnO8XPFE/edit?usp=sharing).

The example package is mostly deprecated already because that stuff has been moved into Lightning - but it probably is useful to keep for experimentation and further development, so I'm keen to keep them around for now.

## Metadata Command

This PR exends `@openfn/cli` with a `metadata`  command

The basic API is:

```
openfn metadata -s path/to/config -a adaptor-name -o path/to/output
```

Adaptors will be resolved just like normal, so you can pass the monorepo flag `-m` or an adaptor name to be loaded from the repo, or a path to a location on disk.

CAVEAT: at the time of wrtiting you have to pass a path to the index.js entry point. This is very fixable, I just want to avoid code re-use so I need to think a little bit about the solution.

Output can be a path with `-o` or an in-line json object with `-O`.

There are no real unit tests yet, just an integration test.

At the moment there's a simple cache in place. Config objects are hashed into a key and stored under this key in the filesystem.

## Demo Server

To run the example in `examples/magic-functions`:

Install stuff:

`pnpm install`

Start the demo (and cross your fingers)

`pnpm -C examples/magic-functions start`

Go to `localhost:1234/` and you should be presented with links to the editor and explorer demos.

The demo contains a pre-loaded metadata.json file and custom dts definitions for adaptors, so it's totally standalone.

There are two parts to the demo:
1) A monaco editor with smart code assist
2) A datasource explorer (targeted at Salesforce but  totally generic)


josephjclark:
Outstanding work here:

* [ ] Update the docs above (they're already a bit out of date, we can remove a lot and link to the new doc)
* [ ] Make sure it'll work without the monorepo (ie, a published adaptor or normal adaptor paths) 
* [ ] Unit tests on the new CLI stuff
* [ ]  `--no-cache` option
* [ ] Remove the `ensure-opts` stub
* [ ] Should this work with autoinstall?
* [ ] Release

josephjclark:
More todos:

* [ ] Track the time metadata was computed (as meta prop on the root?)
* [ ] Force re-generation through the CLI
* [ ] Include adaptor name in the cache? Would config ever be the same across adaptors...?

josephjclark:
Closed in favour of https://github.com/OpenFn/kit/pull/185/, where I've fixed all the CLI stuff.

For now I'm gonna drop the demos as all that stuff is moving into Lightning and I don't want to duplicate maintenance.

PR #127, opened 2022-12-13
describe-package: handle captions in examples better

josephjclark:
Some examples include a caption, which is like a label.

We don't handle that very elegantly - at least, not until now.

It's an awkward cross-package issue but this This fixes https://github.com/OpenFn/Lightning/issues/503

I'll test this against a corresponding fix in Lightning but when we're ready, I suggest we release `describe-package` here then merge down into main.

Update: I've updated workspace dependencies to `workspace:*`, otherwise things get a bit inconsistent if you just try to bump and release a single package.

### DIFF
--- a/.changeset/breezy-baboons-scream.md
+++ b/.changeset/breezy-baboons-scream.md
@@ -1,0 +1,5 @@
+---
+'@openfn/describe-package': patch
+---
+
+Handle captions in examples
--- a/.changeset/breezy-baboons-scream.md
+++ b/.changeset/breezy-baboons-scream.md
@@ -1,5 +1,0 @@
----
-'@openfn/describe-package': patch
----
-
-Handle captions in examples
--- a/packages/adaptor-docs/package.json
+++ b/packages/adaptor-docs/package.json
@@ -29,7 +29,7 @@
   "author": "Open Function Group <admin@openfn.org>",
   "license": "ISC",
   "dependencies": {
-    "@openfn/describe-package": "workspace:^0.0.11",
+    "@openfn/describe-package": "workspace:*",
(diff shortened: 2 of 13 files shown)


josephjclark:
@stuartc This is ready to go, with a new `describe-package` bundle release.

I've hit an interesting question with Changesets.

I only want to publish `describe-package` with this change, but `pnpm changeset` will bump anything the depends on it to accept the new version in its package.json. This results in a bunch of meaningless versions being published.

Ok, that doesn't really matter, but I don't much like it.

If I discard the irrelevant versions, the repo falls in an inconsistent state. `describe-package` gets bumped to `0.0.12`. But dependencies reference `workspace: ^0.0.11`. I think that should map to 0.0.12 but it doesn't, and I get an error like `In packages/compiler: No matching version found for @openfn/describe-package@^0.0.11 inside the workspace`.

One solution is to update dependencies to say `workspace: *` for `describe-package`. This I think means "use whatever is in the workspace", which is resolved at publish time.

That should let us do partial releases. I am not sure what the consequences are.

Thoughts? Maybe we should keep it simple and release everything...

stuartc (approved):

stuartc:
@josephjclark yeah hmm... I'm not opposed to trying out a wildcard for workspace deps... Perhaps we can do a local pack and just see what it does.

Either that or we adopt semver a little more.. ala `~0.1.0` kinda thing.

I'm not going to make the changes today so I'll ship a new everything now, but yeah I'm with you on having a problem cutting dozens of new versions for a typo.

josephjclark:
Alright, just pushed the change. You should be able to install, build and release and it'll just publish `describe-package`

Issue #130, opened 2022-12-14
cli: log version details


stuartc:
### Details

When running cli `execute`, I'd like to see:

- The current version of node
- The adaptor version (both the resolved and specified version)  
  In the case of `@latest` it should show `@1.2.3 [latest]`.
- The version of the CLI at least, and possibly the version of the runtime and/or runtime-manager.

### Implementation notes

Lets try log this out to STDERR, so it's easier to filter in Lightning

josephjclark:
The resolved adaptors should come up in debug logging at least. I think we put something in info? I'll check.

Logging to stderr feels pretty weird and won't play well with how the logger works. Not without effort, anyway. How does it help lightning? Because it's easier to isolate the version output?

We could add a [VER] log level or something which prints version info. That would be easier to filter/detect. You could even suppress all logging apart from VER through the CLI itself

stuartc:
We don't _have_ to log to STDERR - it's just easier to split pipes in childprocess streams when calling out to the shell.

Like if it's gonna break a bunch of stuff we can leave it. But if it's doable it would be nice, regular shell users won't notice anything.

josephjclark:
I've had a little swing at this in #138 but I'm really not that happy about it.

You can now do `openfn version`, which will print out version diagnostic info. It could also include adaptor version if you pass an argument.

Would it make more sense to run `openfn version` first, divert that output to wherever Lightning needs it, and then run the actual job?

taylordowns2000:
@josephjclark , @stuartc , not sure of _how_ (of course... sorry!) but the goal here is to have the adaptor version show up inside this BEAUTIFUL list:
![image.png](https://images.zenhubusercontent.com/566ac5e0638e718c995f5356/d4c548aa-d117-43ff-b6a4-3369b1e28f93)

stuartc:
@josephjclark this output is usually used in debugging on production, like knowing exactly what version the job was executed with, since `latest` doesn't mean much a week after the event - and whenever we were on occasion fixing things with core, it was good to know that a given machine in the cluster was in fact up to date etc.

So these kinds of details are pretty important, if you feel they are noisy when using it via the terminal we can perhaps come up with a way to make them opt-in or tagged in a way they can be filtered out.. Maybe a flag is the way forward for that.

To summarise, we want to know the adaptor version as well. In core after we resolve the adaptor, we resolve the 'root' (this may actually be very difficult with ESM modules :$ ) and read the package.json file to get the specific version

[...]

PR #136, opened 2022-12-16
CLI: Report adaptor versions

### LINKED ISSUE #76: CLI: Add a preflight check at the start
It would be really nice to have a sort of pre-flght validation check at the start.

This should try to resole all the input paths - the adaptor, the state, the job etc - and report an error if any path can't be resolved.

This would provide nicer, cleaner error reporting and make it easier for users to understand if they've mis-typed a path. At the moment it'll just report fairly arcane errors saying it can't find certain paths, and it may not be clear what's wrong with those paths.

The validation should log passes on info (ie, tell us what you found), and raise errors by default.

An alternative to this, I guess, would just be to have better error messages which give a little more feedback.

josephjclark:
This PR logs the version of the adaptor being used in a run. It does this in two ways:

1) At the top of the run, the requested adaptor version is printed alongside other version details. This is really just an echo of the input arguments
2) When the linker loads an adaptor (and at the moment I do mean an adaptor specifically), it will print to `info` the name and version of the resolved adaptor. The specific path used is available in `debug` logs.

We log the details to `info` because, on a successful run, it's probably not that relevant which version was actually used. The user would reasonably assume they got what they asked for, and default logging is really about what's absolutely necessary. Sticking it in `info` logging makes useful to power users, who I expect to use info logging most of the time.

I've also added a little validation before the job starts (although not much, in the end, because pre-validating is too difficulty to do safely and responsibly).

## Output

Ignore `compiler: workspace` etc, that's just from running out of my local source

`openfn job.js` (no adaptor)

image

`openfn job.js -a common@1.7.5`

image

`openfn job.js -a common`

image

## Issues

Closes #76 and #130


packages/cli/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.21",
   "description": "CLI devtools for the openfn toolchain.",
   "engines": {
-    "node": ">=16",
+    "node": ">=18",
```
josephjclark, line 6: Just want to flag this - this branches off of #134 and so it it also requires us to bump node up 18. I think this is probably OK but need to make the risk really clear.

josephjclark:
I think this is done - it would be great to get a little QA from someone to make sure I'm covering the most common cases of getting adaptors wrong (in the new CLI, at least)

josephjclark:
This has given me some trouble this afternoon and I've gone down a bit of a rabbit hole on it. I'm gonna step away and come back and review with fresh eyes tomorrow.

Basically the validator will now lookup the latest version if no version information is passed, and report that in logging. This is probably correct and good.

image

The trouble with this is that the validator is now implementing its own adaptor version/path lookup logic, but not feeding that back to the actual command. So the compiler and runtime may decide to do their own lookups based on the adaptor name specifier.

I think it would be smarter for the CLI to work out the paths to the adaptor and then write that back to the command, to be shared by other stages. This makes the validator less a validator and more a pre-calculator or resolver, but there we go.

Basically it's that old problem that we repeat complex path calculations in  different parts of the stack. This had gotten a lot better with the repo, because most code now just asks the repo "get me the path to this adaptor". But the validator is now short of bypassing that.

josephjclark:
Decisions with Stu:

1) echo out the specifier in the main version list (even if it says latest)
2) When we do the ACTUAL loading, print the version that we find (at info)

josephjclark:
@stuartc Ok, I've been looking at this for a while and I think it's in good shape. It's changed a lot since the original PR.

Can you give it a review and merge?

stuartc (approved):
@josephjclark so if I'm reading this correctly, you're falling back on a CJS resolution strategy for finding the package.json right?

Anyway this looks good, I'll leave this approved and will hold off merging until I get your thumbs-up on the json logger status. :+1:

josephjclark:
@stuartc The json logging stuff is separate to this and doesn't need to get involved (EDIT: you've merged while typing this :+1: )

Yes, we use CJS module versions internally. We load package.jsons by path, look up the CJS entry point, and load that. This is fine in node 18 and 19 for adaptors, whereas the ESM load actually gives us problems. We'll have to revisit this one day but  not for a long while. This was all done in #147

PR #141, opened 2022-12-20
Fix import undefined

josephjclark:
Turns out that if you have `undefined` in your JS, we'll treat it as a global variable and try to import it from the adaptor.

Whoops.

This PR fixes that and covers a couple of similar cases.


mtuchi:
@josephjclark I have tested this fix and it works as expected

josephjclark:
@mtuchi Thanks - I've added the changeset!

I'll have this merged into the next release. Thanks for testing!

Issue #133, opened 2022-12-15
cli: Running "openfn docs adaptorname" should print a list of available functions


josephjclark:
.

PR #145, opened 2022-12-21
CLI: more docs fixes

josephjclark:
So `describe-package` changed how it handles examples in #127 and someone forgot to update the CLI to the new structure. So docs example would output [object Object]. 

This PR fixes this.

I've also sneaked in a fix for #133 because I was right in the zone.


mtuchi:
Merging this since I need this feature for Ayder's job writing training

Issue #140, opened 2022-12-20
Runtime: linker is broken in node 19


josephjclark:
Ruh Roh.

The CLI is broken in node 19.3.0. This is trouble.

The problem is that directory imports (ie, `import repo/adaptors/http`) have... sort of stopped working?

In node 18 we use the flag `--experimental-specifier-resolution=node` to be able to import from folders and not use file extensions in imports (which is a real pain in ts). But that flag has been removed in node 19.

In node 19 I can't even do this in a clean npm repo
```
import _ from "lodash/fp";
console.log("ok");
```
This throws `ERR_UNSUPPORTED_DIR_IMPORT`

This doesn't just affect the runtime - the http and salesforce adaptors both use lodash/fp, but apparently that won't work in node 19.

josephjclark:
The problem here is that we're using node's modern module system, esm, but writing imports like in the old way (ie, `./script` instead of `./script.js`).

That's one thing - if we have to we can rewrite (or more likely) transpile the runtime codebase to use file extensions, as node seem to insist we do.

I am a bit confused by why `import('repo/@openfn/language-http')` has stopped working. I want the importer to look at the package.json found in that folder and use that to know how to load the module. Somehow that (not to mention `lodash/fp` are broken.

Also worried  that this is going to run into `adaptors` and we're  going to have to get even more dogmatic about how modules are defined.

josephjclark:
Apparently this is the answer: use this loader:

https://github.com/nodejs/loaders-test/tree/main/commonjs-extension-resolution-loader

Issue #78, opened 2022-11-16
Linker: Disable global imports?


josephjclark:
Since 0.14 the runtime uses a repo - basically a folder with a package json - to load dependencies from. It's up to the runtime manger to ensure these dependencies are installed to the repo so that the runtime can find them (maybe layer we'll let the runtime autoinstall but that's a problem for later).

But if the linker can't find a dependency a) at an explicit path passed in or b) in the repo, it'll do `import(specifier)` as a fallback, resorting to nodes standard module resolution. This means that packages loaded into the runtime or installed globally will happen to work. But there's very little transparency on this, so if you're accidentally using a globally installed adaptor, there's basically no way to know about it (there is a little debug warning but it's subtle and easy to miss, let alone understand).

So I think we should remove this global fallback entirely. Maybe for now I'll wrap it in an option which defaults to false.

stuartc:
All for this, can’t find a package after an attempted auto-install (or not); explosions.

PR #147, opened 2023-01-04
Fix linker in node19

josephjclark:
Plot twist: the CLI is pretty much broken in node 19.

Luckily this PR fixes it with minimum drama (?). All changes are compatible with node 18.

## The Issue

We use `--experimental-module-specifier=node` to enable the `import()` statement to load ESM and CJS modules. Specifically we rely on the ability to import from a directory - which is allowed in CJS but not in ESM.

Node 19 removes this flag which leaves us rather up the creek. I misunderstood the flag - I thought it enabled new behaviour, but it turns out its enabling legacy behaviour.

## The Fix

The fundamental problem is that ESM won't `import()` from a directory, like `/repo/node_modules/@openfn/language-common`. Which is basically what we need.

So I've tweaked the linker so that it'll look in the repo for a matching module, load the package json, find the entry point, and load that file directly.

Now, the catch is, the linker will only load CJS formats of any dependencies. Essentially, for all the modern module support, the runtime actually runs in legacy CJS mode. I think this is absolutely fine - it's just worth being aware of.

If you try and load the ESM format, and the module's dependencies use a directory import (like `lodash/fp`, which comes up a lot), it'll throw an exception. For now I've commented out the ESM import code and tests.

## Other Notes

As I've been in the area I've also disabled global imports. So now you either have to map a module to a specific path, or rely on it being installed in the repo. This does have some consequences in unit tests (which is actually quite reassuring!)

This issue affects adaptors too because ESM style imports of anything which uses `lodash/fp` will now fail. This branch is now robust to this, and platform probably is too as it'll all be CJS imports. I have raised a PR in adaptors to start future-proofing usage of lodash.

## Issues

#140 #78

## Testing

Here's my test job (thanks to @mtuchi for the cool http test). On main, in node 19, this fails. 

```
// job.js
get('https://api.chucknorris.io/jokes/random');
fn((state) => {
  console.log(state.data.value);
});

// cli
openfn job.js -a http
```


packages/runtime/src/modules/linker.ts
```diff
@@ -90,35 +88,37 @@ const linker: Linker = async (specifier, context, options = {}) => {
...
     ({ path, version } = options.modules?.[specifier]);
   }
 
   if (!path && options.repo) {
-    // Try and load a matching path from the repo
     const specifierWithVersion = version
       ? `${specifier}@${version}`
       : specifier;
-    path = await getModulePath(specifierWithVersion, options.repo);
+    path = await getModuleEntryPoint(specifierWithVersion, options.repo);
```
josephjclark, line 108: This is the actual fix.

Instead of looking up the path to the module root in the repo, we look up the path to the entry point. Because I'm only interested in the CJS entry point this is actually pretty trivial.

PR #152, opened 2023-01-06
CLI integration tests

### LINKED ISSUE #75: CLI integration tests
I'd like to setup a suite of integration tests for the CLI.

There's a lot of stuff that's hard to mock (and I'm not a fan of intricate mocks) or isn't really suitable for unit tests (not a fan of calling out to the web in unit tests)

What we need to do is:
* Setup something like a docker image
* Checkout the latest commit
* Build it
* Install the CLI globally
* Run a bunch of CLI commands
* * `repo list`
* * `install <various versions of the same module>`
* * `execute a simple job against each adaptor version, with different arguments provided`
* Clean the repo

These are designed to be run sequentially.

This should include tests with a global module too

josephjclark:
Starting point for an integration test suite for the CLI.

The aim here is not to write an exhaustive set of tests - just to set up a framework to make integration tests easy.

This should not be tied to any particular release (although I'd like to put 0.0.27 out before merging).

## Architecture

The basic idea here is to test the built (but unreleased) CLI bundle in the global space. I want integration tests which are just commands that you can copy and paste and run on your own machine.

To achieve this, we do the following
* Build and pack:local the repo (note that we can now a) exclude version numbers from the back and b) specify the output directory of the packages). This gives us a set of tarballs which reference each other's @openfn dependencis (rather than calling out to npm)
* Install the CLI globally
* Using `child_process.exec`, run commands against the global openfn
* Assert against std and the output file.

Most of this is done through CircleCI config. There is a dockerfile which basically does the same thing in a docker image - this is useful for isolating and simulating the text environment on your local machine.

When running tests locally, we don't want to run form the global bin - we want to run from the local repo. So there's a little utility function which in a non-production environent will replace the `openfn` command with `pnpm -C ../../packages/cli openfn`. Note that this does used the built `dist` and not the actual source.

## A note on logging

One of the main goals of the CLI is to be transparent and communicative - it should tell the user what its doing. No magic. 

The logging plays a really important function in this and should be tested.

Right now, I'm not doing much test of logs. Just some rough regex on stdout.

When the JSON logging stuff is merged, we'll have much more capacity to test the logs. I'll write a helper API so you can query the logs for eg a debug-level log which matches the regex "adaptor loaded from /repo/http", or just get all error level logs.

Then I can write assertions against log output on specific tests.

## Questions

When do we want these to run? Every branch? Just release branches? Manually?

It should be a slow and expensive test suite, relying on third party services. And it should pass before we merge into main. 

New features should be adding integration tests, in which case they need to run.

Maybe we just always run them until it becomes a problem.

## Related issue

Closes #75


stuartc (approved):
This is great work here dude, really like your approach.

Happy to merge this, one question - the Docker stuff; is that not fully baked yet? Just double checking what I'm seeing - or rather not seeing... Like is there anything that runs the container?

josephjclark:
@stuartc the docker container is just a resource for local dev. Let me double check the npm commands and docs to ensure that's documented

stuartc:
> @stuartc the docker container is just a resource for local dev. Let me double check the npm commands and docs to ensure that's documented

Sweet, some docs would be great!

josephjclark:
@stuartc I've updated some docs.

Are we happy to just to have integration tests run on every PR at the moment? I think "run them until they become a problem" is my preferred strategy

stuartc:
@josephjclark yeah I'm happy with that approach as well

PR #155, opened 2023-01-10
Ensure dist folder exists

### LINKED ISSUE #154: pack:local will fail if there isn't a dist directory
.

josephjclark:
Annoying little problem we just found in #154.

`pnpm pack` will FAIL if `--pack-destination` points to a missing directory. I think this is a `pnpm` issue because that seems very flaky to me.

Obviously we could just run `mkdir dist` before packing, but I'd like a cross-platform solution so ensure that builds work in Windows (very very useful for debugging windows issues). So I've written a tiny node script which does the same thing.

This isn't a very elegant solution @stuartc can you think of anything bettter?

## Repro steps

* Delete the `dist` folder in the repo root (if it exists)
* Run `pnpm -r pack`
* This should explode

On this branch it'll work. It should also work if `dist` already exists of course.

## Related issue

Fixes #154

### DIFF
--- a/build/ensure-dist.js
+++ b/build/ensure-dist.js
@@ -1,0 +1,4 @@
+const fs = require('node:fs');
+const path = require('path');
+
+fs.mkdirSync(path.resolve('dist'), { recursive: true });
--- a/package.json
+++ b/package.json
@@ -10,7 +10,7 @@
     "clean:local": "rimraf dist/**/*",
     "test": "pnpm -r --filter './packages/*' run test",
     "test:types": "pnpm -r --filter './packages/*' run test:types",
-    "pack": "pnpm -r run pack",
+    "pack": "node build/ensure-dist.js && pnpm -r run pack",
     "pack:local": "pnpm run pack && node ./build/pack-local.js",
     "export": "sh scripts/export.sh"
   },


stuartc (commented):
Yeah I see what you mean, I can't think of a better solution to this.

But I do see a small gap for errors here (I mean they exist before this), if you run pack within a individual package you'll still get an error if you don't have the `dist` directory.

The only slightly more elaborate solution would be to 'stub' the entire pack script with our own one, that does the mkdir safely (for each & every call to pack) and then calls pnpm pack. Essentially more of our own build scripts.

But yeah this is good enough for now I think, could you add a changeset and I'll merge this. :+1:

josephjclark:
Yeah, this only helps if you run the top level script. The package packs will still break.

But the docs say to run from root, I always run from root. There isn't much reason to pack a single package, and if you do, you've probably run from root before and so have a `dist` folder. I think it's reasonable to leave that small gap (especially as imo this is all pnpm's fault, surely it should be more robust than that).

There's no public-facing change so I don't think we need a changeset. It's all build stuff.

josephjclark:
Actually there is an easier solution you know - check-in the `dist` folder with a `.gitignore` file which ignores its contents. So the repo guarantees an empty folder.

josephjclark:
Closing as this is superseded by #158

Issue #143, opened 2022-12-21
Runtime: error reporting


josephjclark:
When the runtime generates an exception (usually this will be explicitly through job code, but occasionally it will be a timeout and maybe very occasionally the runtime itself), we need to work out how to display/present it.

A big issue here is stack traces. A stack trace is rarely helpful here. Certainly not for a timeout exception (although it would be useful to know which was the last line to execute). 

This error is from my local system but a runtime error isn't going to be that much different:

![image](https://user-images.githubusercontent.com/7052509/208943185-3df3d56d-1228-4e53-8608-98938e9629ea.png)

The stack trace isn't helpful because either it's been generate from built code, or it's been generated from the job.

We do need to work out how to source map from exceptions in job code and report against the job source. That should be doable.

Another issue here is: it probably makes sense for the runtime to throw Error objects, but maybe the CLI shuold be selective about how it handles them. 

Maybe a first step is for the CLI to not log stack traces with error messages. So long as the error objects are fairly useable and traceable, that's OK.

Then we need to work out source mappings in exceptions generated from jobs.

josephjclark:
See also errors when loading adaptors/modules, or errors in compilation.

There's a rule like system errors don't have a stack trace, while job errors have a mapped stack trace.

Also consider exit codes

josephjclark:
An interesting issue has crept up out of JSON logging. I don't want to deal with it there, but it's an input to this design.

When the synthetic module (ie the job) throws an error, it seems to throw a serialized error object which is not of type `Error`.

What the runtime will need to do is throw some kind of `RuntimeException` (a general error that means something when wrong when running job code), and add properties to reflect what caused the exception (Ie, if the runtime error is a ReferenceException, we need to trap that original type and its message).

We could design an error like this:

```
class RuntimeError extends Error {
 constructor(error) {
    super(error.message)
    this.name = 'RuntimeError''
    this.error = error
 }
  toString(): { return this.error.toString() }
}
```

Such an error would help us understand that we're dealing with an error in user, not system, code, but also preserve the details of the underlying error to deal with as we wish.

I am not sure about the toString. At the moment we log a general `runtime exception` error message and then print the error object. Does the tostring need to specifiy that it's a runtime error, or does it just include the inner error details? That stuff needs a little thought

josephjclark:
The new epic covers the most, if not all, stuff being discussed here. I'll spin out more specific issues afterwards if appropriate

PR #159, opened 2023-01-11
Logger: output as JSON

### LINKED ISSUE #130: cli: log version details

### LINKED ISSUE #157: Logger will show [Object] for objects with deep nesting
Running this job:
```
fn(() => {
  console.log({
    a: { b: { c: { d: 22 } } },
  });
});
```

Results in this output:
```
[JOB] ℹ { a: { b: { c: [Object] } } }
```

So it'll print nesting two levels deep, but not three.

Nest that object in an array and that counts as one nesting, so you get slightly less output.

Basically we should prettify and log the whole object.

This MIGHT cause knock-in issues in Lightning if the string/object gets too big (I can easily imagine someone logging their whole 200mb query JSON). Solutions to that would  be to limit the nesting level, or maybe somehow detect large objects and truncate them aggressively (so basically support a max character limit on the serialized string).

josephjclark:
This PR enables the logger to output  JSON objects rather than strings. So instead of logs like:
```
[CLI] i Loaded adaptor
```
You get logs like:
```
{ level: 'info', name: 'CLI', message: ['Loaded adaptor'] }
```
Objects are shown in-line rather than embedded as a string:
```
{ level: 'info', name: 'CLI', message: [{ data: {}, configuration: {} ] }
```

This is useless to the CLI, but dead handy for integration tests and Lightning as we can easily filter logs.

This also fixes #157, where deeply nested objects will be truncated. We now stringify all output objects, which has the added bonus of supporting circular structures. Note that properties with function values are ignored, which I think is appropriate.


packages/logger/src/sanitize.ts
```diff
@@ -1,31 +1,39 @@
...
+const sanitize = (item: any, options: SanitizeOptions = {}) => {
+  const maybeStringify = (o: any) =>
+    options.stringify === false ? o : stringify(o);
+  if (
+    Array.isArray(item) ||
+    (isNaN(item) &&
+      item &&
+      typeof item !== 'string' &&
+      // duck type an error object (this is not tight enough!)
+      !(item.stack && item.message))
```
josephjclark, line 21: This is a problem: if an error object is passed to the log, it gets serialised to a string.

That's not actually the end of the world but it does break my tests, which expect to read an error object out of the log history, not a string.

The current code is not acceptable because the guard on the error object is too broad. For reasons I cannot grok right now `item instanceof Error` is false for the error object :thinking: 

Solutions:
1) Use json logging in the tests that want an error object. JSON logging won't stringify the object.
2) Make tests work on a string. It's fine!
3) Find a better way to write this guard case (or, work out why the error doesn't seem to actually be an Error)

I think I'm gonna do #1

josephjclark: In fact I'm gonna do it now

josephjclark: Ok 1 doesn't work because the json longer does actually serialize. Disappointing.

josephjclark: Alright - what's going on here is that when the job, inside a SourceTextModule throws an error, we seem to get an error-like object back, not an actual Error. I think that's a detail of `node:vm` and I can't see any documentation about it - but the facts in the data are plain.

I've updated #143 with some extra notes. Basically we need to think a little bit more about how these errors get handled, propagated and reported, and this isn't the place to do it.

The runtime will now stringify this anomalous error, rather than logging it directly, which works fine if unit tests are updated a little.

packages/cli/src/cli.ts
```diff
@@ -23,6 +23,10 @@ export const cmd = yargs(hideBin(process.argv))
     description: 'Set the default log level to none, default, info or debug',
     array: true,
   })
+  .option('log-json', {
+    description: 'Output all logs as JSON objects',
+    boolean: true,
+  })
```
stuartc, line 29: We need to make a concerted effort to reflect the flag available on the CLI at least on the README for the cli package.

packages/logger/test/options.test.ts
```diff
@@ -45,3 +45,12 @@ test("don't mutate default options", (t) => {
...
 });
+
+test('json should be set to true if the env var is prsent', (t) => {
+  process.env.OPENFN_LOG_JSON = 'true';
+
+  const opts = calculateOptions({});
+  t.true(opts.json);
+
+  delete process.env.OPENFN_LOG_JSON;
+});
```
stuartc, line 56: Should this not be in the CLI? I've been thinking of `logger` strictly being a dependency.

stuartc (commented):
Great work here @josephjclark, it did take me a little while to get my head around the 'transparency' of this implementation.

I was trying to work out what the behaviour is when JSON logging is enabled and you log a string. But if my understanding is correct, the logger instance is created with a namespace already and when logging a string it emits JSON with the namespace attached right?

I'm happy to merge this but could you take a look at my comment around the logger using an env var vs the CLI explicitly being in charge of that (via a flag or env, either or)?

I can imagine some situations where your approach is useful down the line - but want to make sure I'm understanding your approach here.

packages/logger/test/options.test.ts
josephjclark: Yes, you're absolutely right

packages/cli/src/cli.ts
josephjclark: Happy to be noisier about it but I think json logging is more of a system admin thing and less a general user thing?

josephjclark:
Thanks Stu - let me sort out the env var thing, you're dead right about that

> I was trying to work out what the behaviour is when JSON logging is enabled and you log a string
```
logger.log(`hello world`) // { level: 'info', message: 'hello world' }
```

> the logger instance is created with a namespace already and when logging a string it emits JSON with the namespace attached right

:100:

packages/cli/src/cli.ts
stuartc: Yeah thats cool, when I looked that the README for the CLI I realised we need much more than just documenting the json logging thing.. so lets leave that for now and come back to do a big whack of docs on the CLI for newcomers at a later point.

josephjclark:
@stuartc All done!

PR #169, opened 2023-01-19
integration-tests: run correct test suite

josephjclark:
Whoops - integration tests were running the wrong test suite!

This PR fixes that...


josephjclark:
@stuartc FYI you were right!

Luckily the fix is simple - just a brain fart in the last commits on the feature :sweat_smile:

PR #174, opened 2023-02-08
CLI: refactor options

josephjclark:
This PR contains a re-write of the CLI's options.

At a high level, we are:
* Utilising yargs better for defaulting values
* Improving the consistency of options from the user's point of view
* Making it easier to share and re-use options between commands
* Removing the monolithic validator function
* Updating unit tests around all this. 

The big idea is that each option is defined as a stand-alone object with its own validating function:
```
export const immutable: CLIOption = () => ({
  name: 'immutable',
  yargs: {
    description: 'Treat state as immutable',
  },
  boolean: true,
  ensure: (opts) => {
    def(opts, 'immutable', false)
  },
});
```
Then the command in yargs is built from an array of these objects;
```
const options = [
  o.adaptors(),
  o.immutable()
]

const executeCommand = {
  command: 'execute [path]',
  handler: ensure('execute', options), 
  builder: build(options)
}
```
The options are are functions, which is  a bit wierd, but allows us to pass options (like is this option mandatory for this command)

## API changes

This PR does introduce a couple of user-facing changes to the API:

* `compile` now prints to stdout by default
* `compile` will always save to a path if you provide one (even if `stdout` is explicitly set to true)

So far they're all against the compile command, which so far as I know no-one is using.

## Outstanding work

The aim of this PR is just to refactor the execute command. We can do the others later.

Still needed for this:

* [x] Continue using ensure for all other commands (I've probably broken this by removing logic from ensure)
* [x] Maybe add some unit tests around `yargs.parse()`. The current unit tests only check `ensure` in isolation, which is fine, but I think an extra layer of tests is worthwhile
* [x] Finish migrating ALL `execute` commands (nearly there!)
* [x] Make sure the CLI still works properly in execute
* [ ] Could be a good opportunity to add a few more integration tests
* [ ] Probably delete integration.test.ts and rename command.test.ts -> cli.test.ts

## QA Notes

None of the logic of the CLI commands has changed in this PR - but the parsing of the command's options should consider to be changed.

So for QA we basically need to run all options against all commands and ensure nothing has changed.

That's too much effort, but I'd appreciate a quick test of each command and its major options. Make sure they still work as you expect.

Check the help pages as well - they should still be consistent.


josephjclark:
Alright, tests are finally passing! What a monstrous job this turned out to be (and it's only half done really).

I'll revisit, self review, and run some more manual QA tomorrow.. Would be nice to move on from this at this point.

josephjclark:
Updated the unit tests to be lovely and clean. Really pleased with this now.

mtuchi:
@josephjclark I have tested the following  and the work fiine `-i auto-instal, -a adaptors, monorepo-adaptors, -s path to state, -O, -o path to output`

stuartc (approved):

PR #185, opened 2023-02-24
CLI: Metadata command

josephjclark:
This PR adds a `metadata` command to the CLI, which generates metadata for a given adaptor.

## API

The basic API is:

```
openfn metadata -s path/to/config -a adaptor-name
```

Adaptors will be resolved just like normal, so you can pass the monorepo flag `-m` or an adaptor name to be loaded from the repo, or a path to a location on disk.

CAVEAT: at the time of wrtiting you have to pass a path to the index.js entry point. This is very fixable, I just want to avoid code re-use so I need to think a little bit about the solution.

Output is written to the CLI repo and the path is returned as the final argument. You can tail it to load the metadata and pipe it into other commands.

At the moment there's a simple cache in place. Config objects are hashed into a key and stored under this key in the repo.

A timestamp is added to the generated data.

## Integration Tests

There are an integration tests!

So useful here because it's hard to unit test some of this stuff. And the heavy lifting is in the adaptor anyway.

### DIFF
--- a/packages/cli/src/metadata/cache.ts
+++ b/packages/cli/src/metadata/cache.ts
@@ -1,0 +1,32 @@
+import { createHash } from 'node:crypto';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { writeFile, mkdir } from 'node:fs/promises';
+
+const getPath = (repoDir: string, key: string) => `${repoDir}/meta/${key}.json`;
+
+// TODO should we sort keys so that equivalent objects generate the same key?
+const generateKey = (config: any) =>
+  createHash('sha256').update(JSON.stringify(config)).digest('hex');
+
+const get = (repoPath: string, key: string) => {
+  try {
+    const data = readFileSync(getPath(repoPath, key));
+    const json = JSON.parse(data);
+    return json;
+  } catch (e) {
+    return null;
+  }
+};
+
+// lock the cache to prevent another process generating
+// this is very similar to how docgen works
+const lock = async (repoPath: string, key: string) => {};
+
+const set = async (repoPath: string, key: string, data: any) => {
+  const fullPath = getPath(repoPath, key);
+  await mkdir(path.dirname(fullPath), { recursive: true });
+  await writeFile(fullPath, JSON.stringify(data));
+};
+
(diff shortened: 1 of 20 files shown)


josephjclark:
I think this is done now - I'll give it a final review in the morning then hand it over

packages/cli/src/metadata/cache.ts
```diff
@@ -0,0 +1,32 @@
+import { createHash } from 'node:crypto';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+import { writeFile, mkdir } from 'node:fs/promises';
+
+const getPath = (repoDir: string, key: string) => `${repoDir}/meta/${key}.json`;
+
+// TODO should we sort keys so that equivalent objects generate the same key?
+const generateKey = (config: any) =>
+  createHash('sha256').update(JSON.stringify(config)).digest('hex');
```
stuartc, line 10: Excellent question, would we need it to be recursive as well?
I'm perhaps not that worried as the JSON.parse and the files we generate should stable right? We don't do anything to them along the way...

Thought experiment though, if we stringified it, and sorted all the characters and then hashed it? Would that be stable/equivalent?

packages/cli/src/metadata/handler.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (mod.metadata) {
+      logger.info('Metadata function found. Generating metadata...');
+      const result = await mod.metadata(config);
+      decorateMetadata(result);
+      await cache.set(repoDir, id, result);
+      finish();
+    } else {
+      logger.error('No metadata helper found');
+
+      process.exit(1); // TODO what's the correct error code?
```
stuartc, line 97: For now, not `1`... generally imo `1` is for something the user entered in wrong or unknown... Like bad/incompatible arguments.. Can you give some thought for the range of errors that we might return on the CLI, would be nice for a dedicated `--help exit-codes` output or something?

stuartc (approved):
@josephjclark great work here! Really thorough. My only question is, where is (if there is one) some definition/examples/docs around metadata in general? Not that I expect it to exist right now - but we will need something to tie the CLIs usage and what these functions will return.

PR #187, opened 2023-03-09
describe-package: recognise magic functions

josephjclark:
To address https://github.com/OpenFn/Lightning/issues/657 - enable adaptor docs to indicate which functions support :sparkles: magic :sparkles: - we need to add a little more information to the package description returned by `describe-package`.

Phew, that's a mouthful.

Basically when `describe-package` generates a json object to describe a function, it includes a `magic: false` flag. Now that we have two published adaptors with magic annotations, we sometimes want to set `magic: true`.

This PR does that an unblocks Lightning#657.

Describe package really needs a couple of days of love on it - but luckily this is a trivial implementation and I think we can rush it in.

I have committed version bumps so we  can release directly from this branch.

### DIFF
--- a/packages/describe-package/src/describe-project.ts
+++ b/packages/describe-package/src/describe-project.ts
@@ -32,11 +32,12 @@ const describeFunction = (
     // @ts-ignore symbol.parent
     [parent] = symbol.symbol.parent.escapedName.match(/(language-\w+)/);
   }
+
   return {
     name: moduleName ? `${moduleName}.${symbol.name}` : symbol.name,
     description: symbol.comment,
     parameters: symbol.parameters.map((p) => describeParameter(project, p)),
-    magic: false,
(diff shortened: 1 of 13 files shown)


packages/describe-package/src/describe-project.ts
```diff
@@ -32,11 +32,12 @@ const describeFunction = (
     // @ts-ignore symbol.parent
     [parent] = symbol.symbol.parent.escapedName.match(/(language-\w+)/);
   }
+
   return {
     name: moduleName ? `${moduleName}.${symbol.name}` : symbol.name,
     description: symbol.comment,
     parameters: symbol.parameters.map((p) => describeParameter(project, p)),
-    magic: false,
+    magic: symbol.jsDocTags.some((tag) => tag.tagName.escapedText === 'magic'),
```
josephjclark, line 40: This is the only actual code diff in this PR

stuartc (approved):

PR #198, opened 2023-03-22
Include timestamp in json logs

### LINKED ISSUE #181: Logger: include a timestamp
When logging to JSON, include a timestamp on every log object.

I don't think we should include a timestamp in text logging because it's just in the way.

josephjclark:
Simple feature which logs a timestamp as epoch-ms with Date.now().

Feels like a useful thing to track as a first-class property in the log object, and doesn't hurt anyone.

Timestamps are not included in stdout logs.

## Related issue

Fixes #181

### DIFF
--- a/.changeset/late-vans-divide.md
+++ b/.changeset/late-vans-divide.md
@@ -1,0 +1,5 @@
+---
+'@openfn/logger': patch
+---
+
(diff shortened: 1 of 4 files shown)


stuartc (approved):
:+1:

Issue #204, opened 2023-03-29
Compiler: cannot load common from monorepo


josephjclark:
When running adaptors out of the monorepo, the compiler will fail to resolve imports from `common`.

This is a direct result of  #196 and a reminder to add better integration tests, which would have caught this quickly.

As a result, monorepo support is basically broken right now

josephjclark:
The CLI knows about the monorepo. But the compiler, which does the lookup, does not.

A quick fix would be to support `../common` as a path to the common adaptor and just try it. Not ideal, and `../common` could point to anything really.

This might be a call to address #197. It's certainly an input to it, and actually I'll go and make a note over there.

josephjclark:
As a workaround, you can explicitly import functions from common in your job code:
```
import { fn } from '@openfn/language-common';
```

josephjclark:
This is proving to be quite a big problem so I'm going go to ahead with a quick fix. I still think this problem should eventually be abstracted into `describe-package`

PR #205, opened 2023-03-31
Fix monorepo

josephjclark:
This PR should fix monorepo support when loading common dependencies.

This is a quick fix to #204 which actually ends up feeling quite natural.

What's going on is: when we compile code, we try to autogenerate an import statement for it. To do this, we need to know which adaptor is in use and what that adaptor's exports are. This is the CLI's job.

If the adaptor also exports things from common, we also need to lookup common's dependencies. And were is common? If we're loading an adaptor from the monorepo, the lookup is different  - and this PR just adds a branch which uses a different path if loading from the monorepo.

Side node: in the process of testing this I discovered an issue in describe-package which will blow up if it tries to parse a file with no exports.  I've added a fix.

This is quite an important fix so I suggest we release straight from here.

Versions have been bumped.

### DIFF
--- a/.changeset/honest-cups-work.md
+++ b/.changeset/honest-cups-work.md
@@ -1,5 +1,0 @@
----
-'@openfn/describe-package': patch
----
-
-Handle dts files with no exports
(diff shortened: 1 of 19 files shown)


josephjclark:
@stuartc are you able to take a quick look at this and release?

stuartc (approved):

Issue #207, opened 2023-04-04
Runtime: always return a state object (which may have an error key)


josephjclark:
We need to give the runtime a nice simple contract:

Run THIS job/workflow and return the result as a state object.

This is particularly important in the case of error handling: any errors/exceptions need to be caught and returned as a state object.

We are hardening the concept of the state object as the contract between the runtime and its callers.

In the case of an exception, the only returned state will be an error key.

So our state  definition now looks a bit like this:
```
type State = {
  data?: object
  configuration?: object
  error?: {
     code: number; // should align to old exit codes
     message: string; // human readable error
     exception?: string; // serialized error object with stack trace
     source: 'cli' | 'compiler' | 'runtime' | 'job'; // might be interesting to try and trap this
  }
}
```
The CLI and runtime manager will both reflect this behaviour. If there's an error, the CLI should write an error state to the output.json (and print the code and message to the console). The runtime manager should report the Workflow Attempt as completed as save the error state. Lightning should recognise the error key and respond appropriately.

Note that if there is an error key, there should never be `data` or` configuration` or anything else.

This is an argument in favour of strict mode by the way. The error key will be interpreted by Lightning as a failure, so users/adaptors who write their own `state.error` will not trigger Lightning behaviours. In strict mode, their `error` key would be ignored. Note that `state.data.error` would still be passed to downstream jobs.

This is a major input to the new runtime manager work.

josephjclark:
So this may not work because in platform, the `error` key contains the logs of the previous job.

That means that legacy jobs migrated to the new platform would fail because we're handling errors differently.

We may have to do the new error handling in an `err` key, or `result.error` or something

PR #208, opened 2023-04-05
 Workflow Support in the Runtime

### LINKED ISSUE #202: Runtime: accept an execution plan as input
The runtime should accept as an input, instead of a string expression, an Execution Plan (or XPlan) as a JSON object.

An XPlan is what we informally call a Workflow. But we can't really call it a workflow because that means something specific to Lightning, and calling it an Attempt doesn't make sense in runtime world.

An XPlan is (probably) an array of jobs, defined like this:
```
type XPlanJob = {
   id: JobId;                // string
   expression: string;  // the actual job code
   credential: object;   // state.configuration
   adaptor: string;       // @openfn/language-common@1.0.0
   data: object;            // state.data (or maybe just Omit<State, 'configuration'>)
   next: { onSuccess: JobId, onError: JobId } | JobId
}
```
The idea simply is that the runtime executes the jobs in order, sequentially. A job can specifiy where to go next via onSuccess/onError "pointers", or this can be inferred by array order.

The state from the previous job, if any, should always be passed in. A job may want to provide its own initialisation state from the trigger (In practice I think this is only on the first node?).

Credentials at this level must be provided in-line in the XPlan JSON. The Runtime Manager should implement some kind of secure credential loading solution.

By pushing down the workflow pipeline logic into the Runtime, all Runtime Managers benefit from it. A Runtime Manager implementation (remember the CLI is a runtime manager too!) can then pass an XPlan or job string into the runtime, which will go off and execute and then return the final state to the caller.

One implication of this is that in a multi threaded RTM environment, each workflow will be executed in a single thread. If this ever becomes a problem, the workaround would be for the RTM to break up the XPlan into smaller pieces and feed them off into several threads. But unless workflows ever develop parallel jobs I don't see this being something to worry about.

### LINKED ISSUE #213: Runtime: lock down environment
I am not sure how strict we need to be on this - but the default answer should probably be "very strict".

We should ensure that when an expression runs in the runtime's VM, it cannot manipulate the environment around it.

That means the job should not be able to:
* Write data to the global scope
* Write data to global objects (like `console`, `parseInt`, `setTimeout`)
* Write data to adaptors (ie `fn.myMaliciousThing = 66`)

These writes are all transient and only affect the containing context. So when running a job, the writes only affect that job and are lost when the process closes.

In a workflow, these environment writes can pass data from one job to another in the same workflow. That's naughty, because you should use state for that. But it probably doesn't matter because a workflow was created by one user, so why would you want to hack your own job? But the recent addition of workflows has at least opened up the problem space here.

In a multi-threaded environment in the runtime manager, it _might_ be possible to write the environment of one job and read it back in another, unrelated, job running in the same process. I that because we're building a context for each job in that process, we're actually  securing against this. You'd have to break out of the sandbox first, and THEN you'd be able to read from the globals.

So while I feel like we ought to do this, I don't feel like it's a big priority.

josephjclark:
The Runtime now supports execution of Workflows or Execution Plans.

Closes #202

Instead of passing an expression, you can pass an Execution Plan. This is basically a list of jobs and some rules as to which execute when. Expressions can still be passed as before (they'll actually be wrapped inside an execution plan internally).
```
openfn workflow.json
```

The Execution Plan looks like this:

```
const plan = {
    start: 'a',
    jobs: [{
      id: 'a',
      adaptor: "common",
      state: { /* globals */ },
      configuration:  { /* credentials */ },
      expression: 'export default [(s) => { s.data.a = true; return s}]',
      next: {
          b: {
            condition: 'state.data.count > 0',
          },
          c: {
            condition: 'state.data.count == 0',
          },
        },
      },
      {
       id: 'b',
       adaptor: "common",
       configuration:  { /* credentials */ },
       expression: 'path/to/some/expression.js', // useful for local dev
      },
    {
       id: "c",        
      adaptor: "common",
        expression: 'export default [(s) => { s.data.c = true; return s}]',
      },
    ],
  };
```

Some things to note:

* A plan can have a precondition, which is evaluated (in a sandbox) against the initial state. If false, the workflow will not run (initial state will be returned).
* A job can have edges to any number of other jobs. Each edge can be `true` (upon completion of the job (including if it errors), execute this job next), or an object with a condition (a sandboxed expression evaluated against the output state of that job).
* Before each job is executed, the state is assembled from the state of the previous step (or initial state), the state defined on the job, and credential information passed to the job.
* The runtime itself doesn't use the adaptor information. That's used by the CLI to tell the compiler how to generate the import statement.

There are some constraints on Workflows:
* Right now, a workflow cannot use two versions of the same adaptor
* A workflow cannot have a circular reference (a -> b -> a)
* A workflow cannot have two inputs (upstream jobs that a job depends on).

These last two cause us headaches in parallelism. We can add support for this, but since the workflow diagram probably doesn't support it, and since the handling of state in a parallel workflow is a bit of a headache, it seems sensible to cut cloth for now.

If the execution plan is deemed invalid, the workflow will return an error state object.

There's a whole bunch of refactoring going on to support this work.

Still TO DO:

* [x] Some typescriptery
* [ ] ~~Consistent handling of error states - see #207, which I am likely to do off-branch~~
* [x] ~~Add something like an "acceptError" flag on an edge, because I think it's a very common use case that a job should only execute if the previous job was successful (or the inverse). This is enables a nice shorthand for on-success and on-error edges.~~ I think this is a lightning UI problem actually, because an on-success condition is just !state.error
* [x] Changeset etc
* [x] Update the CLI in tandem
* [x]  State from the previous job should override data defined on a workflow (globals). This means workflow state is a default and can be overridden by input or job logic.

## Other changes

* The runtime will now enforce that every job returns serializable state. That means circular references are broken with a `"[Circular]"` string and functions are removed. This uses `fast-safe-stringify`. We used to do this in the CLI but for workflows to work, the runtime needs to do it (and actually this just makes more sense)
* I've tightened up the  global scope so you can't write to eg `console`, which fixes #213. I don't think this ACTUALLY matters, given the encapsulation of each workflow, but it does feel more robust
* A pass over documentation/readmes

## QA

To test this out, run `pnpm install:global` from root. This will install a global command called `openfnx` from this repo, which doesn't interfere with your production install.

## Breaking change

There is a tiny breaking change here but I doubt anyone cares

You used to be able to pass any old state object. A string, number, ` { data: 22 }`, anything you want and the runtime would just pass it through.

In the new runtime we are doing a bit more management of the state object - for example, we merge state for each job in the plan. This means its harder to accept arbitrary input. Really, the runtime manager now depends on state being object with a few keys. This may get stricter if we start to only return `state.data` from a job,

I don't think it's worth supporting arbitrary state because the system gets a bit less predictable. The only use case is testing, and as you can see in this PR I've just had to be a be stricter with state to pass against the new runtime. No big deal.


josephjclark:
Proposed changes based on this morning's conversation:

* No start node (but you can pass a start node id an argument)
* `nodes` is an array of nodes, where we run the first node first (unless a start is provided externally)
* Explicit node/edge naming (kinda hate this)
* Note that well need to support no-op nodes, with no expression. This is  basically what lightning would call a trigger.
* Version number (should this be `schema: "url"` ?)

```
version: 1,
nodes: [
   { id: 'a', edges: { 'b': true } },  // trigger style node
   { id: 'b', expression: 'state => state' }
]
```

Big fan of dropping the special `start` edge, but not really keen on the terminology changes. I think that a `jobs` array of nodes and a `next` object of links is semantically richer.

After executing a node, we will NOT try to execute the next node in the array. We should be strict on this because it only leads to confusion. We only ever follow edge defined on the node, and if there are not, we return.

A node id is actually optional. In practice all Lightning workflows will have ids, but for local dev a singleton workflow doesn't need an id at all. This is not a big deal but convenient for testing and local dev. Unless there's a reason to be strict?

A plan may have an id string, but the runtime doesn't care much about it. Maybe for reporting. But that stuff  is probably handed by the RTM server.

josephjclark:
Additional note @taylordowns2000, re:
```
configuration: "https://app.openfn.org/api/<< SOME_URI_TO_A_CREDENTIAL >>",
```
The runtime itself will NOT support this*. A credential must be a JSON object.

But the Runtime Manager Server WILL support a string in an Attempt it receives from Lightning. The server is responsible for fetching credentials and state and feeding them into the runtime.

* (Well, probably. Maybe the runtime will be given a callback to lazily resolve state and config objects, so the runtime would have some understanding of a string, but the handling of this will still be done by the server (or CLI)).

josephjclark:
Some issues raised by Mtutchi:

* You can pass paths in `job.expression`, which wasn't my intent. Is this desirable? Confusing? What is the path relative to? The workflow file I guess. Lightning won't use this, but you can see how it would be useful in local dev (and not that Mtuchi instinctively tried to do this). 
* #219

josephjclark:
Confirmation of the redesigned structure:
```
{
  id: "bf371a", // optional/irrelevant
  version: 1, // probably be a schema url
  start :'a', // optional, defaults to jobs[0] (lightning needs this). No longer an edge.
  jobs: [
     { id: 'a', next: { 'b': true } },  // trigger style node with no expression
     { id: 'b', expression: 'state => state' }
  ]
}
```
Where job ids are technically optional but as soon as you have a reference to a job in the plan, you need an id for it.

taylordowns2000 (commented):
@josephjclark , this is ready to go from my end but i'm holding off because i didn't your discussion of "scribbling" between jobs. ping me to discuss super quick, any time today, and we'll get this approved/merged.

josephjclark:
Two things coming in:
1) Ability to load config from a path (the CLI must support this, the runtime must not)
2) Ability to pass start into the CLI (whoops I forgot about this)

josephjclark:
@taylordowns2000 This is ready to go :+1:

PR #231, opened 2023-04-26
Fix strict output

### LINKED ISSUE #225: Runtime: Workflows only forward state.data, not state
My state processing in between jobs in a workflow is only passing through the data object, but really everything on the state needs to be passed through.

Whoops, I am being overly dogmatic in my state assembly between jobs. 

Also consider strict output. I think that's a CLI think, not a workflow thing. So internally all the state will be passed to the different jobs (meaning job b sees the request of job b), but the CLI only returns state.data.

That gives us better compatibility for now, and if we don't like it we can add strictness in the runtime later.

josephjclark:
This PR fixes strict output in the runtime.

Closes #225 

The runtime now has a strict mode. This controls what state gets passed between jobs in a workflow.

In strict mode, only `{ data, references, error }` gets passed between jobs.

In non-strict mode (legacy mode, needed for platform compatibility), everything gets passed onwards.

In addition to these changes, the CLI command has been renamed `--no-strict`. Passing `--no-strict-output` will still work, but the option has been deprecated.

Note that the CLI still does some strict mode enforcement, stripping away things that the runtime allows, like `references` and `configuration`.

## Notes & Questions

* Are there any other properties other than `references` and `error` we should preserve in strict mode?
* In non-strict mode, should we delete state.configuration at the end of each job?


josephjclark:
@mtuchi can you test your workflow against this?

You should be able to write `patients` directly to state, instead doing of `data.patients`, so long as `--strict-output` is set.

I am dealing with the other issues seperately.

taylordowns2000:
Tested this AM and it works like a charm. Thanks @josephjclark . For the Qs:
1. Those seem good for now, but because adaptors have so much freedom I can't guarantee that this will always work. Of the top of my head though, I bet it covers 85% of the use cases.
2. That's what happens in OpenFn.org v1 - I'd actually envision this being a really useful convention to maintain because it would help devs from accidentally exposing creds while sharing an `output.json` for collaborative troubleshooting

Issue #234, opened 2023-04-27
Runtime: Log job start


josephjclark:
Whoops, the runtime doesn't log when a job starts. Oversight on my part!

PR #235, opened 2023-04-27
Log job name 

josephjclark:
image

There's a few fixes in here:

* #234 - the main one. Log to info when a job starts and success when it ends (including a duration). Demote operation start/end/duration down to debug
* 57 - add a new log level `always`, which is like success but doesn't include the tick. In the end I didn't actually use this :sweat_smile: But it's worth merging
* Improve the `compile job` output from the CLI - which is a bit meaningless in a workflow - to `compilation complete"

This is all low priority but I wanted to get it out of the way before dealing with the error stuff.

Here's the always log output
image


josephjclark:
Bugger, I need to sort out that conflict...

PR #239, opened 2023-05-02
cant read length from empty array

taylordowns2000:
@josephjclark , this is a funny one. I ran into this with Mtuchi recently, when trying to read the length of an empty array that was passed in via initial state.

Do you know why this test is failing?


taylordowns2000:
handled by strict mode

Issue #233, opened 2023-04-27
Strict mode is wrong


josephjclark:
I learned this morning that I've fundamentally misunderstood the handling and usage of `state.data`, and as a result, strict mode is wrong.

I thought:

The runtime will manage various bits of state (`config`, `references`), but `data` is for your application data. `state.data` is the thing you want forwarded on to the next job, it's the thing you want in your final result.

I also knew that adaptors like `http` will scribble on `data` (ie, the response). Which I never liked because it erodes the idea that `data` is a safe space for the user.

But all this is incorrect. In fact the opposite is true:

The runtime will manage the `configuration`, `references` and `data` properties of `state`. The rest of the object is left alone for the user to do what they want with.

Adaptors don't modify `state.data` , they _replace_ it (via `composeNextState`)

The idea of strict mode was to establish new, possibly breaking, conventions for the v2 runtime. This includes encouraging users to lean more heavily on `state.data` and take ownership of it, but it turns out that doesn't make any sense.

## Options

* Remove strict mode entirely (this may be a wise short-term measure)
* Keep strict mode but change its meaning. Maybe strict mode returns state apart from `configuration`, `references`, `data` - because that's all  transient stuff needed for the job.

I kinda wish we had a property of state which represents the result, the final or persistent state, the thing that SHOULD get passed to the next job. A part of state which we promise we won't touch or interfere with, since we're a bit loosey goosey about a small number of keys.

Some possible rules for strict mode, if we want to keep it:

* At the end of a job, `configuration`, `data` and ` references` are removed (are there any other managed props?)
* * Removing data is kind of a big deal, but at least is forces you to process response data in that job and prepare it for passing down the line.
* * It would also mean adaptors don't have to tidy up after themselves
* All properties beginning with `_` are removed (including on nested structures)
* Introduce a new property on state, maybe `result` or `final` or `persist`, which is the only thing that gets returned or forwarded.

taylordowns2000:
@josephjclark , I don't know if this helps, but I _like_ the fuzziness of non-strict-mode because it accounts for two major use cases: (1) where `state.data` does exactly what we (I _think_ you too) want, and (2) where it doesn't.

## Thing 1
Consider this common pattern where each operation is from a different adaptor. (i'll use namespaced operations—which aren't a thing) to indicate the adaptor)
```js
OpenMRS.get('/patients/007');
DHIS2.post('/trackedEntityInstances', dataValue('patient')); 
```
[...]

PR #240, opened 2023-05-03
Non-strict mode by default

josephjclark:
I've had enough of strict mode and the confusion it brings! 

It's also fundamentally wrong, see #233.

While we work out what to do, this PR simply runs in non-strict mode by default, which makes everyone's life a bit easier.


taylordowns2000 (approved):
approved, once tests are passing

PR #247, opened 2023-05-19
Ensure jobs only receive upstream state

### LINKED ISSUE #227: Runtime: Jobs should receive the state of their upstream job (?)
Here's an interesting one:

```
   a
 /   \
x     y

jobs: [
 { id: 'a', expression: '...', next: { x true, y: true },
 { id: 'x', expression: '...' }, 
 { id: 'y', expression: '...' }
]
```

In this workflow, `a` will generate some state and pass it down to `x` and `y`.

But what happens in the workflow right now is:
* Execute `a` with initial state `s`
* Save the state
* Execute `x` with the latest state (`s'1`)
* Save the state
* Execute `y` with  the latest state (`s'2'`)
 
There are a few consequences of this:

* If `x` returns no state, `y` will receive no state as input. This isn't unreasonable from `x`'s point of view as its a leaf node, it has no dependencies
* If `x` happens to overwrite or remove some state that `y` depends on, `y` will likely fail.

This means we stop thinking about a single state that is shared by every job to return a final answer, and start thinking of state as more isolated thing which is passed down various independent streams.

## Solution

The runtime needs to save the state produced by each job. When running each new job, it needs to look at the upstream job, find its state, and pass it in.

If there are multiple upstream jobs (not possible right now) the states must be merged.

## Final state

But what does this mean for the final state returned by the workflow? It has two end points, two results, two final states.

Is it an aggregation of all the state objects merged together? If there are conflicting keys, who wins? Should we deep merge nested objects? Nested arrays?

Is it just the thing returned by the last job?

Or do we return an index of the states returned by each job?

josephjclark:
At the moment, when a job runs in a workflow, we pass the _last_ state into it (merging in credentials and defaults of course).

This PR ensures that we only pass the state from the upstream job into the next one.

Closes #227

```
  a
 /   \
j     k
     / \
    x  y
```

In the workflow above, it's important that:
* `j` and `k` receive  immutable copies of the result of `a`
* `x` and `y` receive immutable copies of the state of `k`

As things stand at the moment it's a bit random. Because we just pass state as a pipeline, it'll be something like `k` will get the result of `j`; `x` will get the result of `k`, and `y` will get the result of `x`.

## Changes

Some changes introduced in this PR:

* Workflows with branches will now return multiple results
* The `configuration` key is now removed from the result state. We've discussed this but not actually done it (actually I think the CLI does something along these lines, but since workflows, the runtime needs to be a bit stricter about it).

## Final States

The runtime is designed to return a state object as the "result" of the workflow or expression. Up until now that's been straightforward - we just return the last state produced.

But if you have a branching workflow, what is the final state?
```
  a
 /   \
x     y
```
If `a`, `x`, and `y ` all return a state object, what should the result of the run look like?

This PR implements the following rules:

* If there is one one "leaf" node - one node which doesn't go anywhere else - we just return the final state
* If there are multiple leaf nodes, or branches, we return an object with each state as a value under the key of a job id.

So for the use-case above ,we probably return an object like this:
```
{
   x: { data: { ... } },
   y: { data: { ... } },
}
```
Note that errors will be written to the relevant nested state object. So if `x` and `y` both fail, you'll get:
```
{
   x: { data: { ... }, errors: [] },
   y: { data: { ... }, errors: []},
}
```
If a `a` fails but `x` and `y` both run, you'll get the same structure, with the error from `a` will be duplicated into both downstream states.

Also note that branches are dynamic - a branch is only a branch if two downstream jobs are created at runtime. So if we have a workflow like this:
```
        a
    /         \
error       success
 /                \
x                  y
```
Only x OR y will run, so a single result state will be returned.

In other words, we're only sensitive to ACTUAL branches, not theoretical ones.

## Future Work

This isn't the end of the story for handling multiple return states. It's just a first pass.

I don't like that the final result object is a bit inpredictable. Will I get a state or an object of states? It's actually made a bit worse by dynamics, because the same workflow with different inputs or at different times could return different data structures.

At the moment, with workflows so new, we don't really have any  real-life cases to test against (and no legacy to break). So it's a safe point to try something.

And in practice, for a workflow which branches like this, there probably isn't a desire to return a result state. The branch is because state is being pushed to two different systems and the result is immaterial.

I think we'll want to come back and revisit this in a  couple of months and add some kind of option. Maybe you want to force all your state to be squashed into a final result object (this is fine if your jobs use different slices of state).  Maybe you only care about the result of a single job and you want to mark it as  "result" or "final" node. Maybe we need to introduce an aggregator node, as discussed in #227 .

### DIFF
--- a/.changeset/mighty-timers-agree.md
+++ b/.changeset/mighty-timers-agree.md
@@ -1,0 +1,5 @@
+---
+'@openfn/runtime': patch
+---
+
+Ensure jobs only receive direct upstream state
--- a/integration-tests/cli/test/execute-workflow.test.ts
+++ b/integration-tests/cli/test/execute-workflow.test.ts
@@ -24,7 +24,8 @@ test.afterEach(async () => {
 
 // Autoinstall adaptors
 test.serial(`openfn ${jobsPath}/wf-count.json -i`, async (t) => {
-  await run(t.title);
+  const { err } = await run(t.title);
+  t.falsy(err);
 
   const out = getJSON();
   t.is(out.data.count, 42);
@@ -34,7 +35,8 @@ test.serial(`openfn ${jobsPath}/wf-count.json -i`, async (t) => {
 test.serial(
   `openfn ${jobsPath}/wf-count.json -S "{ \\"data\\": { \\"count\\": 6 } }"`,
   async (t) => {
-    await run(t.title);
+    const { err } = await run(t.title);
+    t.falsy(err);
 
     const out = getJSON();
     t.is(out.data.count, 12);
@@ -43,7 +45,8 @@ test.serial(
 
 // Multiple steps, shared array
 test.serial(`openfn ${jobsPath}/wf-array.json`, async (t) => {
-  await run(t.title);
+  const { err } = await run(t.title);
+  t.falsy(err);
 
   const out = getJSON();
   t.is(out.data.items.length, 2);
@@ -55,7 +58,8 @@ test.serial(`openfn ${jobsPath}/wf-array.json`, async (t) => {
 test.serial(
   `openfn ${jobsPath}/wf-array.json -S "{ \\"data\\": { \\"items\\": [\\"z\\"] } }"`,
   async (t) => {
-    await run(t.title);
+    const { err } = await run(t.title);
+    t.falsy(err);
 
     const out = getJSON();
     t.is(out.data.items.length, 3);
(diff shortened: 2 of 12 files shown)


stuartc (approved):
This looks good for now, I finally understand the conundrum you mentioned the other day.

I think this may be flushed out (or rather only be a consideration for local execution) as each run would get get it's output set on the backend.

Issue #115, opened 2022-11-30
CLI: ensure-opts needs a refactor


josephjclark:
There's a utility file in the cli at `util/ensure-opts`. The idea is you pass in the options object generated by yargs and it'll set up all the defaults and stuff to enable to the command handlers to run.

The problem is that this was written with just the `execute` command in mind. Now that we have commands for `compile docs docgen test repo` it's starting to creak a bit. The handling of defaults depends on which command is run, so really each command needs its own bespoke ensure-opts function.

Similarly, the Opts type is really designed for a single command and need to be broken up to support all the different commands the CLI now takes.

None of this is urgent. It'll likely get dealt with in some other piece of work. I just want to track it here.

josephjclark:
I have made a big step towards this in #174  - but it turns out to be a massive job, so I'm only about half way.

We have a much nicer pattern now for `execute` and `compille`. Other commands need migrating to the new syntax, and `ensure-opts` eventually need to be removed entirely.

One outstanding sticking point is the typing of `log`. The incoming `log` from the command line is an array of strings which are log options. This needs to be converted into an object to be passed into the logger (including the `json` flag). RIght now we mutate `log` which is horrible - we need to create a `logOptions` object which has all the parsed stuff in it, and that can then be forwarded to the logger. This is a fair chunk of work but will clean up a bunch of type issues,.

The other oustanding thing, I expect, is the `Opts` type, which needs updating  to reflect the new option obejcts and remove some of the conradictory typings (like the way `adaptor` and `adaptors` is used).

josephjclark:
So what actually needs doing now?

I've already re-written the `execute`, `compile`, `test` and `metadata` commands. The `versrion` command doesn't use ensure-opts.

That means the only affected commands are: `repo`, 'docs' and maybe 'docgen`

* The log options need to be refactored (see #190, which we should fix at the same time)
* Tweak the repo command (#62) and make sure it doesn't depend on ensure-opts. I think this is trivial.
* Check that docs docgen doesn't actually depend on ensure-opts at all, as I suspect they don't.

Then ensure-opts and all related tests can be removed.

So really I think this is: fix log (#190) and repo (#62), refactoring into the new style at the same time. Then delete ensure-opts and see what breaks.

josephjclark:
There's a bit of confusion about where the log and log-json commands are registered.

I think they should be registered at the top, and each sub command should not re-declare them. We should have a common type called LogCommands or something that all other commands extends.

PR #248, opened 2023-05-19
Repo Command Refactor

### LINKED ISSUE #62: CLI: clean up repo pwd
Raised out of #56

`openfn repo pwd` isn't very clear.

Instead, let's just do:

`openfn repo` - print out diagnostic information (including the repo working dir)

Also, this needs to report `defaultRepoPath` if the env var isn't set, so it never returns undefined.

josephjclark:
This PR updates the `openfn repo` command, refactoring it to use the new options format, removing `pwd` and making `list` the default.

So `openfn repo` will now do what `openfn repo list` used to do (which includes printing the working directory).

I've added a bunch of new unit tests and added a new pattern of tests against yargs option parsing (I haven't done many of these, I just wanted to migrate some old coverage and setup the pattern).

Changes:

* [x] Add integration tests for the repo
* [x] Restructure to use new options format
* [x] Simplify (remove list and pwd, `openfn repo` should just list, including the pwd). Maybe keep list as an alias. 
* [x] Rethink some of the repo tests  in `cli/test/commands.test.ts` - ideally make sure they're covered by other tests, and remove

Closes #62. Part of #115 too.


josephjclark:
Phew, this ended up taking a lot longer than I expected.

Happy with the unit tests and the refactor generally. I've set up a useful pattern which we MAY want to apply to other tests, enabling us to test option parsing in standalone (a lot like what ensure-opts tests used to do, but if we refactor those to this this format we'll have a much better test suite).

stuartc (approved):

PR #249, opened 2023-05-24
CLI Release 0.0.40

josephjclark:
Release for CLI 0.0.40 and dependencies:

This contains:

* State gets properly passed to downstream jobs #247 
* CLI doc fixes #243
* Internal refactor of the repo command (`openfn repo pwd` is removed; `openfn repo` is now an alias for `openfn repo list`)

It would be great to get a bit of basic QA on the new state handling stuff.

Remember to install `openfn` by running `pnpm install:global`.

Versions have been bumped here, so you should be able to install, build and publish from this branch.

### DIFF
--- a/.changeset/beige-hats-tease.md
+++ b/.changeset/beige-hats-tease.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Refactor repo command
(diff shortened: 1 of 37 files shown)


taylordowns2000 (approved):
have tested locally with both the workflow interface and individual jobs. looks good

PR #256, opened 2023-05-31
Ensure workflows use the monorepo

josephjclark:
Turns out that if you use workflows with the monorepo, the adaptors in the workflow don't  get mapped properly.

There's a utility function which look at the `options.adaptors` array and map each adaptor to a path in the monorepo.

This PR updates that function to also look at a workflow object.

To support this I've refactored the function a bit.

I've bumped versions: I suggest after a bit of QA we release this straight off this branch.

### DIFF
--- a/.changeset/little-doors-do.md
+++ b/.changeset/little-doors-do.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Ensure workflows can use the monorepo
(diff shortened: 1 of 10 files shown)


josephjclark:
Hey @mtuchi, the monorepo should work with workflows now. Can you check it from this branch?

You can now do `pnpm install:openfnx`, which I think is easier to remember.

mtuchi:
@josephjclark i have test the fix, it work perfect 👌🏽

taylordowns2000 (approved):

PR #283, opened 2023-06-30
Tidy up deploy, work with new lightning APIs

taylordowns2000:
Added description and cron_expression to `cli deploy`

- [x] Taylor has QA'd this feature


packages/deploy/src/stateTransform.ts
```diff
@@ -87,6 +87,9 @@ function mergeTriggers(stateTriggers, specTriggers): WorkflowState['triggers'] {
             {
               id: crypto.randomUUID(),
               ...pickKeys(specTrigger, ['type']),
```
josephjclark, line 89: Instead of all that spreading, you could do this:
```
{
  id: crypto.randomUUID(),
  ...pickKeys(specTrigger, ['type', specTrigger.type === 'cron' && 'cron_expression'].filter(Boolean)),
}
```

PR #290, opened 2023-07-10
Initial pass add adding describe option to deploy

### LINKED ISSUE #278: Add "--describe" to CLI
Add a `--describe` function to the `openfn deploy` CLI, which given a configuration, returns a `project.yaml` and `state.yaml` for the requested project. 

We'll be using this for the "automated commits to github" story so that changes made through the UI in Lightning can be committed to github. 

## More context

1. you’ve just built an API in lightning that returns a project.yaml. Make it possible to call this API using an API-TOKEN
    - (That may require a PR against Lightning _before_ a PR against Kit.)
2. there’s an existing API in lightning that already returns a .state.json, and it’s already accessible via API-TOKEN
3. once you’re happy with how #1 and #2 work, extend the CLI with a function that, when given a configuration.json and (somehow…) a projectUuid (via cli argument maybe? via config?) overwrites whatever is in the projectPath and statePath with the results of #1 and #2 from Ligthning.

## Another way to think of it

1. this is being used by a local developer to _DISCOVER_ changes to a remote project.
2. Imagine you’re in a git repo.
3. You run `openfn deploy --describe .c config.json`
4. You run `git diff`
5. You see how the project has been changed remotely and decide whether or not to commit or discard those changes to your repo.

zacck:
Adds a describe flag for the deploy command so we can sync remote project.yaml to current directory

## Related issue

Fixes #278 

### DIFF
--- a/packages/deploy/src/client.ts
+++ b/packages/deploy/src/client.ts
@@ -1,5 +1,10 @@
 import { DeployConfig, ProjectPayload } from './types';
 import { DeployError } from './deployError';
+import { Readable }  from 'stream';
+import { finished } from 'stream/promises';
+import path from 'path';
+import fs from 'fs';
+
 
 export async function getProject(
   config: DeployConfig,
@@ -36,6 +41,30 @@ export async function getProject(
     throw error;
   }
 }
+
+export async function downloadSpec(
+  config: DeployConfig,
+  projectId: string
+): Promise<{ data: ProjectPayload | null }> {
+  let base = new URL(config.endpoint).origin; 
+  let url = new URL(base + `/download/yaml`);
+  url.searchParams.set('id', projectId);
+
+
+  try {
+      const res = await fetch(url);
+      const destination = path.resolve(".", config.specPath);
+      const fileStream = fs.createWriteStream(destination, {flags: 'wx'});
+      const doc = await finished(Readable.fromWeb(res.body).pipe(fileStream));
+      return doc;
+  } catch (error: any) {
+    handleCommonErrors(config, error);
+
+    throw error;
+  }
+}
+
+
 
 export async function deployProject(
   config: DeployConfig,
--- a/packages/deploy/src/client.ts
+++ b/packages/deploy/src/client.ts
@@ -3,7 +3,7 @@ import { DeployError } from './deployError';
 import { Readable }  from 'stream';
 import { finished } from 'stream/promises';
 import path from 'path';
-import fs from 'fs';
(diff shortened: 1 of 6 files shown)


packages/deploy/src/client.ts
```diff
@@ -37,6 +42,30 @@ export async function getProject(
   }
 }
 
+export async function downloadSpec(
+  config: DeployConfig,
+  projectId: string
+): Promise<{ data: ProjectPayload | null }> {
+  let base = new URL(config.endpoint).origin; 
+  let url = new URL(base + `/download/yaml`);
+  url.searchParams.set('id', projectId);
```
josephjclark, line 51: I think this would be cleaner as:
```
const url = new URL(`/download/yaml?id=${projectId}`, config.endpoint)
```
This will construct the URL properly for you from the endpoint (the signature is `new URL(path, base)`) and set the URL query properly.

Side note, as we're in control of this URL, shouldn't it be more like `/download/:projectId` with an `Accept: application/yaml` header? Defaulting to yaml?

packages/deploy/src/client.ts
```diff
@@ -37,6 +42,30 @@ export async function getProject(
...
+
+
+  try {
+      const res = await fetch(url);
+      const destination = path.resolve(".", config.specPath);
```
josephjclark, line 56: You shouldn't need that leading `.`, as `resolve` will resolve from the current working directory anyway. I'd also condense these lines, like so:
```
const fileStream = fs.createWriteStream(path.resolve(config.specPath), {flags: 'wx'});
```

packages/deploy/src/client.ts
```diff
@@ -37,6 +42,30 @@ export async function getProject(
...
+      const fileStream = fs.createWriteStream(destination, {flags: 'wx'});
```
josephjclark, line 57: I've not tested this but I'm pretty sure you can stream the reponse to file like this:
```
// import like this
import fs from 'node:fs/promises';

// back inside downloadSpec
const res = await fetch(url);
return fs.writeFile(path.resolve(config.specPath), res.body)
```

`writeFile` can take a stream as the input data, and `Response.body` returns a readable stream. So it should jsut work!

josephjclark (commented):
This is pretty good - I've made a few suggestions but nothing critical.

I note there's no unit tests but I can see how testing this stuff would be tricky. Also the deploy command could use some CLI tests generally, it doesn't really have any yet. So that's something to worry about later!

josephjclark (approved):

PR #293, opened 2023-07-11
Typesync?

josephjclark:
This is a cool utility I just stumbled across.

[typesync](https://www.npmjs.com/package/typesync) will add @types packages (and I think it version matches too - if it doesn't version match I don't want it!!). 

I've added a top-level `typesync` command, which will sync types on the monorepo root and all nested packages. Instructions and explanation are in the readme.

Usually typesync is set up automatically install types when a new package is added. But it doesn't seem to play well with `pnpm`. I've raised #333 but I still think this is worth having as a manual command.

**A note on @types versions**

I had to look this up: @types packages always sync with the major and minor version of the package they relate to. The patch version on a @types package refers to the version of the types, rather than the version of the library.

So react 18.1.7 will have a corresponding @types/react package version 18.1.0. If there's bug in this definition and @types/react is updated, it will now have version 18.1.1 (but still describe react 18.1.x).

Observations and testing

* The CLI has `yargs@17.7.2` declared. `typesync` automatically added `@types/yargs@17.0.24`, which is the latest version. So I guess i n the case of yargs `@types` only matches to the major version.
* If I downgrade yargs to 15, `typesync` will NOT also downgrade the @types package. `typesync` docs say it will "try" to respect version parity but fallback to latest.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,8 @@ To manually run typesync, do `pnpm run typesync` from the repo root. Note that y
 
 [TODO] typesync will run automatically when a new package is installed.
 
+Note that @types packages only synchronise with the major and minor versions of a package. So for `@types/x@major.minor.patch`, `major` and `minor` refer to the versions of the corresponding package `x`, and `patch` is the verrsion number of the actual types package.
+
 ## Testing the CLI on a branch
 
 From the repo root, run `pnpm install:global`.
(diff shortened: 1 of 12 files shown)


josephjclark:
It would be nice if this could run everytime dependabot update a version, so that types are already synced.

I don't know enough about how dependabot works but maybe that's something we can add script for?

PR #297, opened 2023-07-13
Pull command

### LINKED ISSUE #278: Add "--describe" to CLI

zacck:

- Adds a pull command to the cli
- Pull command should pull state json to the statePath as well as the project yaml to the specPath

## Related issue

Fixes #278 

### DIFF
--- a/packages/cli/src/pull/command.ts
+++ b/packages/cli/src/pull/command.ts
@@ -1,0 +1,29 @@
+import yargs from 'yargs';
+import { build, ensure } from '../util/command-builders';
+import { Opts } from '../options';
+import * as o from '../options';
+
+export type DeployOptions = Required<
+  Pick<
+    Opts,
+    | 'command'
+    | 'log'
+    | 'logJson'
+    | 'statePath'
+    | 'projectPath'
+    | 'configPath'
+  >
+>;
+
+const options = [o.statePath, o.projectPath, o.configPath];
+
+const pullCommand = {
+  command: 'pull',
+  desc:  "Pull a  project's state and spec from an instance to the local directory",
+  builder: (yargs: yargs.Argv<DeployOptions>) => {
+    return build(options, yargs).example('pull', '');
+  },
+  handler: ensure('pull', options),
+};
+
+export default pullCommand;
--- a/packages/cli/src/pull/command.ts
+++ b/packages/cli/src/pull/command.ts
@@ -15,13 +15,13 @@ export type DeployOptions = Required<
   >
 >;
 
-const options = [o.statePath, o.projectPath, o.configPath];
+const options = [o.statePath, o.configPath];
 
 const pullCommand = {
   command: 'pull',
-  desc:  "Pull a  project's state and spec from an instance to the local directory",
+  desc:  "Pull aproject's state and spec from a Lightning Instance to the local directory",
   builder: (yargs: yargs.Argv<DeployOptions>) => {
-    return build(options, yargs).example('pull', '');
+    return build(options, yargs).example('pull', 'Pull an updated copy of a project spec and state from a Lightning Instance');
   },
   handler: ensure('pull', options),
 };
--- a/packages/cli/src/pull/command.ts
+++ b/packages/cli/src/pull/command.ts
@@ -15,7 +15,7 @@ export type DeployOptions = Required<
   >
 >;
 
-const options = [o.statePath, o.configPath];
+const options = [o.statePath, o.projectPath, o.configPath];
 
 const pullCommand = {
   command: 'pull',
--- a/packages/cli/src/pull/handler.ts
+++ b/packages/cli/src/pull/handler.ts
@@ -1,0 +1,54 @@
+import {
+  DeployConfig,
+  getConfig,
+  getState ,
+} from '@openfn/deploy';
+import type { Logger } from '../util/logger';
+import { DeployOptions } from '../deploy/command';
+import path from 'path';
+import fs from 'node:fs/promises';
+
+
+
+
+async function pullHandler(
+  options: DeployOptions,
+  logger: Logger,
+) {
+  try {
+    const config = mergeOverrides(await getConfig(options.configPath), options);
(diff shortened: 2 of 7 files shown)


josephjclark:
These build errors are strange. They come out of `packags/deploy`. They don't occur on `main` or `release/next`. And yet I see nothing here which looks like it'll introduce the failure.

The type errors and build fails look legit to me. I've pushed a commit which fixes the typings - although why every other branch works remains a mystery :thinking:

packages/cli/src/pull/command.ts
```diff
@@ -0,0 +1,29 @@
...
+    | 'command'
+    | 'log'
+    | 'logJson'
+    | 'statePath'
+    | 'projectPath'
+    | 'configPath'
+  >
+>;
+
+const options = [o.statePath, o.projectPath, o.configPath];
```
josephjclark, line 18: The `projectPath` option doesn't seem quite right - the deploy config takes a `specPath` key, and when we write our output we write to `specPath`, not `projectPath`. There's a confusing mismatch there.

I may come back to you with more comments about the API, but at the very least `--project-path` needs to be `--spec-path`. You can keep `-p` and `--project-path` as aliases I guess, since `-s` is taken (for the record I don't think we should need to take -s as an argument at all, the CLI should manage that stuff automagically without user intervention)

Also, are all of these commands optional? Are they all required? My instinct is that the only one you really need is the config path (because you use a config for the first deploy, and afterwards all you do is sync the state of that config).

packages/cli/src/pull/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+// Config
+// Env vars
+// Options
+function mergeOverrides(
+  config: DeployConfig,
+  options: DeployOptions
+): DeployConfig {
+  return {
+    ...config,
+    apiKey: pickFirst(process.env['OPENFN_API_KEY'], config.apiKey),
```
josephjclark, line 46: This can be rewritten as `process.env['OPENFN_API_KEY'] || config.apiKey`.

packages/cli/src/pull/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+
+
+async function pullHandler(
+  options: DeployOptions,
+  logger: Logger,
+) {
+  try {
+    const config = mergeOverrides(await getConfig(options.configPath), options);
+    logger.always("Downloading project yaml and  state from instance");
+    const state = await getState(config.statePath);
```
josephjclark, line 22: The only thing we use state for here is  to get the project id. Do we really need to do that? Is it not available anywhere else?

packages/cli/src/pull/handler.ts
```diff
@@ -0,0 +1,58 @@
...
+    fs.writeFile(path.resolve(config.statePath), new_state);
+    logger.always("Project pulled successfully");
+    process.exitCode = 0;
+    return true; 
+  } catch (error: any) {
+    throw error;
+  }
+}
+
+// Priority
```
josephjclark, line 36: Comments need removing (and addressing?)

josephjclark (changes requested):
Hi Zaack, I've left a bunch of comments. Mostly trivial stuff, but a couple of meatier ones.

I need to sign out now but I do want to think a bit more about the state, project and spec stuff. I'm a bit confused about how it all interacts.

Some of the semantics from the deploy API look suspect as well (why does deploy.getProject() return state? Why is it wrapped in `{ data }`? And how is that different to deploy.getState()?). That stuff is impacting the semantics of the pull and making it all a bit harder to understand.

packages/cli/src/pull/handler.ts
zacck: I am likely going to change this so it comes in as command line param

packages/cli/src/pull/handler.ts
zacck: I think this is documentation of sorts

packages/cli/src/pull/command.ts
```diff
@@ -0,0 +1,29 @@
...
+  >
+>;
+
+const options = [o.statePath, o.configPath];
+
+const pullCommand = {
+  command: 'pull',
+  desc:  "Pull aproject's state and spec from a Lightning Instance to the local directory",
+  builder: (yargs: yargs.Argv<DeployOptions>) => {
+    return build(options, yargs).example('pull', 'Pull an updated copy of a project spec and state from a Lightning Instance');
```
josephjclark, line 24: Note: this may change but the first argument needs to be `pull -c project/myConfig.js -s project/app-state.json` or something - it needs the arguments as well as the description

josephjclark (approved):
Ok, so this now has parity with deploy command.

I'm still chewing over some stuff with deploy. But we can't make change here that conflict with deploy, nor do I want to go in and rewrite deploy now.

So I'm going to merge this now on the assumption that we'll be coming back here soon to make changes anyway. At least we can put out a working version to start testing and experimenting with!

Thanks @zacck

PR #303, opened 2023-07-17
Allow globals to be injected into the runtime

josephjclark:
Quite a scary title but I'm pretty sure this is OK!

This PR enables better unit tests in adaptors by letting us inject adaptor functions as global varibles.

This adds an API to the runtime which is NOT intended to be used by the CLI or Runtime Manager.

We should ask really hard questions about the safety of this but I think it looks OK (so long as it's not exposed to third party code).

Anyway this is a prerequisite for the adaptor testing PR https://github.com/OpenFn/adaptors/pull/296

### DIFF
--- a/packages/runtime/src/execute/context.ts
+++ b/packages/runtime/src/execute/context.ts
@@ -15,11 +15,14 @@ const freezeAll = (
 
 // Build a safe and helpful execution context
 // This will be shared by all jobs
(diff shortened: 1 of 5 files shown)


josephjclark:
This has been merged into #496  - I'm quite satisfied it's safe and it lets me write really nice unit tests. Closing.

PR #322, opened 2023-07-21
Fix error reporting issue

### LINKED ISSUE #317: CLI error: invalid workflow
An issue has cropped up where a failed dhis2 request is causing the whole workflow to abort with "invalid workflow".

It looks like this will happen whenever a job throws - which is odd because I'm sure I added a bunch of error handling for this.

This issue is to track an investigation into the problem.

josephjclark:
A issue crept up where a dangling unidentified variable was getting referenced in the error reporter causing a blow up.

It's been fixed, and I've improved the typing and tests.

This should really have been picked up by the build failing for type errors.

## Related issue

Fixes #317

### DIFF
--- a/packages/runtime/src/util/log-error.ts
+++ b/packages/runtime/src/util/log-error.ts
@@ -1,15 +1,16 @@
 import { Logger } from '@openfn/logger';
-import { JobNodeID, State } from '../types';
+import type { Error } from '@types/node';
+import { ErrorReport, JobNodeID, State } from '../types';
 
 export type ErrorReporter = (
   state: State,
   jobId: JobNodeID,
   error: Error
-) => void;
+) => ErrorReport;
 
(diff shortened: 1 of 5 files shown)


packages/runtime/src/util/log-error.ts
```diff
@@ -21,9 +22,8 @@ const createErrorReporter = (logger: Logger): ErrorReporter => {
       report.code = error.code as string;
       report.stack = error.stack as string;
     }
-
     logger.debug(`Error thrown by job ${jobId}`);
-    logger.error(`${error.code || error.name || type}: ${report.message}`);
+    logger.error(`${report.code || report.name || 'error'}: ${report.message}`);
```
josephjclark, line 26: This is the actual fix - the rest of the PR is typings and tests.

PR #344, opened 2023-08-02
Finish refactoring opts

### LINKED ISSUE #115: CLI: ensure-opts needs a refactor

josephjclark:
## Short Description

This PR finishes the ensureOpts refactor, removing that code entirely.

## Related issue

Fixes #115

## Implementation Details

Not all commands use the new options APIs - the simpler commands don't really need to. The point is the ensure opts stuff is gone, commands have a cleaner pattern, and we've removed some confusion. The entrypoint to the LCI is also cleaner.

There's more we can do (I'd like to remove the Opts object in favour of the individual command types), but this is good enough for now. This issue has dragged on for a very long time and fundamentally I've done  what I wanted to do.

## QA Notes

I'd love someone to do a quick run through of each command, make sure I haven't broken anything.

- [ ] I have added unit tests


packages/cli/src/options.ts
```diff
@@ -225,6 +228,19 @@ export const inputPath: CLIOption = {
   },
 };
 
+// TODO this needs unit testing
```
josephjclark, line 231: reminder to add a couple of new tests around this and logJson

PR #346, opened 2023-08-08
Update typescript, tsup and esbuild

### LINKED ISSUE #330: Update tsup
Without breaking anything.

Add an integration test on version output too.

josephjclark:
This PR updates TypeScript to v5 and tsup to 7.x (which means updating esbuild to 0.18).

Closes #330 

Unfortunately the update introduced an issue, causing the wrong versions to be reported in `openfn version`. This is cosmetic but it still sucks.

The problem is that, for reasons I fail to understand, the external `package.json` setting in our tsup config is failing, and so package.json is getting built into the cli bundle.

This does not happen in standalone tsup 7 or esbuild 18 projects, the relative path resolution works just fine.

It's a mystery but I've already spent way too much time on this. As a workaround I've imported the package.json as a path. It;s not the most robust solution ever but it should work.


josephjclark:
I would love a bit of basic QA from @mtuchi on this branch before I merge it down!

josephjclark:
In fact, I think we'll run QA on the main release branch because of all the refactoring there.

PR #347, opened 2023-08-08
Logger: print errors to stdout

### LINKED ISSUE #257: logger.error should print to stdout
Logger.error and logger.warn should print to stdout, because otherwise log lines in the CLI might be out of order.

josephjclark:
This might be contentious. Closes #257

This PR makes the logger, by default, redirect any `logger.error` or `logger.warn` message s to stdout.

The problem is that when errors are logged from job code, they appear out sequence in the shell because they're written to a different stream. This can result in very confusing log output.

The solution is to make the default log emitter - the thing the logger eventually calls `console.log` on - redirect errors and warnings. This is kinda nice because it doesn't affect the mock logger and none of our unit tests break.

This behaviour can be overridden by passing a different emitter (ie, `console`) object when calling `createLogger`.

An alternative implementation would be to make the CLI pass a different lo emitter, so that by default the logger uses stderr but the CLI is setup to only use stdout. This kind of puts us back to square one though.

This may affect Lightning so I don't want to merge it until @stuartc has taken a look and shared his thoughts. It will also affect the new runtime manager too - but it does feel appropriate that all job logs should print to stdout.

Note that although we may be able to re-assemble logs in the right order in the Lightning Log Viewer (sorting by timestamp, if we have the accuracy), the CLI still has this problem, which may be confusing for local devs.

### DIFF
--- a/integration-tests/cli/test/errors.test.ts
+++ b/integration-tests/cli/test/errors.test.ts
@@ -8,125 +8,120 @@ const jobsPath = path.resolve('test/fixtures');
 // These are all errors that will stop the CLI from even running
 
 test.serial('job not found', async (t) => {
-  const { stdout, stderr, err } = await run('openfn blah.js --log-json');
+  const { stdout, err } = await run('openfn blah.js --log-json');
   t.is(err.code, 1);
 
   const stdlogs = extractLogs(stdout);
-  const errlogs = extractLogs(stderr);
 
-  assertLog(t, errlogs, /job not found/i);
+  assertLog(t, stdlogs, /job not found/i);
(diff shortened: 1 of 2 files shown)


josephjclark:
Hey @stuartc, I'd be curious as to your thoughts on this - I suspect you may have an opinion!

stuartc (approved):
Yeah this seems reasonable. I am curious though, how were they out of order? Like these things emit so quickly - how were they looking incorrect?

josephjclark:
@stuartc I think there were http requests in involved. I forget. I did see the original case on Taylor's machine and it was genuinely very confusing!

Issue #277, opened 2023-06-21
CLI: log=none still logs runtime logs


josephjclark:
If you run a CLI job with log level `none`, logs from the runtime will still show.

This is because we specially default job logging to `debug`. That special default is actually overriding `none`, which it shouldn't. I suspect this is a regression brought about by recent refactors.

Actually this is probably a broader issue - that the default runtime log level overrides any global log argument.

This is partly by design - on the basis that if you've gone to the trouble to print a console.log from your job, you want the CLI to still show it.

Issue #163, opened 2023-01-13
'none' logging should probably still log errors


josephjclark:
If you run a job with `--log none`, you'll only get logging from the job itself. This is a great way to run a job and ignore all the CLI overhead (which is occasionally useful at dev time).

But if there's an exception running the job, the CLI won't show you the error. Which is a really really bad thing.

I suppose we could also have a `--log error` to handle this. But fatal errors should surely always show, so I'm not sure.

PR #348, opened 2023-08-08
Logger: Improvements to error handling

josephjclark:
## Short Description

A few small improvements here:

* The logger will always log errors (even if logging is disabled) #163
* The CLI will not print job logs if errors #277
* If you throw a non Error (ie, `throw "whoopsie"`), the runtime will log it

This doesn't really affect much in practice. In the unlikely event that someone is using log=none, they'll start to see the output they expect.

- [x] I have added unit tests

### DIFF
--- a/.changeset/fluffy-pets-fly.md
+++ b/.changeset/fluffy-pets-fly.md
@@ -1,0 +1,6 @@
+---
+'@openfn/logger': patch
+'@openfn/cli': patch
+---
+
+Always log errors (even if log=none)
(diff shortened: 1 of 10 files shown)


taylordowns2000 (approved):
hey @josephjclark , this PR does exactly what it says on the tin, but i've noticed that it doesn't show any output for `throw "oops"`.

it works for `throw new Error("oops")` but regular old `throw "oops"` is never logged, regardless of whether the log level is set to `none` or `debug`

maybe we merge this and open a new issue?

josephjclark:
@taylordowns2000 this should be fixed for you - would you care to run your test again and merge if happy?

PR #354, opened 2023-08-16
Add support to supplying a project id when pulling

zacck:
## Short Description

Modifies the pull command to consume a project id rather than using state

## Related issue
Related https://github.com/OpenFn/Lightning/issues/250

## Implementation Details

I added a projectID option to the pull command and used that rather than depending on the presence of a state file

## QA Notes

How to test
- Pull this branch
- build the project using `pnpm build` in the `kit/packages/deploy` and `kit/packages/cli` directories
- provide a statePath, specPath, apiKey and endpoint, either via configuration or by options additionally this takes a positional argument a project id
- run the new build as follows `./dist/index.js pull {some_project_if}` with `some_project_id` being an id of a project in the intance from the endpoint 
- The files at your specPath and your statepath will now be updated with your project'd spec and state, if one or both of these files are missing, they will be created, if the state  file exists it will be used when generating the new state fille.

- [ ] I have added unit tests

## Release branch checklist

If this is a release branch:

### DIFF
--- a/packages/deploy/src/index.ts
+++ b/packages/deploy/src/index.ts
@@ -27,7 +27,7 @@ function mergeDefaultOptions(options: Partial<DeployConfig>): DeployConfig {
   };
 }
 
-export { getProject, mergeSpecIntoState}; 
+export { getProject, mergeProjectPayloadIntoState}; 
 
 export async function getConfig(path?: string): Promise<DeployConfig> {
   try {
--- a/packages/deploy/src/index.ts
+++ b/packages/deploy/src/index.ts
@@ -9,7 +9,7 @@ import {
   mergeSpecIntoState,
   toProjectPayload,
 } from './stateTransform';
-import { deployProject,  getProject } from './client';
+import { deployProject, getProject } from './client';
 import { DeployError } from './deployError';
 import { Logger } from '@openfn/logger';
 // =============== Configuration ===============
@@ -27,7 +27,7 @@ function mergeDefaultOptions(options: Partial<DeployConfig>): DeployConfig {
   };
 }
 
-export { getProject, mergeSpecIntoState}; 
+export { getProject, mergeSpecIntoState };
 
 export async function getConfig(path?: string): Promise<DeployConfig> {
   try {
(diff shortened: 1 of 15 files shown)


taylordowns2000 (changes requested):
Two quick things and a question:

1. Remove "fixes https://github.com/OpenFn/Lightning/issues/250" because it doesn't seem like this PR alone would fix the 4 remaining items to do on that issue.
2. Please ensure that you've followed all the steps on the PR template. Failing tests and empty check boxes there make me think this isn't quite ready.
3. The Q: how is this used when you _do_ have a state file? i.e., how is it used after the first time it's used?

zacck:
@taylordowns2000 
I believe I have addressed your feedback
1. I have moved this back to draft and added instructions of how to test this
2.  on the other checkboxes, I am unsure of whether this is a release branch or not so i left that untouched, and I looked at the other PRs in kit and I hadnt found an example of one with a changeset and intended to ask @josephjclark  about how to use those, on tests as we implemented pull we realised there was no clear way of testing this command and I believe the same is true for deploy hence the unchecked box, 
On the failing type test as far as I can see thats not introduced by the changes in this branch, I may be wrong on this however

packages/deploy/src/index.ts
```diff
@@ -84,7 +84,7 @@ function writeState(config: DeployConfig, nextState: {}): Promise<void> {
 // ==================== Spec ===================
 
 // Given a path to a project spec, read and validate it.
-async function getSpec(path: string) {
+export async function getSpec(path: string): Promise<any> {
```
josephjclark, line 87: We should be able to do better than `Promise<any>` here. Are you sure you need to declare the type at all? I'd expet Typescript to be able to infer it.

If you really need an explicit type, you'll have to:
* in `validator.ts`, pull the return type of `parseAndValidate` out into its own interface - maybe called `ValidatedSpec` and export it
* Then here, you can do `Promise<ValidatedSpec>`

josephjclark: I've removed the `Promise<any>` - the reason we needed it (to my surprise) was that `Error` type in the validator wasn't exported.

josephjclark (approved):

PR #356, opened 2023-08-17
Deploy: fix typings

### LINKED ISSUE #355: deploy: fix types
I made a bunch of changes lately - including updating typescript and being stricter about type errors.

Somewhere down the line a bunch of type errors started coming out of `deploy`.

What I did was: I disabled type testing on deploy. Presumably with a view to sorting it later, but I forgot all about it.

It's currently holding back @zacck's work and needs dealing with.

josephjclark:
Fixes #355 

The deploy command starts failing when `mergeSpecIntoState` is exported from `deploy`. I think this is a result of extra rigidity on the type constraints we added recently, and maybe the typescript version bump.

This PR goes through  and addresses any type issues that get raised. A lot of them are bludgeoned over. I've had to make a few refactors for others. I've tried to not change any functionality but it's pretty complex in there and there will be diffs in the built JS.

I have no means of testing this so maybe someone can take a look before I merge?

### DIFF
--- a/packages/cli/state.json
+++ b/packages/cli/state.json
@@ -1,9 +1,0 @@
-export default [fn((state) => {
-  console.log('multiplying by two');
-  state.data.value *= 2;
-  return state;
-}), fn((state) => {
-  console.log('multiplying by three');
-  state.data.value *= 3;
-  return state;
-})];
--- a/packages/deploy/src/stateTransform.ts
+++ b/packages/deploy/src/stateTransform.ts
@@ -77,9 +77,12 @@ function pickValue(
   return defaultValue;
 }
 
-function mergeTriggers(stateTriggers, specTriggers): WorkflowState['triggers'] {
+function mergeTriggers(
+  stateTriggers: WorkflowState['triggers'],
+  specTriggers: WorkflowSpec['triggers']
+): WorkflowState['triggers'] {
   return Object.fromEntries(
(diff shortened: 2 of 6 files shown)


packages/deploy/src/stateTransform.ts
```diff
@@ -183,25 +178,23 @@ export function mergeSpecIntoState(
   const nextWorkflows = Object.fromEntries(
     splitZip(oldState.workflows, spec.workflows).map(
       ([workflowKey, stateWorkflow, specWorkflow]) => {
-        stateWorkflow = stateWorkflow || {};
-
         const nextJobs = mergeJobs(
-          stateWorkflow.jobs || {},
+          stateWorkflow?.jobs || {},
```
josephjclark, line 182: This is such a strange one. Even though `stateWorkflow` will always have a value, typescript insists on saying `stateWorkflow is possibly 'null'`.

I don't see how that can possibly be so, so I've added optional chaining which should workaround it. It's a pretty inelegant fix though.

packages/cli/state.json
```diff
@@ -1,9 +0,0 @@
-export default [fn((state) => {
```
josephjclark, line 1: This is a random and unrelated fix - this file shouldn't be here and needs removing. Doing it here is the quickest way to main.

josephjclark:
Closing here as this has been merged in to #354

Issue #364, opened 2023-08-22
Linker: more controls and better errors


josephjclark:
The runtime currently supports import statements. It has to - that's how it works!

All imports are handled by our `linker` function, which will attempt to load the module.

The linker will run through a few steps:
* It will check to see if there's a whitelist set and, if so, ensure that the module is on the whitelist of allowed modules
* If all is OK, work out the path to the desired module:
* * If the specifier is am absolute path to a js file, just use that path
* * If the linker has an explicit mapping for that path (which can be set on linker options), use that path
* * See if the module exists in the repo, and return that path
* If no path was mapped, throw an error
* Finally, try to `import()` the resolved path and inject the module interface into the job sandbox.

Note that nested imports will just work - it's on the imports in job code that are directly process

Also note that the runtime or linker will make no attempt to install modules. We're at the mercy of node's standard module lookup algorithm when we call `import()`.

This is fairly secure but kind of by accident. If I try to `import fs from 'node:fs', it'll fail because `node:fs` happens to not be in the repo. When really we should be throwing a message like "cannot import native node modules" or "node:fs forbidden".

Also the devx of error messages could be a lot better. Here's the error for `node:fs`:
```
file:///home/jclark/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:265
  throw new Error(`Failed to load module "${specifier}"`);
        ^

Error: Failed to load module "node:util"
    at loadActualModule (file:///home/jclark/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:265:9)
    at async linker (file:///home/jclark/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:202:19)
    at async file:///home/jclark/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:279:22
    at async ModuleWrap.<anonymous> (node:internal/vm/module:315:24)
    at async Promise.all (index 1)
    at async SourceTextModule.<computed> (node:internal/vm/module:333:11)
    at async SourceTextModule.link (node:internal/vm/module:199:5)
[...]

PR #366, opened 2023-08-23
Security Tests

### LINKED ISSUE #316: Security testing
This is an issue to run (and preferably commit) a bunch of security tests.

I am brainstorming here, some of these things may already be done (I am sure we've already got a unit test on process and eval, for example):

* ensure child_process and its ilk cannot be accessed from a job
* Ensure we're controlling and blocking node imports
* Try and break out of the vm sandbox with the Error exploit (TODO add link)
* Ensure an adaptor can't load an expose fs
* * Wider thought: where do we stand on adaptors? Can I create an adaptor called `@openfn/language-evil` and have it expose a bunch of terrible, terrible functions to the job?

Basically I want a reminder/a bit more assurance on protections the sandbox offers us today, and to do a bit of work to try and break out of the sandbox and see what becomes exposed.

josephjclark:
This PR adds a few security tests to the runtime.

There aren't actually many - I've been investigating ways to break out of the node:vm sandbox and basically every case I've seen comes down to passing a string into an eval-like function.

Our sandbox disables code generation, so all these cases throw an exception when someone tries to do something cheeky like this.

We already had a test on this but I've expanded out the cases a bit.

Maybe I'm missing something, but I it seems to be that by passing `codeGeneration: false` to `vm.createContext`, we shut off a whole class of exploits.

Note that there may be other attack vectors. I haven't yet been able to reproduce the Promise species thing. The `util.inspect` exploit depends on being able to import `node:util`, which we do not do (although our coverage here isn't amazing, see #364). This doesn't help with infinite loops or memory violations. And the multi-threaded RTM may have its own unique security concerns.

What this is about is proving that the core runtime sandbox provides a reasonable layer of security.

We should assume that in the future, someone will work out how to break out of the sandbox in `node:vm`. That's when our other layers of security in the runtime manager - like running in a thread, and keeping that RTM on different VM entirely to the Lightning one - should come into their own. Attackers will break out of a sandbox into a void.

## Related issue

Closes #316


josephjclark:
I did want to check the Promise species exploit but it's not been published yet and I've failed to reproduce it. We'll have to come back to that later.

stuartc (approved):
:kissing:

PR #379, opened 2023-08-25
Update pull and deploy to work together

### LINKED ISSUE #378: Project state created by deploy and pull don't match

taylordowns2000:
## Short Description

This PR updates pull and deploy to work together.

@josephjclark , I'd like your fresh eyes on the logic and _then_ I'd like your help writing tests and making the types not suck. (I've used that `any` escape hatch which you definitely should never have let me see!)

## Related issue

Fixes #378

## QA Notes

To test this:
1. First `pull` a project from a Lightning server and then run `deploy` on that same project. You should see no changes.
2. Then `deploy` a new project to a Lightning server, committing your `project.yaml` and new `projectState.json` to git. Then run `pull` with the project ID that you've just created. You should see no change in your diff between the yaml or the project state.

- [x] I have added unit tests

## Release branch checklist

If this is a release branch:


packages/deploy/src/utils.ts
```diff
@@ -3,6 +3,33 @@ export const uuidRegex = new RegExp(
   'i'
 );
 
+export const hyphenate = (str: any) => {
```
josephjclark, line 6: This will do the job for now but unfortunately I think it's only a matter of time until it breaks.

It currently converts the spaces in names  to dashes, which matches Lightning behaviour. But what about other non-url safe characters? `@ ! ? # * %` etc, or emoji? Those will fail to sync correctly.

I really think the project payload should include a safe name (on everything, including edges and triggers) as well as a human readable label.

taylordowns2000: something like this, @josephjclark ?

https://github.com/OpenFn/Lightning/issues/1053

josephjclark: Yep, that should cover it

josephjclark:
@taylordowns2000 I've tidied, typed and tested this and I'm happy for it to be merged. It could do with one last round of QA from you in case I've done anything subtle

josephjclark (approved):

PR #397, opened 2023-10-04
Worker: integrate with Lightning

josephjclark:
## Getting Started

To use this server:

* Start lightning on the release-0.1.0 branch
* Start the worker server:

```
pnpm start
````

You may want to add `--log debug` or disable the work loop, see below.

## Disabling auto-fetch

When working in dev I actually like to disable the workloop. It's noisy and I can't control it so well. To switch it off, run:
```
pnpm start --no-loop
```
To manually trigger a claim, post to `/claim`:
```
curl -X POST http://localhost:2222/claim
```

## The Runtime Engine is MOCKED

It's really important to note right now that the server uses a **mock** runtime engine. It will NOT run jobs properly.

The mock will trigger a bunch of events (including a log). I will try to parse the job expression/body as JSON, and if it success it'll return that as the result state. Otherwise it will just return the input state.

Dumb as rocks but great for testing.

## Changenotes:

Just noting a few things that had to change to get this working:

* Allow the claim loop to be disabled in dev in favour of a POST request to trigger a claim
* Change `attempts:queue` channel name
* Change `attempts:fetch` event name
* Replace `phoenix-sockets` with `pheonix` (big deal)
* Stub in a status for attempt complete

Also gonna do some refactoring - I've stolen the `channels/worker` folder idea

## TODO

Maybe do to on the base branch but:

* Standard response and error handling for all messages. What do we do on error? Log?


josephjclark:
Issue: we seem to join the attempts channel twice. I'm not sure where or why this is happening. Lightning gets two joined notifications and the worker gets two joined callbacks. I'm pretty sure join() is only called once, it's just getting triggered twice somehow.

josephjclark:
Issue: I think this is just the mock, but handling the arraybuffer dataclip response isn't really working very well. We have to encode the whole `{ status: 'ok', response: buffer } ` object for nu-phoenix to encode it properly... but when it does that the reply doesn't seem to trigger (I think because the reply message status is encoded). This is presumably just something I need to work through in the mock. Seems to work with real lightning. I will disable the dataclip test for the time being, it's not importantenough

josephjclark:
Issue: There are too many files and functions call the same thing to do with connecting and joining. That code really needs breaking down and straightening out better. Hoping to do this before I merge the work but let's see.

josephjclark:
Issue: the initial trigger, which we "run" with a job id, doesn't actually have an associated job.

This is kind of an artefact of how triggers are handled by the runtime - they're no-op nodes evaluated at the start of the workflow.

The worker needs to ignore events on trigger nodes.

Issue #365, opened 2023-08-23
RTM: Kill a thread that timesout (or maybe stops responding to a heartbeat)


josephjclark:
Quick note as I don't think this is covered.

If a job does `while(true) {}`, the thread will loop forever.

Lightning will timeout attempts and re-queue them if appropriate.

But the RTM itself needs to do some management of its threads, killing them if they timeout, We may want to impose a memory limit, if we can. And maybe we should send a heartbeat signal to the worker, ensuring the thread isn't locked, killing a`while(true){}` after 30 seconds rather than 5 minutes. But this may introduce more problems if the thread is genuinely busy parsing 10mb of CSV data or whatever. So I don't know,

josephjclark:
check that the runtime itself can timeout a job

josephjclark:
So this is done at the attempt and engine level now. The engine will timeout an attempt after n ms, if the option is passed.

I've raised #414 because we've sort of got two ideas of what a timeout is now, and it's not particularly good.. This affects lightning so it's not something I can just decide out of the blue.

Issue #263, opened 2023-06-08
RTM: support disabled edges


josephjclark:
Lightning may soon start sending us disabled edges (`disabled: true` or perhaps the inverse).

This is a soft way for users to limit a workflow's functionality without losing anything.

When converting Attempts to Execution Plans in the runtime manager server, we basically want to ignore disabled edges.

josephjclark:
This should now be done in #404 , assuming than an incoming edge on an attempt has `enabled: false`.

For now, at least, I've taken the view that an edge is enabled unless it's explicitly disabled with `enabled: false`

Issue #120, opened 2022-12-06
runtime-manager: lock down pooled threads


josephjclark:
The runtime manager uses thread pooling to run jobs. This reduces the overhead of starting up a new process for every job - idle threads will be re-used.

There is a security issue here: because threads are shared, their environment is ALSO shared. That means a job could corrupt a worker's environment and compromise the next job to run in that thread.

There are two solutions which we should explore more nearer the time:

1) use `Object.freeze()` or `ses.harden()` to freeze the global scope
2) use `vm.runInNewContext()` on every job (probably inside the worker's `run` function) to isolate each global scope.

Breakouts are presumably a theoretical concern in both cases. I don't know which is more secure. If we can't find a reasonable solution then we'll have to abandon node pooling (which makes you wonder if threading is even worth it)

josephjclark:
Some unit tests around this: https://github.com/OpenFn/kit/blob/worker-tests/packages/runtime-manager/test/worker-pool.test.ts

josephjclark:
You know, the runtime currently creates a new context object (basically a global scope) for each run. So actually, subsequent jobs  in the same worker thread should already have their own contexts. This needs checking and testing.

When we create a context, it's probably a good idea to freeze or harden it before running the job. With slightly tighter control over what goes in that scope (see #104) it's probably a pretty secure starting point.

josephjclark:
Ok, confirmed. Each runtime creates its own context, so two runs can't actually interfere with each other. This means we don't need to sandbox threads any more (unless someone can break out of the vm contest into the parenting thread scope)

See unit tests around this stuff here: https://github.com/OpenFn/kit/blob/security-tests/packages/runtime/test/security.test.ts

Note that while the scope is isolated, it may still not be entirely secure. We're doing nothing to harden it, so a job can override eg the array prototype.

josephjclark:
So I think this one is gonna stay open.

I've been testing around this most of the day. Where  we are now is:

Jobs running in pooled threads are totally dependent on the runtime sandbox to preserve their environment.

We can't really freeze the global scope to prevent writes (freezing `global` is easy, but you have to freeze every variable in the top module scope, and that includes things which can't be frozen, like the typed arrays). So if someone was to break out of the job jail, they'd be able to scribble to the top worker scope and set off a long running function or interval or something to spy on the runner's environment.

Having broken out, though, I don't think there's any way to break back in to the next job.

I can confirm that the parent process is not available to the child.

[...]

PR #404, opened 2023-10-18
Runtime Engine

### LINKED ISSUE #393: Epic: Runtime Engine
Epic to track the work needed in the multi-process runtime worker.

Remember we don't have a good name for this (runtime manager is insufficient). See #374. Runtime Worker might be useful.

## Essential

* Review the existing code and work out what needs doing
* #372
* #373 
* #120 (security, need to investigate this)
* Work out what this comment means: https://github.com/OpenFn/kit/issues/52#issuecomment-1371174282

## Useful
* #388
* #365

## Optional
* #265 (idk if we need this any more)
* #263

## To close

* #52 is irrelevant now, I just need to look into that mysterious comment and either transfer notes to a new issue or close it down

## Not in issues

Stuff that needs doing that we don't have issues for 

* [x] Support the `sanitize` log flag
* [x] Does the existing service worker even support workflows? Does it need to do anything for that?
* [x] Lazy loading assets (state/credentials)~~ (basically done but I'm testing)
* [x] At the moment we use worker threads to run the jobs (with compilation in the main thread). I don't know if this is its final form. But we should be able to swap this out later.
* [ ] Caching compiled jobs
* [x] Update API to meet new standards
* [x] Log version numbers (there is a wider story about logging but right now we'll just do the minimum)

## Security?

Security considerations for the actual process worker are more to do with not allowing malicious code to run and not allowing other environments/processes/jobs to be read. A lot of the security is subsumed by the enclosing server.

josephjclark:
This branch is tracking changes needed to update and integrate the multi-threaded runtime engine

* [x] Restructure API (this was a big deal)
* [x] Support job start and complete in the runtime
* [x] Pull the lightning mock out into its own package
* [x] Integration tests of lightning mock, worker and engine-multi running locally
* [x] Containerised integration tests
* [x] Autoinstall improvements and tests
* [x] Handling lazy-loading initial state
* [x] Handling lazy-loading credentials
* [x] Whitelist @openfn/modules
* [x] Better security on pooled threads #120
* [x] Support disabled edges #263
* [x] Support options on attempt (including sanitize arguments) 
* [x] Kill timeout threads #365
* [x] validate worker thread with a handshake event, throw if it it fails https://github.com/OpenFn/kit/issues/393#issuecomment-1761126493
* [x] Check that the parent env is hidden to the running worker https://github.com/OpenFn/kit/issues/393#issuecomment-1760033538
* [x] Integration test on stateful adaptors
* [x] Release bundles

The audit trail on this isn't going to be terribly good because I'm just keeping my head down and making it all work.

An interesting thing about the new engine design is that there's a core engine component, which exposes all of its internals in a large ish API (well, sort of), versus the API component, which wraps the engine, sets defaults and obfuscates parts of the API that no-one should be tampering with.

The idea is that clients/wrappers of the engine will instantiate the API and use its limited surface basically to a) run workflows (we call them workflows or plans here) and b) listen to events emitted during execution.

For unit tests, or for custom deployments or advanced users, you can use the engine directly. This lets us do things like use a mock worker environment, override autoinstall and compile rules.

This PR should close #393 and everything in it. Apart from the logs thing, which is going to be ongoing.

### Reasons

For now, I am returning to attempt:complete with an "OK" reason

Any critical/crash errors along the way will return to attempt:complete with a "FAIL" reason.

This needs a lot more work, it's just where I am right now.


josephjclark:
An issue has crept up with credentials.

The idea is that we the worker thread would call out to the parent thread, the actual engine, and ask for a credential on demand. The engine in turn would call out to the worker, which sends an event out to lightning, and eventually a credential finds its way back into the thread,

A major gotcha with workerpool: the child worker can send an event to the parent, but after a thread has started the parent cannot send a message back.

This is kind of an innate limitation with the workerpool approach. The pool is a black box - you ask it to run a job and just wait until it finishes. The details are sort of supposed to be abstracted from you.

I still don't think we're 100% sold on worker threads as the solution here. If we were, we could implement our own pooling solution (a nice thing we can do here is to pool a worker for each adaptor version, keeping the environment warm and reducing the amount of importing we  have to do).

For the moment I've just preloaded all of the credentials on the attempt and sending a fully-formed execution plan into the runtime.

This is OK but I'd prefer to be super paranoid  about credentials and treat loading them as a last resort.

josephjclark:
Worryingly there seems to be a Flaky Test :tm: coming from some of the workerpool tests. I think something is getting killed early by the test environment (a problem with my setup or teardown code, I'm sure)

josephjclark:
I've ticked off the security box but I've left #120 open because I would like to find a good way to harden the worker environment a bit more. I've added documentation and tests to the repo to cover where we are.

josephjclark:
One thing is worrying me a bit.

I'm pushing for a pattern in adaptors where we use a top-level closure variable in the module to maintain stateful client objects - like database clients, or the undici client.

If you run two workflows in the same worker thread, and they use the same adaptor, will they accidentally be sharing a client instance?

In workerpool alone the answer would be "maybe" - it depends a bit on how the adaptor is instantiated. If it creates the client in `execute` then it'll create it's own client and override the previous. I don't like that so much but its fine.

If the client is just declared at the top of the module, on the assumption that the module will be loaded in a new environment for each run, then the client WILL be re-used. Which provides a gateway to creep information between jobs.

However, I'm not sure what the vm sandbox means for this. The sandbox creates a new context for each run, which I think will re-load the adaptor module (unless we enable some sort of caching).

Basically I want to create a unit test here to better understand where we are and whether we're exposing a leak. As a general principle, I want each workflow to load a clean, fresh module instance. I don't want workflows to share pre-instantiated instances (unless we happen to know that those are frozen or otherwise tightly locked down)

josephjclark:
Ach a late point of confusion has crept up, I think I need to deal with this next week because my head is spinning.

Should the engine accept resolvers in its constructor (one set of resolvers works for all credentials and dataclips), or should they be passed to each execute call?

At the moment they are passed on execute, but suddenly that seems strange to me (actually, I think I've accidentally been supporting both APIs all week and the point is I need to consolidate on just one)

integration-tests/worker/test/integration.test.ts
```diff
@@ -0,0 +1,420 @@
...
+//       },
+//     ],
+//   });
+// });
+// });
+
+
+// This test fails, which is a problem!
+// the test module does not re-initialise
+test.skip('stateful adaptor should create a new client for each job', (t) => {
```
josephjclark, line 359: This test fails, which could be a problem. We ought to talk about it.

PR #420, opened 2023-11-01
Dockerize ws-worker

### LINKED ISSUE #418: Dockerize ws-worker
- [x] Add a Dockerfile
- [x] Add a docker-compose
- [x] Make sure it works
- [ ] Optimize the Dockerfile for production
- [ ] Automatically build images and push them to Docker Hub when releases are sent to NPM

taylordowns2000:
@josephjclark , this is ready for merging and testing if you're happy with _when_ the workflow runs. my idea is that it should be triggered when a tag is added for `ws-worker`. seem right?

Fixes #418 #421

- [ ] I have added unit tests

## Release branch checklist

If this is a release branch:


taylordowns2000:
@josephjclark , there is a failing `engine-multi` test here. should i worry about that?

josephjclark:
@taylordowns2000 There are one or two flaky tests in `engine-multi`. I've got an issue for it. Retrying seems to have helped

packages/ws-worker/package.json
```diff
@@ -13,6 +13,7 @@
     "start:watch": "nodemon -e ts,js --watch ../runtime-manager/dist --watch ./src --exec 'pnpm start'",
     "pack": "pnpm pack --pack-destination ../../dist"
   },
+  "bin": "NODE_OPTIONS=\"--experimental-vm-modules --no-warnings\" ts-node-esm --transpile-only src/start.ts",
```
josephjclark, line 16: I didn't know you could inline a command like this in bin - pretty cool!

This may be a dumb question but why would we want `npx @openfn/ws-worker` rather than `npm start`?

I'd see npx as an end user convenience, making it super easy to download and run the worker. But I don't know who the user is who wants that?

josephjclark: If it's for Lightning's convenience then I defer to @stuartc 

But if Lightning has a dev-time dependency on `ws-worker`, surely it's more transparent and honest to just declare it as a dev dependency?

The bin stub would still be useful though actually because Lightning will want to do `npm exec @openfn/ws--worker` or something to run it from the local repo

josephjclark:
I've pushed a commit with a bin stub that Lightning can use with `npm exec`, ie `pnpm exec worker`

**Really really important**: To pass arguments to the worker CLI, you have to add `--` to the exec command, and then list the arguments. This is an artefact of node `npm exec` works - it'll eat all of the arguments until `--`, and anything after gets passed to the command.

So to set the port for example:
```
pnpm exec worker -- -p 1234
```

josephjclark:
There is one integration test consistently failing here. It passes locally. I'll investigate some more.

josephjclark:
So the integration test is repeatedly crashing. It actually passes all tests and then something in workerpool blows up. I've been seeing this for a while but nothing like as bad as on this branch.

Anyway, I'm satisfied that the test fails are nothing to do with this work. I've also pushed some "fixes" to a couple of flaky tests which i hope will help, but is starting to feature creep on this branch.

So, I'm merging and releasing this.

Issue #394, opened 2023-09-28
Runtime: handle crashes


josephjclark:
Quick issue:

* We currently always finish a job as complete (maybe with errors)
* We need to recognise crashes - worker - out of memory, timeout, syntax error (requires lazy compilation, see other issue), linker error
* On crash, we stop executing the plan immediately
* This means the execute function throws, rather than returns
* The engine (nee rtm) needs to handle this and emit job-error event. This might be separate issue.

josephjclark:
Make this a minor version change and this will be 0.1.0.

This will likely be bumped to 1.0.0 after a patch or two.

josephjclark:
CLI will probably need a patch to fail differently in the event of a crash

josephjclark:
The runtime now handle crashes and fails and formally define errors.

More work will be needed, but I'm calling this issue closed.

Issue #384, opened 2023-08-28
Runtime: throw clear exceptions if code generation is attempted


josephjclark:
If a job tries to do an eval (or eval like), they'll get an error like "Code generation disabled for this context".

We should wrap this in a clear, contextualised error message so that it's clear what's happening.

This is a ux/devx issue

PR #422, opened 2023-11-01
Runtime crashes

### LINKED ISSUE #416: Error Handling
A design vision for error handling.

Lightning recognises the following kinds of errors:
* Fail: A Workflow completed (executed everything it wanted to) but threw one or more errors (http 500, custom error)
* Crash: The Workflow failed to execute because of some unexpected error (runtime error, syntax error, module error, oom)
* Kill: Someone - a user/admin, Lightning, maybe workerpool or a supervising process - killed the workflow before it could complete
* Cancelled: the user asked for a workflow to be cancelled.

Ultimately the worker is responsible for catching reporting a reason-status to Lightning (and so the end user).

But here's the fun bit: the actual error could come from various places and manifest in different ways - including in websocket messaging or inside a worker thread. How is the Worker supposed to know?

Well, here's how.

# Error Classes

Every component (runtime, engine, worker) must declare a clear set of errors.

I think all these errors have a common interface:

```
interface OpenfnError {
   source: 'runtime' | 'compiler' | 'engine', /// which part of the stack generated this error?

   severity: 'fail' | 'crash' | 'kill', // Maybe - this makes like easier for the Worker

   message: string; // user friendly, good enough to be shown in Lightning verbatim

   type: string; // basically the classname I think?

   subtype: string; // An optional soft subtype. For example, a RuntimeError could be a TypeError or ReferenceError subtype

    toString(): string // This should probably be common - report all errors  as strings in the  same sort of way
}
```

They may of course extend this interface with extra information. We can optimise the logger to handle these kinds of error too.

Then each component needs to declare a list of error classes, and throw them. These may be catch-alls or may wrap existing error objects.

The idea is that each component clearly declares the errors it might throw, which should make it easier to handle them downstream.

# Throw in the right place

OK and Fail states should be reported if the runtime returns a finished state. Ie, the runtime completed execution and does not throw.

The runtime needs to be able to crash, and throw if it does.

The Engine needs to emit `workflow-complete` for a success or fail, but emit a `workflow-error` if there's a crash.

# Map in the Worker

Ultimately, the worker will receive a `workflow-complete` or `workflow-error` event from the engine (if it receives neither, the job will timeout at the lightning end).

It will need to map these errors to a fail state and send that back to lightning. If every error has a severity and message, that is relatively straightforward.

# Related Issues

We need to ensure that the runtime crashes in the right way - for example, if it hits a syntax or runtime error  during execution, it needs to abort execution (right now I think it'll carry on to the next job). See #394

josephjclark:
This PR basically does two things:

* Crash in the runtime (ie, throw when we encounter a severe error)
* Report exit reasons in the worker

There's quite a lot of stuff in between those headlines but that's about the size of it.

The work is far from done: it'll be an ongoing process to improve messaging, catch more cases, and report more accurate reasons.

This maybe closes #416 and #394 #384

## Changes in this PR

Runtime:
* Declare a whole bunch of errors in `errors.ts` with reasonable console.log output
* `run` will now THROW for a crash, but for a fail will continue to execute
* Notify error events on fail or  crash
* Try to detect if an error came from adaptor code and classify it as an Adaptor Error (a fail state)

Engine:
* detect error events and report them to the worker
* A bit of refinement to handle the runtime throwing more often

Worker:
* Calculate job exit reasons based on state and any error objects
* Calculate attempt exit reasons - if the runtime crashed, the crash is the reason. Otherwise find the first fail and report that. Otherwise report success.

CLI:
* Tweaked CLI output to handle crashes. This is kind of important, see the comment way down below

## Breaking Changes

This PR will break lightning

* run:complete now includes `{ reason,error_type, error_message }`. 
* attempt:complete now also includes `{ reason,error_type, error_message }`. 
* A successful run now has `reason: "success"`, not `ok`

## Future Work

We can make big improvements here:

* Take a close look at error messages for each error
* Take a close look at CLI output for various messages (with better tests)
* Add fix hints to each error (ie, some kind of detailed drill-down with details about how the error might be resolved by a user or admin)
* Append useful stack traces - which both show where the error occurred within a job, but also the call stack. This is less important for "internal" errors but we still need to log something.


josephjclark:
The changes here affect the CLI, potentially in a big way, because the runtime behaves differently when errors occur.

I was a little worried about this but I've run tests against CLI `0.4.3` and `openfnx` (man, I love openfnx) and basically it looks fine.

Right now I've introduced a small bug in the error message. I can live with this, but I'll try and remove it quickly.

Here's a detailed comparison across the two builds:

## Bad request

In http with a bad request, I get this in prod:
```
[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
[R/T] ✘ Failed job job-1 after 559ms
[R/T] ✘ ERR_BAD_REQUEST: Request failed with status code 404
[R/T] ✘ Check state.errors.job-1 for details.
[CLI] ✔ State written to output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 654ms
```
But I on openfnx I get this
```
[CLI] ♦ Versions:
         ▸ node.js                  18.18.2
         ▸ cli                      branch/exit-reasons
         ▸ runtime                  ./openfn-runtime-0.0.33-local.tgz
         ▸ compiler                 ./openfn-compiler-0.0.38-local.tgz
         ▸ @openfn/language-http    latest
[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
[R/T] ✘ Failed job job-1 after 536ms
[R/T] ✘ AdaptorError: Request failed with status code 404
[R/T] ✘ Check state.errors.job-1 for details.
[CLI] ✔ State written to output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 629ms

```
So it's `ERR_BAD_REQUEST: Request failed with status code 404` vs `AdaptorError: Request failed with status code 404`. I gain AdaptorError but lose ERR_BAD_REQUEST. Otherwise no differnence.

## No adaptor error

The same http job with no adaptor passed in (a common CLI error!)

openfn:
```
[CLI] ⚠ WARNING: No adaptor provided!
[CLI] ⚠ This job will probably fail. Pass an adaptor with the -a flag, eg:

          openfn job.js -a common

[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
vm:module(0):1
export default [get("https://api.chucknorris.io/jokes/err"), fn((state) => {
^

ReferenceError: get is not defined
    at vm:module(0):1:1
    at SourceTextModule.evaluate (node:internal/vm/module:226:23)
    at module_loader_default (file:///home/jclark/.nvm/versions/node/v18.18.2/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:292:16)
    at async prepareJob (file:///home/jclark/.nvm/versions/node/v18.18.2/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:412:21)
    at async file:///home/jclark/.nvm/versions/node/v18.18.2/lib/node_modules/@openfn/cli/node_modules/@openfn/runtime/dist/index.js:370:35

Node.js v18.18.2
```
openfnx:
```
[CLI] ⚠ WARNING: No adaptor provided!
[CLI] ⚠ This job will probably fail. Pass an adaptor with the -a flag, eg:

          openfn job.js -a common

[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
[R/T] ✘ Failed job job-1 after 1ms
[R/T] ✘ CRITICAL ERROR! Aborting execution
[R/T] ✘ ReferenceError: ReferenceError: get is not defined
[CLI] ♦ Workflow failed in 18ms.

```

Openfnx has actually quit with a little more dignity here. I don't like the double reference error but that's something I can improve. It's cleaner output so a marginal gain. Happy with that.

## Syntax error

Exactly the same (because we catch it in the compiler, not the runtime):
```
[CLI] ⚠ WARNING: No adaptor provided!
[CLI] ⚠ This job will probably fail. Pass an adaptor with the -a flag, eg:

          openfn job.js -a common

[CLI] ✘ Failed to compile job
[CLI] ✘ Unexpected token (1:3)
[CLI] ♦ Check the syntax of the job expression:

get"https://api.chucknorris.io/jokes/err");

fn((state) => {
  console.log(state.data.value);
});


[CLI] ✘ Critical error: aborting command

```

## Type error

What if we reference something on state that doesn't exist?

Almost the same - we just get a duplicate reference error in the error string
```
[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
GET request succeeded with 200 ✓
[R/T] ✘ Failed job job-1 after 497ms
[R/T] ✘ TypeError: TypeError: Cannot read properties of undefined (reading 'y')
[R/T] ✘ Check state.errors.job-1 for details.
[CLI] ✔ State written to output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 591ms

```
Not great, but a minor bug that I can stand living in the CLI for a few days

## Reference error

This is a runtime crash, where the biggest changes likely are:

In openfn:
```
[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
GET request succeeded with 200 ✓
[R/T] ✘ Failed job job-1 after 541ms
[R/T] ✘ ReferenceError: x is not defined
[R/T] ✘ Check state.errors.job-1 for details.
[CLI] ✔ State written to output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 637ms
```

And in openfnx:
```
[CLI] ✔ Compiled from http-bad.js
[R/T] ♦ Starting job job-1
GET request succeeded with 200 ✓
[R/T] ✘ Failed job job-1 after 577ms
[R/T] ✘ CRITICAL ERROR! Aborting execution
[R/T] ✘ ReferenceError: ReferenceError: x is not defined
[CLI] ♦ Workflow failed in 673ms.
```
It's a little bit different and that's expected.

The production build will try and keep executing the next job, while the new build will "crash" and abort early.

This is fine.

packages/ws-worker/test/reasons.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    const onComplete = (result) => {
+      done(result);
+    };
+
+    // @ts-ignore
+    doExecute(channel, engine, logger, plan, {}, onComplete);
+  });
+
+test('success', async (t) => {
```
josephjclark, line 59: These tests are a good place to look at the cahnges.

The `reason` object` here is a bucket of keys that are added to the attempt:complete payload

I will add full integrations tests for this but that'll be a later PR

packages/ws-worker/src/api/execute.ts
```diff
@@ -215,34 +263,41 @@ export function onWorkflowStart(
...
+// find the higher priority
+// And return that as the highest exit reason
 export async function onWorkflowComplete(
   { state, channel, onComplete }: Context,
   _event: WorkflowCompleteEvent
 ) {
   const result = state.dataclips[state.lastDataclipId!];
-
+  const reason = calculateAttemptExitReason(state);
   await sendEvent<ATTEMPT_COMPLETE_PAYLOAD>(channel, ATTEMPT_COMPLETE, {
```
josephjclark, line 275: Here's the diff on the `attempt:complete` payload. `status` is removed, and `reason` is as `run:complete`

Issue #415, opened 2023-10-31
Flaky workerpool tests


josephjclark:
We  are experiencing intermittent test fails in CI (and locally)

I'm pretty sure it's coming from workerpool - sometimes when we close our pool, we get an exception from node. It's almost definitely timing related and could be OS limited - tests which spin up a lot of threads are likely to trigger this. I see it a lot in local dev and have yet to work out a pattern.

![image](https://github.com/OpenFn/kit/assets/7052509/a7638a49-e2ec-4f25-a0b4-e06b2b9b3f6c)

This fail seems to come from integration tests, which is kind of odd because all the tests share a pool.

If we move away from workerpool then this might go away.

Alternatively we can look at mocking out in some test suites.

josephjclark:
I've added a `destroy` hook to the engine in #420 which should instantly close all workers. I think this might help fix the flaky fails.

PR #425, opened 2023-11-02
fix flaky test

josephjclark:
This fixes flakiness in integration and unit tests #415

It includes an actual bugfix, so that's nice.

There are two basic issues:
* If you do `execute(workflow).on('workflow-complete')` with a mock worker, execute could complete before the listeners are added. Apparently. I am deferring the execute until after the listener hooks are returned to guarantee stability - although on reflection I'm not quite sure HOW this was causing failures...
* If you do pool.terminate and then exit the process, the pool may not actually finish terminating, and triggers an error. I've made the destroy calls async on a timeout which I hope will just let the worker close out properly.

Another note is that I've made more tests serial, ensuring the engine instances get created and destroyed safely and don't interfere.


packages/engine-multi/src/api/call-worker.ts
```diff
@@ -60,7 +60,14 @@ export default function initWorkers(
   };
 
   // This will force termination instantly
-  api.closeWorkers = () => workers.terminate(true);
+  api.closeWorkers = () => {
+    workers.terminate(true);
+
+    // Defer the return to allow workerpool to close down
+    return new Promise((done) => {
+      setTimeout(done, 100);
```
josephjclark, line 68: This seems to stabilise the tests, but boy it's a bit of a sledgehammer!

With it at 100, test have passed 5 times in a row

image

josephjclark:
This is looking pretty good - one integration test likes to fail very very rarely. I don't really understand why. May merge this now and deal with it later as we're still so early in dev

PR #436, opened 2023-11-06
Worker: Backoff tuning

josephjclark:
* CLI now takes backoff min/max, in seconds. To call every half second with a max backoff of 30 seconds, do `pnpm start --backoff 0.5/30`
* Slightly reduced the backoff multiplier to slow down the backoff
* Refactored some options


josephjclark:
Docker build is failing here because I've pushed the tag but the release isn't done

PR #440, opened 2023-11-06
release docs

josephjclark:
Update docs to handle tags on releases


josephjclark:
Two tests are still flaky :(

PR #445, opened 2023-11-07
High resolution timestamps

### LINKED ISSUE #444: Use high resolution time in timestamps
When sending logs out over the worker, Javascript's milllisecond timestamp resolution isn't fine enough to preserve log order.

We should capture all timestamps in the logger with nanosecond resolution using [process.hrtime.bigint](https://nodejs.org/api/process.html#processhrtimebigint)

Note: performance.now() gives us microsecond resolution natively - although it gives you time elapsed, rather than actual time. We could consider giving logs a relative timestamp and reconstructing in the UI from the start time. But this is probably too much trouble.

Are there performance or memory implications of this? Probably too small to worry about right now.

I think Lightning only wants microsecond resolution AND a string format. We should handle this conversion in the worker - but keep the actual logger on nanoseconds.

josephjclark:
This PR adds high resolution timestamps.

The logger now includes nanosecond precision in the `time` key in JSON logs (as a string).

The worker will convert nanosecond precision down to microsecond precision before sending to Lightning. Note that this does not guarantee log order - some logs will appear at the same timestamp.

Closes #444

### DIFF
--- a/packages/runtime/src/execute/expression.ts
+++ b/packages/runtime/src/execute/expression.ts
@@ -48,7 +48,6 @@ export default (
 
       clearTimeout(tid);
       logger.debug('Expression complete!');
-      logger.debug(result);
 
       exeDuration = Date.now() - exeDuration;
 
--- a/.changeset/cuddly-otters-drop.md
+++ b/.changeset/cuddly-otters-drop.md
@@ -1,5 +1,0 @@
----
-'@openfn/runtime': patch
----
-
-Remove log line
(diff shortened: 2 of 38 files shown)


packages/runtime/src/execute/expression.ts
```diff
@@ -48,7 +48,6 @@ export default (
 
       clearTimeout(tid);
       logger.debug('Expression complete!');
-      logger.debug(result);
```
josephjclark, line 51: Sneaking this in at Taylor's request

josephjclark:
This turned out to be a bit of a monster  - but I think it's ready to go.

Just waiting on confirmation from Taylor that it's working properly in Lightning

PR #448, opened 2023-11-08
Hrtime: send nanoseconds to lightning

josephjclark:
Keeping this open for a little while in case we want it.

This sends full nanosecond log timings to lightning. We can then restore the flaky worker integration test


josephjclark:
Gonna close this as it looks like it may not be needed after all...

PR #471, opened 2023-11-13
Syntax error tests

### LINKED ISSUE #463: Syntax error can crash a worker
This job code `fn(state => throw "ere");` can take down a worker! The implications aren't as bad given the new isolation and restarting, but we should still be able to catch these types of errors.

```
[SRV] ℹ 265d8fb9-9aa3-4f43-a6a6-8a3631cf70d4 :: attempt:log :: OK
[SRV] ℹ 265d8fb9-9aa3-4f43-a6a6-8a3631cf70d4 :: attempt:log :: OK
[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "#<Object>".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
 ELIFECYCLE  Command failed with exit code 1.
```

josephjclark:
Might close #463

### DIFF
--- a/integration-tests/worker/test/exit-reasons.test.ts
+++ b/integration-tests/worker/test/exit-reasons.test.ts
@@ -1,0 +1,73 @@
+import test from 'ava';
+import path from 'node:path';
+import crypto from 'node:crypto';
+
+import createLightningServer from '@openfn/lightning-mock';
+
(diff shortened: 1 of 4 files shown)


josephjclark:
Closing in favour of #474

PR #475, opened 2023-11-13
Runtime: clean state after fail

josephjclark:
The runtime, after executing an expression (job), should remove the configuration property from state before returning.

But,  if there was an error during execution, it does not do this!

image

After the runtime finishes executing it calls out to a function to sanitize the state and ensure that it is safely serialisable. This cleaned state then gets pushed downstream (or back to the user).

But in the event of an error, it'll continue executing the workflow without this cleaning step.

If this is a non-leaf node, the next job will run with a clean state and config and it doesn't really matter. But if it's a leaf node, we would return the state object as it appears at the point of error.

This PR ensures we sanitize state even after error, and pass the sanitised state downstream.

image

## Question

Now, here's a question.

After an error, do we pass downstream/return the state as it was at the point of error, OR the pristine initial state at the start of the job?

Do keep the incomplete state from the job or do we reset it?

At the moment we keep the incomplete state. This seems fine. I do not know the correct answer.

## Bonus Fix

Two things have been annoying me about errors so I've rushed in some quick fixes here:
* Remove the includeStackTrace key from errors, which I am not using
* Rename UserError to JobError (which feels like less of a personal attack)

### DIFF
--- a/packages/runtime/src/errors.ts
+++ b/packages/runtime/src/errors.ts
@@ -62,7 +62,6 @@ export function assertAdaptorError(e: any) {
 // Abstract error supertype
 export class RTError extends Error {
   source = 'runtime';
-  includeStackTrace = false;
   type: string = '-';
 
   constructor() {
--- a/packages/runtime/src/errors.ts
+++ b/packages/runtime/src/errors.ts
@@ -178,9 +178,9 @@ export class AdaptorError extends RTError {
 
 // custom user error trow new Error() or throw {}
 // Maybe JobError or Expression Error?
-export class UserError extends RTError {
-  type = 'UserError';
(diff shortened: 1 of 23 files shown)


packages/runtime/src/errors.ts
```diff
@@ -62,7 +62,6 @@ export function assertAdaptorError(e: any) {
 // Abstract error supertype
 export class RTError extends Error {
   source = 'runtime';
-  includeStackTrace = false;
```
josephjclark, line 65: This is incidental and unrelated. I've been seeing it in error reports and it bugs me.

taylordowns2000 (approved):

PR #478, opened 2023-11-13
Better leaf calculation

### LINKED ISSUE #476: Worker: The leaf for an attempt may not be the leaf for a plan
Right now, we're working out exist reasons based on whether any leaf nodes fail.

But we're defining a leaf node as: a node which has no `next` in the workflow.

But this may not be a leaf in the attempt! A node could have a `next` which fails its condition. In which case it's an attempt/de-facto/ad-hoc leaf. A leaf in practice if not in theory.

This  can cause failed jobs to be marked as success, because they have downstream nodes that weren't executed.

josephjclark:
I thought this would be difficult, but then I thought some more and I thought it would be easy.

Closes #476

Versions are bumped and this is ready to merge.

### DIFF
--- a/.changeset/warm-rules-draw.md
+++ b/.changeset/warm-rules-draw.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+When calculating exit reasons, include non-executed downstream nodes in leaf calculation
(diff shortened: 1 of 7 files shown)


josephjclark:
@taylordowns2000 this is looking fine to me and is passing CI tests.

Perhaps you can run a couple of different workflows before merging (ie this one and the original #472 one), just to be sure. My tests are hardly comprehensive yet and I'm getting more and more nervous about touching the deployed build (that said, I'm pretty confident in these changes)/

taylordowns2000 (approved):

PR #484, opened 2023-11-14
Fix input_data_clip id

josephjclark:
* [x] Add more tests around parallelisation in the runtime (it's fine!)
* [x] In the runtime, publish the next steps when a job is completed (this was a big deal)
* [ ] In the worker, use the next step information to work out the correct input_dataclip_id

### DIFF
--- a/.changeset/shaggy-cups-reply.md
+++ b/.changeset/shaggy-cups-reply.md
@@ -1,0 +1,5 @@
+---
+'@openfn/runtime': patch
+---
+
+Broadcast next steps with job-complete and error events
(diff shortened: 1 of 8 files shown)


taylordowns2000:
@josephjclark , i agree with pushing `final_dataclip_id` into a new issue because, well... from a lightning perspective, what does that even mean?

I really like this concept but not sure how it would be presented. Would you envision someone viewing an initial and final state for an _Attempt_? What would they use it for? Only to validate that the final state for each of the leaf nodes is proper? Typically, I'd envision them clicking on the leaf node for each run.

josephjclark:
@taylordowns2000 I will split it out but I do want to do a little more testing before I merge - its a deep edit!

josephjclark:
Well I'm glad I didn't merge this because I seem to have broken everything :(  Investigating now...

PR #489, opened 2023-11-15
Autoinstall better errors

josephjclark:
Our error handling for autoinstall is pretty bad - so I've made it better and added tests.

I hope this will help us catch errors in production.

No issue.

### DIFF
--- a/packages/engine-multi/test/api/autoinstall.test.ts
+++ b/packages/engine-multi/test/api/autoinstall.test.ts
@@ -2,7 +2,7 @@ import test from 'ava';
 import { createMockLogger } from '@openfn/logger';
 
 import autoinstall, { identifyAdaptors } from '../../src/api/autoinstall';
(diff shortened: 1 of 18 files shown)


packages/engine-multi/test/api/autoinstall.test.ts
```diff
@@ -271,4 +271,114 @@ test.serial('autoinstall: emit an event on completion', async (t) => {
...
+  const context = createContext(autoinstallOpts);
+
+  await t.throwsAsync(() => autoinstall(context), {
+    name: 'AutoinstallError',
+    message: 'Error installing @openfn/language-common@1.0.0: err',
+  });
+});
+
+// TODO this is a bit flaky apparently
+test.serial('autoinstall: throw on error twice if pending', async (t) => {
```
josephjclark, line 293: Yes, this is confirmed flaky.

Maybe this is hinting at whatever underlying problem we have here

PR #493, opened 2023-11-16
Harden lightning mocks

### LINKED ISSUE #492: Lightning mock: be more strict
My lightning mock is far too easy going.

It needs to be stricter:

* Reject messages with incomplete payloads
* Reject messages out of sequence (ie, reject run:complete if we didn't get a run:start`)

josephjclark:
This PR:
* Adds validation to mock lightning endpoints
* Tests validation
* Reorganisses the mock a bit
* Closes #492
* Adjusts and corrects some typings in the worker

### DIFF
--- a/.changeset/tidy-trains-care.md
+++ b/.changeset/tidy-trains-care.md
@@ -1,5 +1,0 @@
----
-'@openfn/lightning-mock': minor
----
-
-Add validation to endpoints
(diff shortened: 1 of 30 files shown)


josephjclark:
This is ready to merge. Annoyingly there's a diff and patch in the actual worker (although I suspect the built.js has no diffs) So I'm gonna have to roll this into a release.

I'll combine it with a bugfix to make it worth the effort

PR #496, opened 2023-11-17
Worker: use real runtime in unit tests

### LINKED ISSUE #486: Worker: use real runtime (but mock engine) in tests
Mocking in the Worker tests is wrong and is starting to hold me back.

The worker includes a mock engine implementation - which is great, because I don't want to run workerpool and multithreading and stuff in my unit tests.

But as part of this is a mock execute funciton which mocks out the runtime. This is wrong - the mock should use the REAL runtime, because a lot of my testing is based on events published by the actual runtime. With the mock in place I can't really test with the right fidelity.

josephjclark:
Closes #486 

The worker at the moment uses a mock runtime engine to test against.

This is quite reasonable because we don't want to be spinning up loads of threads while unit testing worker functions. It's overkill.

Previously the engine mock did include it's own fake little runtime with very limited functionality. But there are two big problems with this:
1) Job code passed into worker functions doesn't look like job code. It looks weird because it's written to run against the mock
2) Events and errors coming out of the runtime are quite complex, and the worker needs to be able to run real and sophisticated workflows to test certain behaviours. It's too hard to simulate this in a mock, so I'm using real code instead.

The mock engine does not autoinstall. It will call the credentials resolver. It will do a sort of fake compilation of jobs - it won't download an adaptor and do real compilation, but it will wrap the `fn()` call in an `export default [fn]` wrapper for compatability. It also injects some "fake" adaptor functions into the job scope.

## Globals

To make this work I'm pulling in some old code from #303. This allows us to set globals in the runtime scope.

This functionality is not exposed to the CLI or the Worker. Just the mock worker .

### DIFF
--- a/packages/ws-worker/test/lightning.test.ts
+++ b/packages/ws-worker/test/lightning.test.ts
@@ -447,5 +447,85 @@ test.skip('should not claim while at capacity', async (t) => {
   });
 });
 
-// hmm, i don't even think I can test this in the mock runtime
-test.skip('should pass the right dataclip when running in parallel', () => {});
+test('should pass the right dataclip when running in parallel', (t) => {
+  return new Promise((done) => {
+    const job = (id: string, next?: string) => ({
+      id,
+      body: `fn((s) => {  s.data.${id} = true; return s; })`,
+    });
+
+    const edge = (from: string, to: string) => ({
+      id: `${from}-${to}`,
+      source_job_id: from,
+      target_job_id: to,
+    });
+
+    const outputDataclipIds = {};
+    const inputDataclipIds = {};
+    const outputs = {};
+    const a = {
+      id: 'a',
+      body: 'fn(() => ({ data: { a: true } }))',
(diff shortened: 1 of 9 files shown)


packages/ws-worker/test/lightning.test.ts
```diff
@@ -444,7 +447,85 @@ test('should not claim while at capacity', async (t) => {
   });
 });
 
-// hmm, i don't even think I can test this in the mock runtime
-test.skip('should pass the right dataclip when running in parallel', () => {});
+test('should pass the right dataclip when running in parallel', (t) => {
```
josephjclark, line 450: As a result of using the new runtime, I can now write really sophisticated tests like this, which deeply test a) how the runtime handles complex Attempts and b) how the worker maps runtime events to Lightning events.

If I had this test two weeks ago, I'd have skipped over a lot of hurt this week!

This proves the value of this work, in the next couple of weeks I'll be building more more complex tests like this.

PR #502, opened 2023-11-21
Autoinstall issues

### LINKED ISSUE #454: Engine/worker: accept a repo dir (env var and CLI)
The engine currently always loads the repo out of /tmp/openfn/repo (which is actually the same place the local cli runs from)

We need a better, safer default, and we need to allow local lightning to override it, and the deployed server to set it.

josephjclark:
We are seeing numerous problems in autoinstall in production: see #503

My theory is that if the worker can install two seperate adaptors at once, which is causing interference when installing common dependencies like `lodash`.

This PR enforces autoinstall to use a strict queuing system. Only one adaptor can be installed at a time, and other adaptors must wait.

Also, as a bit of a prelude for this work I've done gone and fixed #454

### DIFF
--- a/integration-tests/worker/test/autoinstall.test.ts
+++ b/integration-tests/worker/test/autoinstall.test.ts
@@ -24,7 +24,6 @@ const run = async (attempt) => {
   return new Promise<any>(async (done, reject) => {
     lightning.on('attempt:complete', (evt) => {
       if (attempt.id === evt.attemptId) {
-        console.log(evt.payload);
         done(lightning.getResult(attempt.id));
       }
     });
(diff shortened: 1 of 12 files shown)


integration-tests/worker/test/autoinstall.test.ts
```diff
@@ -0,0 +1,60 @@
...
+    repoDir: path.resolve('tmp/repo/autoinstall'),
+  }));
+});
+
+test.after(async () => {
+  lightning.destroy();
+  await worker.destroy();
+});
+
+test('autoinstall three things at once', async (t) => {
```
josephjclark, line 50: This test  regularly fails against main - on this branch it's stable. I think this is really encouraging.

I think I'll soon move this test (see the comments at the top) but right now I'm happy to leave it - I'm more interested in getting the fix deployed.

PR #514, opened 2023-11-24
Runtime: report memory usage

### LINKED ISSUE #28: Reporting on memory usage
We want to be able to report on memory usage when running jobs.

`node.process` gives us a [memoryUsage](https://nodejs.org/dist/latest-v18.x/docs/api/process.html#processmemoryusage) API which makes it look simple.

But in practice, I'm not sure the best way to handle this. It's probably not helpful to just print what the current usage is?

My feeling is:
* Run `gc()` before each operation 
* Capture memory after GC and at the end of the operation. The diff is presumably the operation cost.
* At the end of the whole job pipeline, we should report the min and max usage (reporting which operation was at max)
* We should handle network traffic in a similar way

A bookmark for later: [node-memwatch](https://github.com/lloyd/node-memwatch)

josephjclark:
## Overview

This PR will:
* Record memory usage at the end of every job
* Log it to stdout (which the worker will pick up and send to lightning as log)
* Emit it as an event (which the worker can pick up and send directly to lightning)

Closes #28

Note: This does NOT limit memory usage or kill a worker which exceeds a limit - that's a separate piece of work.

CLI output:

image

## Details

We use `process.memoryUsage()` to report memory usage to the user.

This is a **snapshot** of memory at the moment we call the function. Which, right now, as at the end of a **job**.

This is not a particularly reliable number and represents final, not peak, memory allocation. See the discussion below.

We get a few numbers of out of `process.memoryUsage()`, see the discussion at the end of the issue. I am choosing to represent `heapUsed` as Job memory - that's how much memory the actual job code took and excludes compiler/runtime overheads. I'm using `rss`, which I think is the memory used by the whole node process, as System memory. The user has no control over this stuff - that's the memory we require to run the job in the first place.

Difficulty: I think `rss` represents the entire Engine memory, ie all the worker threads combined, not just the current thread. If I'm reading the docs right.

The runtime will log memory usage to debug at the end of each job, and emit an event with the same usage in it. The worker will send the memory used (along with the thread id and duration) to Lightning. I am using these values in integration tests to report how much memory each job uses, which is neat.

image

## Testing

I've been testing a bunch of behaviours (including manually triggering garbage collection) in `runtime/test/memory.test.ts`. This is not run CI because it's a bit weird. Use `pnpm test:memory` to run it - I think this is quite a good illustration of some of node's memory management and reporting behaviours.

## Discussion

Reporting accurately is actually going to be super difficult.

`process.memoryUsage()` takes a snapshot at a given moment in time. When garbage collection (gc) runs, memory will be freed up. Which is all well and good, but the memory we report is the final memory (after we've serialised state) - not the peak memory usage.

I am not really sure how useful this final usage figure is.

We could introduce higher resolution recording by taking a snapshot at the end of each operation (not just each job). But this still isn't useful if there are 100 lines of code inside an `fn` block. If gc is run at the end, or in the middle, then the bulk of the memory could be freed up and our reporting is very misleading.

Also note that when the state object is large, serializing it at the end of the job can consume a lot of memory.

Alternatively we could try and call memoryusage on an interval, or maybe even listen to gc events (although we only get an event after gc and may not know how much was freed up). But there is a performance cost with the memoryUsage API, it does take time and effort to read the size of the memory stack. I am not sure yet if it's a SIGNIFICANT performance cost. But even so, I'm not sure what the maximum rate of measuring would be.

## TODO

* [x] Should we always log memory usage? Do we really care about it? It's probably a debug thing right?
* [x] Look at logging in CLI and Worker, is there anything to do here?
* [x] Send memory usage to Lightning (it can present it to user s if we want)
* [x] Tweak ava testing so that only runtime tests use the gc stuff. Ideally actually these tests would be disabled and only run from a special command.

### DIFF
--- a/packages/runtime/test/memory.test.ts
+++ b/packages/runtime/test/memory.test.ts
@@ -44,7 +44,7 @@ const run = async (t, workflow: ExecutionPlan) => {
   const notify = (evt: string, payload: NotifyJobCompletePayload) => {
     if (evt === NOTIFY_JOB_COMPLETE) {
       mem[payload.jobId] = payload.mem;
-      logUsage(t, payload.mem, `final ${payload.jobId}:`);
+      logUsage(t, payload.mem, `job ${payload.jobId}`);
     }
   };
 
@@ -70,7 +70,7 @@ const logUsage = (t: any, mem: Mem, label = '') => {
   // I mean you could lose nearly 500kb of accuracy, that's a lot!
   const job = (mem.job / 1024 / 1024).toFixed(2);
   const system = (mem.system / 1024 / 1024).toFixed(2);
-  t.log(`${label} job: ${job}mb / system ${system}mb`);
+  t.log(`${label}: ${job}mb / system ${system}mb`);
 };
 
 // const jobs = {
@@ -158,10 +158,10 @@ test.serial('report memory usage multiple times', async (t) => {
   };
 
   const { state, mem } = await run(t, plan);
-  logUsage(t, state.a, 'state a:');
-  logUsage(t, state.b, 'state b:');
-  logUsage(t, state.c, 'state c:');
-  logUsage(t, state.d, 'state d:');
+  logUsage(t, state.a, 'state a');
+  logUsage(t, state.b, 'state b');
+  logUsage(t, state.c, 'state c');
+  logUsage(t, state.d, 'state d');
 
   // Each job should use basically the same memory interally (within 2%)
   t.true(roughlyEqual(state.a.job, state.d.job, 0.002));
@@ -170,6 +170,7 @@ test.serial('report memory usage multiple times', async (t) => {
   t.true(roughlyEqual(mem.a.job, state.d.job, 0.002));
 });
 
+// This test shows that creating a large array will increase the memory used by a job
 test.serial('create a large array in a job', async (t) => {
   const plan = {
     jobs: [
@@ -185,8 +186,8 @@ test.serial('create a large array in a job', async (t) => {
   };
 
   const { state, mem } = await run(t, plan);
-  logUsage(t, state.a1, 'state a1:');
-  logUsage(t, state.a2, 'state a2:');
+  logUsage(t, state.a1, 'state a1');
+  logUsage(t, state.a2, 'state a2');
 
   // The second read should have more memory
   t.true(state.a2.job > state.a1.job);
@@ -196,8 +197,8 @@ test.serial('create a large array in a job', async (t) => {
   t.true(mem.a.job > state.a1.job + state.a2.job);
 });
 
-// as before but without using closures (implications for gc I think)
-// The result is basically the same (worth knowing!)
+// This test proves that running a job from a string or by passing in a function
+// doesn't really affect memory usage
 test.serial('create a large array in a job without closures', async (t) => {
   const f1 = `(s) => {
     const mem = process.memoryUsage();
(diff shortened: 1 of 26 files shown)


josephjclark:
I'm ready to call this done, although I haven't done the "tracing" stuff I was talking about in the issue. I don't think it's worth it tbh. The only way to get better results I think is to hammer `process.nextTick` with `process.memoryUsed` and track the highest value. And I'm sure that'll impact on performance. Maybe we can add a mode switch for that, idk.

Note that this is wholly distinct from the memory limits issue. This is a diagnostic to users, nothing more.

@stuartc  you are way ahead of me on this stuff so I'd really appreciate if you could take a close look and let me know what you think. Is this TOO misleading or is it OK as an approximation?

packages/runtime/test/memory.test.ts
```diff
@@ -0,0 +1,323 @@
...
+      },
+    }
+  );
+  return { state, mem };
+};
+
+const logUsage = (t: any, mem: Mem, label = '') => {
+  // What kind of rounding should I Do?
+  // Rounding to an integer is best for humans but I think in these tests we lose a lot of fidelity
+  // I mean you could lose nearly 500kb of accuracy, that's a lot!
```
stuartc, line 75: I think MB to 2 decimal places is probably fine hey? You lose at most 49KB (if my understanding is correct).

stuartc (approved):
This looks like a great start @josephjclark, did you manage to determine the interplay (if any) between the RSS of the thread an the worker itself? 

I guess going forward (like when we kill glutenous jobs), we want to make sure their memory contribution is theirs alone (including the job's context/adaptor code etc). But not including say the websocket, or parent process.

Aside from the :+1: for this PR, do you have any interesting tidbits about the cost of serializing? Like before and after serializing, I assume that _once_ it's a string it's much cheaper but as you have described we're at a very sensitive place when we've serialized _and_ have the original object in memory _and_ we're about to send it back to Lightning.

packages/runtime/test/memory.test.ts
josephjclark: Yep, thanks! I'll tidy this comment up - it's a bit too stream-of-consciousness...

josephjclark:
@stuartc I think I'lo have to create a suite of tests in the engine to explore rss memory across workerpool. I'm pretty sure from the docs that's shared between threads, which means that the "system" memory I'm tracking probably isn't a very useful measure for the engine. The engine itself can report `rss` memory as its total usage, but each thread cannot.

I should probably go and verify that but I'm 90% sure that's the case.

Agreed that we need to ensure we terminate based on "Job memory", which I'm hoping that `heapUsed` is a good proxy for. I'm gonna do memory limits next and I will be explicitly testing that stuff.

Re serialising, a couple of things!
1. I'm trying to avoid the trap of optimising anything right now. When you're looking this closely its very easy to say "Ooh I can save 100mb of memory there" and then lose a week to optimisation. I can't go down that rabbit hole now, there's too much to do.
2. I also know that my serialisation right now is very simple, suboptimal, and I'm not really scrutinising WHEN exactly we serialise. I know it happens several times between the runtime and the worker and I've love one day to just stream state objects straight from the runtime to the websocket through shared memory buffers.
3. That said, when a job completes we DO need to ensure an object is serializable. Right now we do that by serialising it and parsing back whatever is left. It's _that_ step which is spiking the memory right now. I'm actually not too worried about this because we are going to impose limits on the size of the final state object (which I think actually means the size of the serialized string).

PR #517, opened 2023-11-28
Catch async errors (fixes postgres)

josephjclark:
Catch async errors to better handle stuff like the postgres syntax error in #509

Fun one this!

Usually when job or adaptor code throws, the exception is caught in, well, a few places actually. It eventually bubbles back out to the worker where it'll be handled.

But! Those wrapping catches only apply to the original callstack. Any async code which uses setTimeout or process.nextTick or something will throw to the process root and cause an exit.

In other words we can't catch this bad boy and bad things happen:
```
fn((s) => return new Promise(() => {
  setTimeout(() => { throw \"e\" }, 1) }
))
```

The solution is to use the handy `process.on('uncaughtException',fn)` as a catch-all. This lets us catch any off-piste errors and return a nice error event out to the worker.

It's worth noting that if async code runs after the job completes, and happens to throw, that throw will be ignored. The worker basically stops listening after the return. This is a bit of a moot point because the thread will be burned anyway.

### DIFF
--- a/.changeset/twenty-oranges-push.md
+++ b/.changeset/twenty-oranges-push.md
@@ -1,0 +1,5 @@
+---
+'@openfn/engine-multi': patch
+---
+
+Handle async errors in the runtime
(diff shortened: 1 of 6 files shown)


taylordowns2000 (approved):
this is perfect, but i've detected a different problem and am tracking that here: https://github.com/OpenFn/kit/issues/520

Issue #392, opened 2023-09-20
Runtime: Allow (enforce?) jobs to be lazily compiled to support Runs


josephjclark:
Right now, when the runtime executes a workflow, it expects all the expressions (jobs) to be compiled.

The CLI will pre-compile all expressions and log about it before the runtime is invoked.

But Lightning has this notion of a Run, which is the execution of a single job within a workflow.

To properly interface with Lightning, we need compilation to run just before the expression is executed so that it can be associated with a run id. So the job starts, gets compiled, executes, and decides what to do next, all as part of a Run.

At the moment it's kind of impossible to reflect this organisation in the runtime because compilation happens before the job id is created.

This probably isn't necessary for release, but I do think it's really important to make compilation visible.

The solution I think is for the runtime to accept a compile hook, which calls out to whatever runtime manager is calling it. It probably makes sense for the CLI to use lazy compilation as well.

PR #518, opened 2023-11-28
Call run:complete if a workflow errors

### LINKED ISSUE #513: Worker: ensure failed runs are completed
I think we can get a state where after an exception, a run fails. This is caught by worklow-complete (I think).

But the active run is NOT marked as complete! And it'll never close out on Lightning.

So when we get a workflow error we need to remember to send run complete with a suitably critical exit reason

josephjclark:
If a workflow blows up - like through a timeout or OOM error or something severe like that  - we can get into a state where we send an `attempt:complete` back to Lightning, but not a `run:complete`.  Thus the job won't ever finish from Lightning's point of view.

This PR adds a simple indirection in `onWorkflowError`. If there's an active run, we'll close off the run before completing the attempt.

Closes #513


josephjclark:
Blimey, took a while to get this one right :grimacing: 

@taylordowns2000 I don't know if you're able to repro this one for QA  this branch includes other fixes so the postgres thing might not exhibit the same behaviour.

But if you want to take a look, it's ready!

taylordowns2000 (commented):
This looks good, but I do want to note that the job expression `bad` yields what I'd consider a proper exit:
image

And the expression `bad =>` yields:
image

Maybe this is a lightning concern and there are certain types of errors that we need to cater for over there. What say you?

josephjclark:
@taylordowns2000 I think these are both what I expect?

With the syntax error one, it'll die during compilation, which [at the moment] occurs _before_ any actual job execution runs. That's why there's no job logs - we didn't even get to run:start.

That'll improve when we address #392

taylordowns2000 (approved):

PR #519, opened 2023-11-29
build from scratch

delcroip:
## Short Description

installs adsf and pnpm in the docker image


delcroip:
I see, now the image is on dockerhost last time I checked (a month ago) it did not work so I build the image locally

this can be closed then

Issue #520, opened 2023-11-29
Bad things happen when an adaptor calls process.exit()


taylordowns2000:
## Update

What's actually happening here is that the operation throws, which is caught by the adaptor  in `execute`, which calls `process.exit()`, which kills the worker thread and instantly terminate all messaging back to the parent worker.

## Original Issue

Misusing (at least) the `sql` function in language-postgres can cause a worker crash (rather than a run/attempt crash). That's no good.

This expression isn't proper:

```js
sql('select * from food_hygiene_interview');
```

It's not how the `sql` function is meant to be used. Apparently it requires a function which returns a string (like this: ``) rather than a string itself. I'd expect this to either `fail` (or `crash`) at the run (or attempt) level, but not take down the worker.

Here's what happens when we run it:

```
[R/T] ❯ Starting operation 1
TypeError: sqlQuery is not a function
    at /private/tmp/openfn/worker/repo/node_modules/@openfn/language-postgresql_4.1.8/dist/index.cjs:206:20
    at file:///Users/taylor/build/kit/packages/runtime/dist/index.js:577:26
    at async file:///Users/taylor/build/kit/packages/runtime/dist/index.js:549:20
[SRV] ℹ 3a04be88-9cdb-43b1-8320-baf625a29031 :: attempt:log :: OK
[RTE] ✘ Error: Workerpool Worker terminated Unexpectedly
    exitCode: `1`
    signalCode: `undefined`
    workerpool.script: `/Users/taylor/build/kit/packages/engine-multi/dist/worker/worker.js`
    spawnArgs: `undefined`
    spawnfile: `undefined`
    stdout: `[object Object]`
    stderr: `[object Object]`

    at Worker.<anonymous> (/Users/taylor/build/kit/node_modules/.pnpm/workerpool@6.5.1/node_modules/workerpool/src/WorkerHandler.js:297:13)
    at Worker.emit (node:events:513:28)
    at [kOnExit] (node:internal/worker:279:10)
    at Worker.<computed>.onexit (node:internal/worker:199:20)
[RTE] ❯ Purging workers
[SRV] ℹ 3a04be88-9cdb-43b1-8320-baf625a29031 :: workflow-error :: OK
```

josephjclark:
I thought I was losing my freaking mind here - then I noticed this in the postgres adaptor:
```
export function execute(...operations) {
  const initialState = {
    references: [],
    data: null,
    queries: [],
  };

  return state => {
    return commonExecute(
      createClient,
[...]

PR #522, opened 2023-11-29
Handle process exit 

josephjclark:
So it turns out that some adaptors trigger a process.exit and that causes us all sorts of problems - see #520 for details and a reproduction.

There is a limit to what we can do about this. Inside the worker we can listen for `process.on('exit')`, but we have to run synchronous code in there. We don't actually seem to be able to send a message out of the worker thread.

So instead, outside the thread in the main process, we keep an eye out for workers which throw with a message like `Workerpool Worker terminated Unexpectedly`. Looking at the workerpool source I'm reasonably confident that this is only triggered by process.exit.

So we can detect an early exit and report a vague exit reason. I'm calling it a `kill` (although it's an internal rather than external kill)

Lightning should now get an attempt:complete with this reason:
```
{
"reason":"kill",
"error_type":"ExitError",
"error_message":"Process exited with code: 1"
}
```
Bonus fix: I caught a small issue with some prior error handling. A promise inside the worker which throws an uncaught error is never resolved, and so the engine never releases the worker . 

To fix this, I'm forcing a process.exit just to be absolutely sure the worker stops executing. This kind of undermines the purge policy but whatever - all this is going to be ripped out soon anyway (probably).

### DIFF
--- a/.changeset/violet-ligers-walk.md
+++ b/.changeset/violet-ligers-walk.md
@@ -1,0 +1,5 @@
+---
+'@openfn/engine-multi': patch
+---
+
+Throw a better error on process.exit
--- a/integration-tests/worker/test/exit-reasons.test.ts
+++ b/integration-tests/worker/test/exit-reasons.test.ts
@@ -100,3 +100,23 @@ test('kill: oom', async (t) => {
   t.is(error_type, 'OOMError');
   t.is(error_message, 'Run exceeded maximum memory usage');
 });
+
+test('crash: process.exit() triggered by postgres', async (t) => {
+  const attempt = {
+    id: crypto.randomUUID(),
+    jobs: [
+      {
+        adaptor: '@openfn/language-postgresql@4.1.8', // version number is important
+        body: "sql('select * from food_hygiene_interview');",
+      },
+    ],
+  };
+
+  const result = await run(attempt);
+
+  const { reason, error_type, error_message } = result;
+
+  t.is(reason, 'crash');
+  t.is(error_type, 'ExitError');
+  t.regex(error_message, /Process exited with code: 1/i);
+});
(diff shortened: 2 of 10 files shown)


taylordowns2000 (changes requested):
hey @josephjclark , the code works but in my mind `kill` is reserved for when _we_ stop a customer's attempt. if we decide to kill their attempt, it was our choice and we tell them why we did it.

can you call this a `crash` instead?

from my perspective, our explanation is this: "hey, in v1 this would have been a normal exit and your downstream jobs "on failure" would still execute. we can't do that here given the old version of the adaptor you're using, but in all new adaptors versions you can expect this to be a normal failure. so for now, we'll call it a `crash` since it's in the category of user/adaptor/destination-database error and we're _not_ going to continue on with your other work in this attempt."

josephjclark:
@taylordowns2000 well in my mind that process exit IS us killing the attempt (and will go away soon).

But sure, happy to call it a crash! I'll push that shortly and merge

taylordowns2000:
thank you! re: your point above... i think this was just bad adaptor design by me in 2020.

for the record, the problem we solved (quite inelegantly!) back then was that certain errors for adaptors that create and then use a database connection for their duration would not (a) properly close the DB connection and (b) not properly terminate the process.

1. https://github.com/OpenFn/language-mssql/releases/tag/v2.1.1
2. https://github.com/OpenFn/language-postgresql/pull/18

Looking back at this, I'm a little startled that I didn't call `disconnect` and `cleanupState` inside that catch. Also, why did I call `process.exit(1)` rather than `throw` _after_ I'd properly disconnected and cleaned up state?

Anyway... this process.exit(1) _only_ gets reached when a user error made the job blow up. So in my mind, this isn't a kill. It's a `crash` where stepped in at the 11th hour to do some cleaning up and then let it die. (or maybe "euthanized it" is a better term with that `process.exit(1)` there.) but since we don't have `assisted crash` let's go with `crash`.

(I know you're already on board @josephjclark , just wanted to have this written out somewhere so we can see what problem we were trying to solve and why.)

PR #526, opened 2023-12-04
Worker: Preserve event order and Print Versions

### LINKED ISSUE #413: engine: log versions
The CLI log versions of key components at the start of a run (cli/runtime/compiler/node/adaptor)

The engine needs to do the same thing. We may be able to steal some code.

Note that this is done at the start of a job, not an attempt. We may need a callback in the runtime for that (unless we cheat a bit).

Maybe the attempt can also log some versions - worker, node, engine. Then the job would log node, runtime, adaptor. 

Compiler version right now will be in the attempt, but we can move it to the run when we sort out #392.

josephjclark:
There are three changes in this PR:

* Appends versions numbers to the `run:start` event
* Prints version numbers to the log when a run starts
* Throttles worker events (ie, each is blocking and executes sequentially) to preserve original order coming out of the runtime

## Issue

Closes #413

## Event Throttling

See the comment below - but the worker should now enforce that events emitted from the engine are "blocking" and re-emitted in sequence.

This should mean that the order of events from the runtime is exactly the same as the order of events sent to Lightning.

## Ordering

EDIT: this is fixed by locking ordering of events (although I'm still fudging the timestamp on the version log!)

We have a problem right now where I cannot make the versions the first thing to appear in the logs. In fact, the version log may be rejected and throw an error because the run could be finished by the time we log the version!

image

It's all do to with the asynchronous master-slave relationship between the worker and runtime.

We listen for the runtime to broadcast a run-start event, send our own run:start to lightning, then when the run is ready,  we issue a version log event.

But this is all asynchronous code - while the worker is waiting for lightning to acknowledge the run-start event, the runtime is busily executing and sending out its own events. So the version stamp is way behind the first runtime execution logs.

I was kinda hoping that the order would sort itself out, but it does not.

At the time of writing I am forcing an offset on the version log timestamp so that lightning thinks its earlier than it really is. This is a blatant hack. Also it doesn't work for streaming, because runtime logs still come through early.

Solutions?

* Accept that the live stream will be out of order and celebrate that the 
* Defer sending logs to lightning until the version log has been acknowledged. The difficulty with this is that the log uses the timestamps on the runtime events, right, which are issued BEFORE the version log. So unless we overwrite the timestamps, this doesn't help us.
* Build versions straight in to the Lightning UI (I think this is way better UX anyway)
* Rethink this whole strategy and go back to sending versions from the runtime (but we still need to wait for run-start to return!)

## Adaptor Versions

I wanted the adaptor versions in particular to be as close to the autoinstall as possible, so that they report what is _actually_ used rather than what was asked for. So the autoinstall function in the engine will return the number. It doesn't scrape it from the packag.ejson, but it does use the same number used in the `npm install` command. So I think it's reasonably robust.

## Refactoring

I've taken the opportunity to do a little refactoring here, improving the structure of the worker a bit. I don't intend to finish that work in this PR - although I might nibble at it some more.

In particular I've split up the worker's `api/execute` file and started pulling out the event handlers into seperate  files. I think this makes it much easier to ask the question "so what happens when a run starts?", because you just look in `src/events/run-start.ts`.

### DIFF
--- a/packages/ws-worker/src/api/execute.ts
+++ b/packages/ws-worker/src/api/execute.ts
@@ -186,9 +186,9 @@ export function onJobError(context: Context, event: any) {
   // because it'll count it as a crash
   // This isn't very good: maybe we shouldn't trigger an error
   // at all for a fail state?
-  const { state = {}, error, jobId } = event;
+  const { state, error, jobId } = event;
   // This test is horrible too
-  if (state.errors?.[jobId]?.message === error.message) {
+  if (state?.errors?.[jobId]?.message === error.message) {
     return handleRunComplete(context, event);
   } else {
     return handleRunComplete(context, event, event.error);
--- a/.changeset/eighty-moons-smile.md
+++ b/.changeset/eighty-moons-smile.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Fix an issue where an error is thrown if a job does not return state
--- a/.changeset/ten-carpets-mix.md
+++ b/.changeset/ten-carpets-mix.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Better handling of job-error
--- a/integration-tests/worker/test/integration.test.ts
+++ b/integration-tests/worker/test/integration.test.ts
@@ -338,13 +338,13 @@ test('a timeout error should still call run-complete', (t) => {
         },
       ],
       options: {
-        // Including the timeout here stops the attempt returning at all
         timeout: 100,
       },
     };
 
     lightning.once('run:complete', (event) => {
       t.is(event.payload.reason, 'kill');
+      t.is(event.payload.error_type, 'TimeoutError');
     });
 
     lightning.once('attempt:complete', (event) => {
--- a/packages/cli/src/execute/handler.ts
+++ b/packages/cli/src/execute/handler.ts
@@ -58,12 +58,12 @@ const executeHandler = async (options: ExecuteOptions, logger: Logger) => {
     const result = await execute(input!, state, options);
     await serializeOutput(options, result, logger);
     const duration = printDuration(new Date().getTime() - start);
-    if (result.errors) {
+    if (result?.errors) {
       logger.warn(
         `Errors reported in ${Object.keys(result.errors).length} jobs`
       );
     }
-    logger.success(`Finished in ${duration}${result.errors ? '' : ' ✨'}`);
+    logger.success(`Finished in ${duration}${result?.errors ? '' : ' ✨'}`);
     return result;
   } catch (err: any) {
     if (!err.handled) {
--- a/packages/cli/test/commands.test.ts
+++ b/packages/cli/test/commands.test.ts
@@ -123,6 +123,12 @@ test.serial('run a job with defaults: openfn job.js', async (t) => {
   t.assert(result.data.count === 42);
 });
 
+test.serial('run a job which does not return state', async (t) => {
+  const result = await run('openfn job.js', 'export default [s => {}]');
+
+  t.falsy(result);
+});
+
 test.serial('run a workflow', async (t) => {
   const workflow = {
     jobs: [
--- a/packages/cli/test/execute/execute.test.ts
+++ b/packages/cli/test/execute/execute.test.ts
@@ -255,3 +255,16 @@ test('run a job without compilation', async (t) => {
   const result = await handler(options, logger);
   t.is(result.data.count, 42);
 });
+
+test('run a job which does not return state', async (t) => {
+  const job = `${fn}fn(() => {});`;
+  const options = {
+    ...defaultOptions,
+    job,
+  };
+  const result = await handler(options, logger);
+  t.falsy(result);
+
+  // Check that no error messages have been logged
+  t.is(logger._history.length, 0);
+});
--- a/packages/engine-multi/test/integration.test.ts
+++ b/packages/engine-multi/test/integration.test.ts
@@ -169,6 +169,29 @@ test.serial('compile and run', (t) => {
   });
 });
 
+test.serial('run without error if no state is returned', (t) => {
+  return new Promise(async (done) => {
+    api = await createAPI({
+      logger,
+    });
+
+    const plan = createPlan([
+      {
+        expression: `${withFn}fn(() => {})`,
+      },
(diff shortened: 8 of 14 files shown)


josephjclark:
TODO - I dithered over this a bit but I WILL add the runtime version too - it'll be useful for our debugging

josephjclark:
TODO - fix the indenting

packages/ws-worker/src/api/execute.ts
```diff
@@ -155,25 +160,14 @@ export const sendEvent = <T>(channel: Channel, event: string, payload?: any) =>
     channel
       .push<T>(event, payload)
       .receive('error', reject)
-      .receive('timeout', () => reject(new Error('timeout')))
+      .receive('timeout', (e) => {
+        console.log(e);
```
josephjclark, line 164: TODO: revert this

josephjclark:
TODO: I am seeing problems in my local lightning where run logs seem to be duplicated? Here's the output I keep seeing (this is after a single run triggered from the editor)

I think this is a lightning frontend bug but I need to look a bit more closely to be sure

```
Versions:
      ▸ node.js     18.12.1
      ▸ worker      0.2.12
      ▸ engine      0.2.2
      ▸ compiler    workspace:*
    ▸ runtime     workspace:*
Starting job 6328285c-fd51-4ee5-bd31-9217df1e4547
Intialising pipeline
Timeout set to 300000ms
[linker] loading module @openfn/language-salesforce
[linker] Loading module @openfn/language-salesforce from /tmp/openfn/worker/repo/node_modules/@openfn/language-salesforce_4.3.0/dist/index.cjs
Resolved adaptor @openfn/language-salesforce to version 4.3.0
Executing expression (1 operations)
Versions:
      ▸ node.js     18.12.1
      ▸ worker      0.2.12
      ▸ engine      0.2.2
      ▸ compiler    workspace:*
    ▸ runtime     workspace:*
Starting operation 1
Operation 1 complete in 0ms
Expression complete!
Completed job 6328285c-fd51-4ee5-bd31-9217df1e4547 in 1.058s
Final memory usage: [job 23mb] [system 284mb]
Starting job 818a0aaa-5825-4dc1-87cb-2fd930fe4e5e
Intialising pipeline
Timeout set to 300000ms
[linker] loading module @openfn/language-http
[linker] Loading module @openfn/language-http from /tmp/openfn/worker/repo/node_modules/@openfn/language-http_5.1.0/dist/index.cjs
Resolved adaptor @openfn/language-http to version 5.1.0
Executing expression (1 operations)
Starting operation 1
Operation 1 complete in 0ms
Expression complete!
Completed job 818a0aaa-5825-4dc1-87cb-2fd930fe4e5e in 58ms
Versions:
      ▸ node.js     18.12.1
      ▸ worker      0.2.12
      ▸ engine      0.2.2
      ▸ compiler    workspace:*
    ▸ runtime     workspace:*
Final memory usage: [job 37mb] [system 285mb]
Starting job 0eeeffe2-ae43-4614-9b7e-c774dfd11c40
Intialising pipeline
Timeout set to 300000ms
[linker] loading module @openfn/language-http
[linker] Loading module @openfn/language-http from /tmp/openfn/worker/repo/node_modules/@openfn/language-http_5.1.0/dist/index.cjs
Resolved adaptor @openfn/language-http to version 5.1.0
Executing expression (1 operations)
Starting operation 1
Operation 1 complete in 0ms
Expression complete!
Completed job 0eeeffe2-ae43-4614-9b7e-c774dfd11c40 in 2ms
Final memory usage: [job 37mb] [system 285mb]
```

josephjclark:
Also TODO: why am I seeing this locally?
```
[SRV] ✘ Error: timeout
    at Object.callback (file:///d/repo/openfn/kit/packages/ws-worker/src/api/execute.ts:165:16)
    at file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/push.js:76:23
    at Array.forEach (<anonymous>)
    at Push.matchReceive (file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/push.js:76:8)
    at Object.callback (file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/push.js:107:12)
    at Channel.trigger (file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/channel.js:278:12)
    at Push.trigger (file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/push.js:126:18)
    at Timeout.<anonymous> (file:///d/repo/openfn/kit/node_modules/.pnpm/phoenix@1.7.7/node_modules/phoenix/assets/js/phoenix/push.js:111:12)
    at listOnTimeout (node:internal/timers:564:17)
    at processTimers (node:internal/timers:507:7)
```

How can the event time out if it's been accepted?

josephjclark:
I really want to merge the refactors on this branch, but I don't particularly want to merge the broken version log. Hmm.

josephjclark:
For now:

* Make sure the version logs include the run id in the data
* And put the run id in the actual text
* Just let them be logged in any order
* Don't worry about the refresh

josephjclark:
Alright, here's something fun.

I've added a "throttler" to the event bus between the engine and the worker.

Basically the worker listens to events in coming from the engine - like `job-start` or `log` - and asynchronously does stuff and speaks to lightning.

I'm pretty sure there are windows here where events can creep through in the wrong order - maybe even cause an attempt to close before all of its logs are processed (certainly we see this in error cases).

The throttle makes the worker's events handlers  blocking: while it's handling say, a log event, it won't process any other events that come in from the engine - like the starting of the next run.

Maybe I shouldn't have bound these features together BUT the theory is that the throttle causes events to block. When run:start fires, it sends out the version log, and in the meantime all other version logs are blocked. So this should help keep the ordering of the  version log much more robust.

josephjclark:
Great - yes, actually, now that you mention it I do see the same thing. I worried for a moment but it's actually quite reassuring that we see the same thing.

Also this is consistent with the runtime! It logs that starting line before sending the job-start event. I actually disagree with that but it's reflecting the true log ordering, which is exactly what this PR now does.

Issue #539, opened 2023-12-08
Publish Action has suddenly stopped working


josephjclark:
Apparently the publish action has stopped working, literally overnight, which means we can't publish new packages.

Last night we released just fine, today we can't:

![image](https://github.com/OpenFn/kit/assets/7052509/1e151319-4daf-486c-8784-bc46a00d5d8b)

The error is on the test step:

![image](https://github.com/OpenFn/kit/assets/7052509/b3757453-9c7a-4231-b32c-62f4d1526044)

Locally, I can reproduce this by removing the `'--loader=ts-node/esm'` arugment from the root ava.config.js. That makes me think that somehow, in the action environment, ts-node isn't getting loaded.

Need to be tracking this in an issue because I don't think I'm going to be able to fix it today.

The issue was introduced by merging this PR: https://github.com/OpenFn/kit/pull/537

josephjclark:
Both actions use runner version `2.311.0`

josephjclark:
Still failing after a couple of days of downtime. Pushing an empty commit didn't help.

josephjclark:
A trivial bump of package.json, which should change the git and npm checksums and break any caching in CI, also doesn't help.

josephjclark:
Here's a clue.

The build step for `describe-package` on the last good run said this:

```
packages/describe-package build: > @openfn/describe-package@0.0.18 build:worker /home/runner/work/kit/kit/packages/describe-package
packages/describe-package build: > tsm esbuild-worker.ts
packages/describe-package build: (node:2170) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
packages/describe-package build: (Use `node --trace-warnings ...` to show where the warning was created)
packages/describe-package build: (node:2170) DeprecationWarning: Obsolete loader hook(s) supplied and will be ignored: getFormat, transformSource
packages/describe-package build: Done
```

But on the dead builds, it says this:
```
packages/describe-package build: > tsm esbuild-worker.ts
packages/describe-package build: (node:2158) ExperimentalWarning: `--experimental-loader` may be removed in the future; instead use `register()`:
packages/describe-package build: --import 'data:text/javascript,import { register } from "node:module"; import { pathToFileURL } from "node:url"; register("file%3A///home/runner/work/kit/kit/node_modules/.pnpm/tsm%402.2.2/node_modules/tsm/loader.mjs", pathToFileURL("./"));'
[...]

PR #541, opened 2023-12-11
Support node 18.19 in ava

josephjclark:
Node 18.19 seems to break our ava tests. Something to do with the module loader.

Fixes my worst enemy, #539 

Fix to follow.

First commit should fail in circle CI.


josephjclark:
Confirmed - on 18.19.0 circle CI now breaks in the same way as the github action.

I'm still not sure WHAT this is, although 18.19.0 did touch the module loading code. But there should be a better way to setup ava which will make this go away.

However, that approach is refusing to work...

PR #545, opened 2023-12-12
Engine child process

josephjclark:
Rough cut of an alternative engine which runs all attempts in a child process

* instead of calling `workerpool.exec`, we now call my own child-process-worker. I've tried to match the workerpool API to keep things simple.
* My worker will create and then destroy a forked child process for each attempt, so each attempt runs totally in isolation
* It's a rough implementation - I haven't mapped all the error handling stuff, as it's not really needed for these tests

### DIFF
--- a/integration-tests/worker/test/benchmark.test.ts
+++ b/integration-tests/worker/test/benchmark.test.ts
@@ -8,7 +8,7 @@ import { run, humanMb } from '../src/util';
 let lightning;
 let worker;
 
-const maxConcurrency = 1;
+const maxConcurrency = 5;
 
 test.before(async () => {
   const lightningPort = 4322;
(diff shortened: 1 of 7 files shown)


josephjclark:
Just for the record, here is the benchmark table with worker threads:

| Workers     | # attempts | duration | max job memory (mb) | max system memory (mb) |
|-------------|:----------:|:--------:|:-------------------:|:----------------------:|
| 50          |     100    |   4.555  |          23         |          1612          |
| 20          |     100    |   3.79   |          22         |           979          |
| 5 (default) |     100    |   3.295  |          23         |           452          |
| 1           |     100    |  41.377  |          19         |           362          |

The key here is 3 seconds with 5 threads.

josephjclark:
Not the best benchmark but:

Polling mpstat every second shows CPU utilisation around 50% on main:

image

On this branch, that spikes to 67%, and stays high longer, under the same test:

image

PR #547, opened 2023-12-14
Fancy new engine

### LINKED ISSUE #412: engine: replace workerpool
Workerpool is not going to fly for us because:

* We don't have two-way communication with the worker, so we cannot lazy load credentials and dataclips
* We do not have control over purging specific threads to clean memory leaks  or isolate modules
* A future worry but we have no capacity to re-use a specific thread, like re-running a workflow in the same thread twice to reduce module initialisation time. IF we can safely isolate modules (even in the sandbox), a warm thread/environment is a great optimisation
* Running two workflows in the same pooled thread is insecure (and may be unstable, depending on module initialisation)
* If a worker blows up, the parent thread - the whole engine! - also dies

We need to investigate threads and piscina (or whatever), or maybe other libraries, so that we get the control we need.

We need to consider performance, stability and security. I'd like to have some kind of benchmark for this (a benchmark integration test would be nice - run the same 10 workflows 50 times and measure the total time).

josephjclark:
This PR adds a fancy new child-process pool, replacing workerpool.

Closes #412 

All the relevant workerpool tests still pass against my new child process, so I'm fairly confident nothing is lost. And I'm sure I've introduced some fresh new bugs.

The next step is for the child processes to run each attempt inside a worker_thread, guaranteeing isolation even outside the sandbox.

Almost all the tests are passing - I think I have one fail and an OOM explosion.

* [x] Fix memory reporting 
* [x] Reinstate memory limits
* [x] Spin out a worker thread for each actual attempt execution 
* [x] Remove all purge logic
* [x] Remove or map any other functionality related to workerpool (process.exit maybe)
* [x] Fix typings
* [x] Migrate any old/skipped tests to new architecture


josephjclark:
I'm a bit concerned that tests are regularly failing with OOM fails when all run together. When run standalone they're OK.

To be fair we're spinning up a lot of processes, and because of how ava tests its processes within processes really.

Still, I'm going to have to look into how to keep it stable. Perhaps re-using the pool more, rather than creating quite so many pools, will help (although of course that brings a risk of tests interfering with each other)

josephjclark:
Forced the tests to run serially - and no more oom errors! Well, hopefully...

josephjclark:
TODO: check the packed package to make sure the contents are correct

josephjclark:
I've now implemented the worker-thread architecture.

So now we have a pool of child threads, and when you call `exec` on that pool, it'll find an available thread and pass the exec arguments in. Then inside that thread, we spin off a worker thread to do the actual execution - ensuring a perfect, neatly encapsulated sandbox.

Now I've got a whole bunch of  tidying and testing to do on that.

josephjclark:
New  benchmarks coming in. Performance seems comparable to main - perhaps a little more memory usage per attempt (makes sense!) atlhough I don't really trust the memory readings I get.

Unlike with the pure child process approach, CPU and memory usage is much less intensive.

```
| Workers     | # attempts | duration | max job memory (mb) | max system memory (mb) | est total memory |
|-------------|:----------:|:--------:|:-------------------:|:----------------------:|------------------|
| 50          |     100    |    8.0   |          24         |           96           | 4800             |
| 20          |     100    |   7.67   |          24         |           97           | 1920             |
| 5 (default) |     100    |   10.6   |          24         |           97           | 480              |
| 1           |     100    |   42.4   |          24         |           97           | 97               |
```

These are the same numbers as main:
```
| Workers     | # attempts | duration | max job memory (mb) | max system memory (mb) |
|-------------|:----------:|:--------:|:-------------------:|:----------------------:|
| 50          |     100    |   4.555  |          23         |          1612          |
| 20          |     100    |   3.79   |          22         |           979          |
| 5 (default) |     100    |   3.295  |          23         |           452          |
| 1           |     100    |  41.377  |          19         |           362          |
```

With the bonus that it's got an ironclad sandbox, is robust to OOM errors, and can reclaim memory more effectively.

josephjclark:
Critical outstanding fixes here:
 * [ ] Fix or skip that flaky test
 * [ ] #569
 * [ ] #568

josephjclark:
This is almost ready to go, although a  couple of tests are being very very flaky in CI. Unsure how worried I should be about this.

PR #552, opened 2023-12-23
Handle initial edge condition and log edge evaluations

### LINKED ISSUE #551: Handling first edge condition
Edge conditions from the starting trigger to the first job in the workflow don't appear to be respected.

While we're in here, it would also be AMAZING to add a single log line to the attempt (not the run) which says: "Edge condition <XYZ> evaluated to `true`; starting job <ABC>" Or "Edge condition <XYZ> evaluated to `false`; not executing job <ABC>"

Cc @josephjclark , @jyeshe - I think this behavior should be locked down in the e2e tests when delivered.

Ooooh... update:

Check out this execution plan... I'm not seeing any `startCondition` in there. Two possibilities:

1. It's not being sent from Lightning? @jyeshe ?
2. It's not being read/handled by Kit? @josephjclark ?

```js
converted attempt as execution plan: {
  "id": "01d5042a-42e6-480c-86ed-111fb319ecd1",
  "initialState": "c78db39d-46c2-4705-bbfb-34e9146c69ca",
  "start": "cae544ab-03dc-4ccc-a09c-fb4edb255d7a",
  "jobs": [
    {
      "id": "cae544ab-03dc-4ccc-a09c-fb4edb255d7a",
      "next": {
        "d2a72d03-61b4-4783-90bf-38940edcf1e5": true
      }
    },
    {
      "id": "d2a72d03-61b4-4783-90bf-38940edcf1e5",
      "configuration": null,
      "expression": "fn(state => state);\n",
      "adaptor": "@openfn/language-http@5.1.1",
      "next": {
        "11d22939-3fc4-49f9-83c9-4d2248302c91": {
          "condition": "state.funky == \"seven\""
        }
      }
    },
    {
      "id": "11d22939-3fc4-49f9-83c9-4d2248302c91",
      "configuration": null,
      "expression": "fn(state => state);\n",
      "adaptor": "@openfn/language-http@5.1.1",
      "next": {
        "07ebd0e8-bbc8-45ed-a3aa-32b754c50f51": {
          "condition": "Boolean(!state?.errors?.[\"11d22939-3fc4-49f9-83c9-4d2248302c91\"] ?? true)"
        },
        "2a8f6ff8-dff6-430f-9d22-cadb0baf565f": {
          "condition": "Boolean(state?.errors && state.errors[\"11d22939-3fc4-49f9-83c9-4d2248302c91\"])"
        }
      }
    },
    {
      "id": "07ebd0e8-bbc8-45ed-a3aa-32b754c50f51",
      "configuration": null,
      "expression": "fn(state => state);\n",
      "adaptor": "@openfn/language-http@5.1.1"
    },
    {
      "id": "2a8f6ff8-dff6-430f-9d22-cadb0baf565f",
      "configuration": null,
      "expression": "fn(state => state);\n",
      "adaptor": "@openfn/language-http@5.1.1"
    }
  ]
}
```

taylordowns2000:
This handles the initial edge condition from a webhook trigger and logs edge evaluations to the attempt

## Related issue

Fixes #551

- [x] I have added unit tests

## Release branch checklist

If this IS a release branch:

- [x] Run `pnpm changeset` from root to bump versions
- [x] Run `pnpm install`
- [x] Commit the new version numbers
- [x] Run `pnpm changeset tag` to generate tags
- [x] Push tags `git push --tags`

### DIFF
--- a/packages/runtime/CHANGELOG.md
+++ b/packages/runtime/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @openfn/runtime
 
+## 0.2.3
+
+### Patch Changes
+
+- f228fd5: Add support for initial edge conditions in worker and add edge evaluation logging to runtime
+
 ## 0.2.2
 
 ### Patch Changes
--- a/packages/runtime/CHANGELOG.md
+++ b/packages/runtime/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Patch Changes
 
-- f228fd5: Add support for initial edge conditions in worker and add edge evaluation logging to runtime
+- f228fd5: Add edge evaluation logging to runtime
 
 ## 0.2.2
 
--- a/packages/runtime/src/execute/job.ts
+++ b/packages/runtime/src/execute/job.ts
@@ -67,11 +67,11 @@ const calculateNext = (job: CompiledJobNode, result: any, logger: Logger) => {
           } catch (e: any) {
             throw new EdgeConditionError(e.message);
           }
+          logger.always(
+            `Edge ${edge.condition.toString()} evaluated to true; will execute job ${nextJobId} next`
+          );
         }
       }
-      logger.always(
-        `Edge ${edge.condition.toString()} evaluated to true; will execute job ${nextJobId} next`
-      );
       next.push(nextJobId);
       // TODO errors
     }
--- a/packages/runtime/src/execute/job.ts
+++ b/packages/runtime/src/execute/job.ts
@@ -60,7 +60,7 @@ const calculateNext = (job: CompiledJobNode, result: any, logger: Logger) => {
           try {
             if (!edge.condition(result)) {
               logger.always(
-                `Edge ${edge.condition.toString()} evaluated to FALSE; will not start job ${nextJobId}`
+                `Edge ${edge.condition.toString()} returned false; ${nextJobId} will NOT be executed`
               );
               continue;
             }
@@ -68,7 +68,7 @@ const calculateNext = (job: CompiledJobNode, result: any, logger: Logger) => {
             throw new EdgeConditionError(e.message);
           }
           logger.always(
-            `Edge ${edge.condition.toString()} evaluated to true; will execute job ${nextJobId} next`
+            `Edge ${edge.condition.toString()} returned true; ${nextJobId} will be executed next`
           );
         }
       }
--- a/packages/runtime/src/execute/job.ts
+++ b/packages/runtime/src/execute/job.ts
@@ -59,7 +59,7 @@ const calculateNext = (job: CompiledJobNode, result: any, logger: Logger) => {
         if (typeof edge.condition === 'function') {
           try {
             if (!edge.condition(result)) {
-              logger.always(
+              logger.debug(
                 `Edge ${edge.condition.toString()} returned false; ${nextJobId} will NOT be executed`
               );
               continue;
@@ -67,7 +67,7 @@ const calculateNext = (job: CompiledJobNode, result: any, logger: Logger) => {
           } catch (e: any) {
             throw new EdgeConditionError(e.message);
           }
-          logger.always(
+          logger.debug(
             `Edge ${edge.condition.toString()} returned true; ${nextJobId} will be executed next`
           );
         }
--- a/packages/runtime/src/execute/job.ts
+++ b/packages/runtime/src/execute/job.ts
@@ -43,7 +43,7 @@ const loadState = async (
   return job.state;
 };
 
-const calculateNext = (job: CompiledJobNode, result: any) => {
+const calculateNext = (job: CompiledJobNode, result: any, logger: any) => {
   const next: string[] = [];
   if (job.next) {
     for (const nextJobId in job.next) {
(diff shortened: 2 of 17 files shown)


taylordowns2000:
hey @josephjclark , this is built and working properly on demo.openfn.org . if it looks OK to you please feel free to merge to `main`

packages/runtime/src/execute/job.ts
```diff
@@ -58,11 +59,17 @@ const calculateNext = (job: CompiledJobNode, result: any) => {
         if (typeof edge.condition === 'function') {
           try {
             if (!edge.condition(result)) {
+              logger.always(
```
josephjclark, line 62: We shouldn't log this stuff to `always`, it's not important enough. `always` is for critical information that users need to know.

If the system is trusted and working, you know the result of a condition because the next job did or did not fire. This log line imparts no new information - it just confirms that failed edges had their expression tested.

I think we should log this to `debug`. It's the kind of thing you might want to see when something goes wrong and you're trying to trace exactly what happened.

Ditto for the true condition.

josephjclark: There's a bit more to think about really.

The condition that gets printed here may not be the same thing the user entered in Lightning. In fact the conditions we generate right now are very far removed from what's in the lightning data and UI. So this may not be useful for end users. Maybe that will get better in time.

Also the condition may not return `true` or `false`, but truthy or falsy values (especially with custom expressions). To the language may be a little confusing.

We can do this in the short term, but in the medium term I think we need to use the condition label (or if not present, use the two job names `job-1 -> job-2`)

packages/runtime/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # @openfn/runtime
 
+## 0.2.3
+
+### Patch Changes
+
+- f228fd5: Add support for initial edge conditions in worker and add edge evaluation logging to runtime
```
josephjclark, line 7: This should be two different patch notes - changes to the worker should not be reflected in the runtime's log.

I'm happy to jump in and change this once we agree on the actual logging. This comment is kinda just a reminder to me.

taylordowns2000:
Both comments are super interesting! I'd like to loop @aleksa-krolls in here (cc @NickOpenFn ) to talk about when/how the edge conditions need to be displayed to the end user.

In my opinion, this is a really important part of the business record. Especially because we have a little bit of misdirection (ok, maybe a harsh word choice but also want to convey my opinion on the subject!) in Lightning where we hide the actual condition expression in some cases ("always", "on fail", etc.) I think it's really important to always display this information to the end user so they know why the workflow got executed in the way it did. It would also be wonderful to be able to search through the logs for an old, bad edge condition and reprocess 1000s of work orders which ran using the old condition after updating the condition. Basically, I think it's critical to see the edge condition as it is truly evaluated in the logs.

Now to argue the other side... yes, this is almost by definition "debug" information. Continuing my line of thinking above, it's critical for _debugging_ why a workflow didn't behave as expected.

Given that we don't yet have a facility that allows end users to adjust log visibility in lightning, I'd argue that we have to leave this as always for now, but that as soon as a user can go back to an old attempt and change the log level that gets displayed in the UI then we should make this "debug" level.

Thoughts @josephjclark, @aleksa-krolls , @NickOpenFn ?

josephjclark:
@taylordowns2000 I think this is getting over-egged a bit. It's not big design question.

Can we just log to `info` or `debug` please? It's really not an `always`.

Users will see both levels  by default in Lightning today, so it doesn't hide anything. In future we'll still be able to search for these logs in the business record, even if they're filtered out by the default view.

If you wish to include the expression source that's absolutely fine - I just don't think it's going to be terribly helpful in practice. Happy to leave it in and see how we feel about it in a few months.

taylordowns2000:
All makes sense. I didn't realize we send debug logs to Lightning!

Yes, we can absolutely change this to debug. (You're welcome to do so now, otherwise I'll likely get to this tomorrow AM so no rush.)

Nice to have this convo here for the future. Thanks, Joe.

taylordowns2000:
Changes made @josephjclark . Should be ready for merge and for a patch to be released on the runtime which notes the `always -> debug` change.

> Also for the record, I agree with this:
> 
> > think it's really important to always display this information to the end user so they know why the workflow got executed in the way it did.
> 
> But!
> 
> a) We might want to think about more debug logging around edge conditions to make it clearer what's happening and what decisions the runtime is making. That needs some thinking about but I can see some value in that.
> 
> b) To understand _why_ an edge condition evaluated a certain way, you need to know what the state object looks like when the condition runs. You may even need to drop a breakpoint on it (which is a nice future challenge for the CLI). I mean the expression itself only tells you so much (and right now the expression we run is actually pretty complex and it may be hard for users to understand what it's even doing)

And for this one, yes... more design required. For now I'm hoping that since in the attempt viewer the user can easily view the output state for step 1 and (now) the edge condition for the edge between step 1 and step 2, they'll have what they need to understand why step 2 was or wasn't executed.

image

In the image here, we're checking "output" for step 2 to see why step 3 got executed.

josephjclark (approved):

PR #555, opened 2024-01-10
Deploy with condition type, expression and label

### LINKED ISSUE #553: cli-deploy-action failure on demo-openhie repo
### User story

Lightning allows users to sync the project from github repo to Lightning database. 

The demo-openhie has deploy.yml GH action that calls the [CLI](https://github.com/OpenFn/demo-openhie/blob/0a585e44043fa84b1b55c691e67956c1e7c7946f/.github/workflows/deploy.yml#L12). 

When `deploy.yml` this [error](https://github.com/OpenFn/demo-openhie/actions/runs/7298571555/job/19889802383) shows that the CLI needs to be adapted to differentiate a condition type from a condition expression:

```
Project found.
[CLI] ♦ Changes:
 {
   workflows: [
     {
       edges: [
         {
-          condition_type: "always"
+          condition: "always"
+          enabled: true
         }
         {
-          source_job_id: "05c2b0a3-5e6c-404c-9ce1-cbab27da8319"
-          condition_type: "on_job_success"
-          target_job_id: "080e20a3-7406-4424-9753-c974a3c847d7"
+          delete: true
         }
```

opragmatico:
## Short Description

Merge and save condition type, expression and label.

## Related issue

Fixes #553 

- [x] I have added unit tests

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm changeset` from root to bump versions
- [ ] Run `pnpm install`


packages/deploy/test/fixtures.ts
```diff
@@ -75,14 +76,18 @@ export function fullExampleState() {
         edges: {
           'trigger-one->job-a': {
             id: 'ea264d6d-9767-4a2c-810f-deb10961a6dc',
-            condition: null,
+            condition_type: null,
```
taylordowns2000, line 79: i'd expect this to be `always`

taylordowns2000 (commented):
@jyeshe , looks great. two questions:

1. why is that `trigger-one->job-a` edge_type set to `null`? i don't _think_ that this is a case that could happen... if i think about the select lists in Lightning, the only options that are available for a trigger-to-job edge are `always` and a custom javascript expression.
2. the tests dont seem to be checking for any condition expressions. that's probably OK, but changing one of the `condition_type: 'always'` expectations to a custom javascript expression would make the suite feel more robust.

packages/deploy/test/fixtures.ts
jyeshe: @taylordowns2000 The pre-existing field `condition` (now renamed to `condition_type`) is a nullable column. May I apply a migration on Lightning converting all `NULL` to `'always'`? To be more conservative, we can update only where `source_trigger_id is not NULL`.

taylordowns2000 (changes requested):
@jyeshe , this looks great but i couldn't get it to pass my click test! let's use cli deploy tomorrow AM together and figure it out.

taylordowns2000 (approved):

PR #557, opened 2024-01-10
Remove props  from final state

### LINKED ISSUE #554: Runtime: accept a list of strings to remove from final state
* When the runtime returns state from a job, it should remove certain keys
* Right now we remove state.configuration
* if a different array of strings is passed, we ONLY remove those strings

The runtime by default just removes configuration. This is fine for CLI. 

The worker needs to remove configuration and response

Worker needs to accept keys form the CLI/environment so that this can be flexible

josephjclark:
This PR adds an option to the runtime to remove extraneous or sensitive properties from the state object at the end of a job.

Closes #554 and https://github.com/OpenFn/adaptors/issues/453

## Runtime Changes

The runtime now supports a `statePropsToRemove` option. This takes an array of strings. All strings will be removed from the final state at the end of a step/run. If a property is actually removed, a debug log will be printed.

By default, if `statePropsToRemove` is not set,  the `configuration` key will be removed.

The `deleteConfiguration` option - which was never really used - has been removed in favour of the `statePropsToRemove`  array.

## Engine Changes

The reason we're doing this work is that the Worker must not send `response` back to Lightning. The fix here means that modern and legacy adaptors will be made secure without any additional dev effort.

This is actually implemented in the Engine, which is easier and makes more sense in the current architecture.

## CLI

The CLI should be unaffected by this work. Jobs run through the CLI will continue to remove `configuration`, but will leave `response`

### DIFF
--- a/packages/ws-worker/src/start.ts
+++ b/packages/ws-worker/src/start.ts
@@ -19,9 +19,15 @@ type Args = {
   backoff: string;
   capacity?: number;
   runMemory?: number;
+  statePropsToRemove?: string[];
 };
 
-const { WORKER_REPO_DIR, WORKER_SECRET, MAX_RUN_MEMORY } = process.env;
+const {
+  WORKER_REPO_DIR,
+  WORKER_SECRET,
+  MAX_RUN_MEMORY,
+  STATE_PROPS_TO_REMOVE,
+} = process.env;
 
 const args = yargs(hideBin(process.argv))
   .command('server', 'Start a ws-worker server')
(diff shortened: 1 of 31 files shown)


packages/ws-worker/src/start.ts
```diff
@@ -72,6 +78,12 @@ const args = yargs(hideBin(process.argv))
     default: 5,
     type: 'number',
   })
+  .option('state-props-to-remove', {
```
josephjclark, line 81: @taylordowns2000 you can now override `statePropsToRemove` from a worker arg or env. It's quite a verbose name but I'm not sure I'd want to make it anything shorter :thinking:

josephjclark:
I removed the old 0.4.1 worker tag and build (which was never released) and added a new 0.5.0 tag and build.

We should probably rebuild the tag back on the release branch when the deploy stuff is merged

Issue #439, opened 2023-11-06
Allow backoff configuration via environment variables


taylordowns2000:
For kubernetes deployments of ws-worker, I think it would be valuable to be able to pass the backOff configuration in via the env (kinda like [WORKER_SECRET](https://github.com/OpenFn/kit/blob/main/packages/ws-worker/src/start.ts#L76)).

Right now, we can control the backoff with an arg used in the start command, so this is low priority, but I think managing this via ENV (or at least having the option to do so) would be preferable down the line.

josephjclark:
Update: all CLI args should be settable through env var and overridable through CLI.

Also, related, we should launch the worker through `node` and not `pnpm`

taylordowns2000:
ah nice. i can change the deployment quick if you've got a better starting command. should it be npm exec? here's what i've got now for the demo site.

most important things I can think of now are `max_run_duration`, `backoff`, and `log`.

```
      containers:
        - image: openfn/ws-worker:v0.2.6
          command: ["pnpm"]
          args:
            [
              "start",
              "-l",
              "ws://demo-service:443/worker",
              "--log",
              "debug",
              "--backoff",
              "0.1/2",
            ]
```

josephjclark:
@taylordowns2000 we've had some problems with running through npm. Turns out thats why the worker isn't getting killed in lighting - npm creates an orphaned process which is hard to kill.

Should be ok in the container but still, I think we're a bit scared of npm now!

The lightning URL is the other thing we need to add to the environment.

It's not a big job but quite low priority for me at the moment...

josephjclark:
Options to track:

* maxworkers
* backoff
* log level i guess
* lightning url
* global memory limit
* global run timeout
* global attempt timeout? But I really don't think the worker cares about this, it's a lightning concern

Issue #414, opened 2023-10-31
Run Timeout design/reconciliation


josephjclark:
What is a timeout?

The runtime supports a timeout, which is actually a timeout for each job. If the timeout is set at 1 minute, then each job has 1 minute to run. Not an unreasonable design, biased by the initial runtime implementation which just executed a single expression, not a whole workflow.

Lightning will want to send a timeout to the Worker for an attempt. Is this a timeout for the whole attempt, or per run?

I don't know whether we should support AttemptTimeout and JobTimeout as two distinct options, or whether we should only support an attempt timeout down in the runtime, or what.

This needs a bit of wider thinking.

## Design Overview

* Run timeout is tracked in runtime AND engine
* Attempt timeout is tracked in worker

josephjclark:
I am about to implement a "quick fix" for the worker, where the engine has a timeout for the whole workflow, and the runtime itself does not timeout.

josephjclark:
Attempt timeouts are probably only interesting to a) Lightning (to track lost attempts) and b) to kubernetes, to know how long to allow for a graceful shutdown when a pod is closed down. In other words, external services.

I think the run, which is the unit of work, is the thing you want to timeout out. If a job takes 4 minutes to complete, the next downstream job may only have a minute left to run. Why should it be affected by the earlier job being slow?

Here's what I think we need with this:

* The runtime continues to have a timeout per job/run.
* The worker should double-buffer the same timeout per run. This should be managed from the main thread (with maybe a 10% grace period so that the runtime fails first), and allows us to trap things like infinite loops, or the thread crashing silently (although I don't think that can happen)
* The worker accepts a global run-timeout flag. Attempts can override this via an option on the attempt object.
* The engine or worker has no timeout for the whole attempt/workflow

All timeouts should be set in minutes (not ms)

All timeouts should be prefixed (`ie, jobTimeout`) so that it's always clear what we mean.

This needs to be agreed on the Lightning side.

josephjclark:
I've opened this up on Lightning, which I think is a better way to handle lost (but has nothing to do with anything in kit) https://github.com/OpenFn/Lightning/issues/1410

josephjclark:
An update! So Lightning is also interested in an attempt duration for billing purposes. We may wish to kill long-running attempts (not runs, because the attempt is the unit of value and the split of runs is arbitrary) at some billing tiers.

This is a strategic timeout, rather than an error handling one.

[...]

PR #565, opened 2024-01-24
Options improvements

### LINKED ISSUE #500: Worker: Add an Attempt Timeout
The worker should have the ability to track the duration of a given attempt, and kill it if it runs too long.

We need to support a global/default timeout, as well as an option on an individual attempt.

This requires the ability to cancel an event, which overlaps #408. There's almost a mini-epic:

* engine: allow event to be cancelled
* worker: listen to `attempt:cancel`
* worker: timeout attempts

josephjclark:
This PR bundles a few things:
* Uses env vars in worker config
* Adds an `attemptTimeout` to the worker and engine. I don't think this is the place to rename it to `runTimeout` for the new terminology, so that will follow
* Adds better logging of config to the worker, so it's clearer what's happening

Targeting the fancy-engine rework (I may merge it in)

Closes #500 #439

Note that #414 is is still outstanding in the runtime and CLI. I want to create `attemptTimeout` and `runTimeout` options there so that it's always clear what you mean by timeout.

### DIFF
--- a/.changeset/angry-ladybugs-breathe.md
+++ b/.changeset/angry-ladybugs-breathe.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Added env vars for CAPACITY, BACKOFF, LOG_LEVEL, PORT
--- a/.changeset/angry-ladybugs-breathe.md
+++ b/.changeset/angry-ladybugs-breathe.md
@@ -2,4 +2,4 @@
 '@openfn/ws-worker': patch
 ---
 
-Added env vars for CAPACITY, BACKOFF, LOG_LEVEL, PORT
+Updated start env vars and arguments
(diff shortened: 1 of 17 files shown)


josephjclark:
We should use `MAX_RUN_DURATION` as attempt timeout

josephjclark:
TODO: ensure the yargs command documents the env var, so you can do `start --help`

josephjclark:
TODO: Add `WORKER_` to env vars for all env vars

Check the lightning_URL env var. Also port.

josephjclark:
TODO: changelog

josephjclark:
@taylordowns2000 Do you maybe want to eyeball this one before we merge it?

You should check `worker/src/start.ts` to see the options, and maybe run `start --help` locally to see the help text.

If you'd prefer to test on the next RC that's fine too!

PR #573, opened 2024-01-25
Worker: Log exit reason in attempt log

### LINKED ISSUE #567: Worker: Log the exit reason
It would be super help to log the exit reason to the attempt log in lightning. Something like:

```
Run completed with status: success
```

Or

```
Run completed with status: crash
Reason: Process exited with code: 1
```

josephjclark:
This PR will print the exit reason at the bottom of the log

image

This log isn't generated by the runtime or the engine - it's added as a sort of post-script by the worker itself. We can't do it anywhere else because only the worker knows what the reason is.

But it's such an important point of feedback I think it's important to include it. Even for success states.

This closes #567

### DIFF
--- a/.changeset/seven-seas-sneeze.md
+++ b/.changeset/seven-seas-sneeze.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Send the exit reason to the attempt logs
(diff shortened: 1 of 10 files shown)


josephjclark:
TODO:  changelog

PR #575, opened 2024-01-26
Debug oom errors

josephjclark:
.


josephjclark:
Couple of learnings

-  the VM OOM test seems fine. Running standalone, anyway
-  The timeout tests don't fail when running standalone

josephjclark:
Aha! I think I found the case of the occasional test fails.

In one test, we return from the test BEFORE the task has finished executing.

That means that when the next tasks starts, it is added to the queue while we wait for the last to return. I can see in the fail log that there's a deferred call.

That's enough to throw the timings out I think.

Is that  revealing some other problem in deferred tasks?

PR #576, opened 2024-01-26
Fix flaky test

josephjclark:
Caught an issue in the pool timing out deferred workers.


josephjclark:
Interesting, I may have stumbled on another shape of OOM error

image

josephjclark:
Alright, well, apart form that one freak fail its passed 4 in a row, and I'm not going to sit here and watch it all afternoon. In she goes.

PR #579, opened 2024-01-28
finish renaming

taylordowns2000:
Work in progress:

- [x] use "try" instead of "attempt" when not referring to `Attempts`
- [x] figure out the runTimeout thing
- [x] replace `attempt` with `run`

Notes:

1. [Sometimes runTimeout is the new run?](https://github.com/OpenFn/kit/blob/main/packages/ws-worker/src/types.d.ts#L78-L79)
2. [But sometimes it's the old run?](https://github.com/OpenFn/kit/blob/main/packages/engine-multi/src/engine.ts#L100-L102)
3. I still can't work out which is which. Will pick this up with fresh eyes soon.


packages/ws-worker/src/events.ts
```diff
@@ -8,7 +8,7 @@ export type ClaimPayload = { demand?: number };
 export type ClaimReply = { attempts: Array<ClaimAttempt> };
 export type ClaimAttempt = { id: string; token: string };
 
-export const GET_ATTEMPT = 'fetch:attempt';
+export const GET_RUN = 'fetch:run';
```
josephjclark, line 11: TODO: I think `get run` is the wrong semantics here - this ought to change to `GET_PLAN`. There will be a bunch of matching renames.

josephjclark (approved):

PR #581, opened 2024-01-30
Engine: Log adaptor messages

### LINKED ISSUE #406: Logger: logged objects are serialised too early
If you setup the logger to log JSON, then the output JSON obejct is flattened to a string before being emitted.

Basically the logger assumes that it's logging a json object straight to stdout. Which is true in the CLI, but not true in the worker. The worker will actually re-parse the log back into JSON, pass it down the line a bit, then re-stringify it back to Lightning.

It's all a bit of a waste of time and basically the logger shouldn't stringify the output JSON itself. Maybe we can take an option for this.

I think the reason for this, by the way, is to prevent the data logged to stdout being truncated (which I think we started to see). Using fast-safe-stringify to log the JSON gives us more robust output.

### LINKED ISSUE #570: Worker: adaptor logs don't get sent back to lightning
Spun out of https://github.com/OpenFn/kit/issues/566

josephjclark:
It turns out that the engine is not forwarding `console.log` statements inside the runtime sandbox. This means that Lightning doesn't see stuff being logged by the adaptor - like error messages, for example.

This PR overrides the top level console object inside the worker thread, and maps all calls to it to an adaptor logger. This means that anyone outside of the VM sandbox who calls console.log will trigger an "adaptor" level log line. This should be fine.

The runtime uses an `@openfn/logger` instance, which does not log to stdout but instead logs an event out of the engine (picked up by the worker and sent to lightning). The only other console.logs will come from the actual runtime (which should never console.log!) or from an imported module - ie, an adaptor.

For debugging in local dev, any calls to `console.log` will get sent to the adaptor logger and not sent to stdout. But `console.debug` will still print to stdout. It's a little weird but it sort of works.

Closes #570

# Logger Changes

This got interesting - having printed the errors being logged by the adaptor, it turns out they're not serialisable - they just log `{}` instead of an actual message. So I've had to go and fix #406 as well, and that's ended up being by far the bigger piece of work.

Basically the logger in JSON mode used to stringify its output with  `fast-safe-stringify` BEFORE emitting it. This gave us two problems:
- Error objects are not serialized properly
- The logger emits a string, which needs to be parsed back into a JSON object to be properly handled by the engine.

So this PR makes three changes:

1) If you ask the logger to emit JSON, it'll emit objects. Not strings. The good side of this is that anyone using the logger in-process gets clean objects to reason with. The downside is that those objects are not JSON safe and not serialisable
2) The logger will serialize Errors into an object before sending it to the emitter, so they're safe to log.
3) The Engine, on receipt of a log event, will stringify the message into JSON before sending it out of the thread. This is guaranteed to be safe.
4) The Worker will parse the JSON of each log message before sending it out to Lightning. This is a bit silly - we can optimise by building a string and "pre-serializing" before sending to Lightning. But I'm not doing that yet.

The result of all this is that if job or adaptor code logs an error, it'll now get serialized before being sent to Lightning.


josephjclark:
Janky screenshot showing some of the outputs now. We can still handle circular objects, functions and stuff in the logs - even though the logger itself doesn't stringify

image

josephjclark:
This branch has made debugging (which is already quite hard in the engine) even harder. Any console.log calls I add for debugging will get redirected to the adaptor-level logger and swallowed up.

I may need to add some workaround, like a logger.trace API which doesnt get handled by the adaptor logger or something along those lines.

It's just annoying when the logs get swallowed up.

josephjclark:
Better handing of console.log(error) from within job code (this took AGES)

packages/ws-worker/src/util/log-final-reason.ts
```diff
@@ -12,7 +12,7 @@ export default async (context: Context, reason: ExitReason) => {
 
   await onJobLog(context, {
     time,
-    message: [message],
+    message: JSON.stringify([message]),
```
josephjclark, line 15: The construction of log messages is really fiddly now.

Options:
1) Create a nice helper called `sendLog(level, message, name?)`, which handles the bigint and the string parsing
2) Be more lenient about the shape of message. I mean, it doesn't HAVE to be JSON - if it's a string we parse it as JSON, if its an array we leave it as JSON

I actually think we want a little of both...

packages/engine-multi/src/api/execute.ts
```diff
@@ -82,7 +84,9 @@ const execute = async (context: ExecutionContext) => {
         threadId: '-', // no thread at this point
         message: {
           level: 'debug',
-          message: [`Timeout: ${workerOptions.timeout / 1000}s`],
+          message: JSON.stringify([
```
josephjclark, line 87: Again, I should just be able to send an array with a string here. No need to make my life difficult.

packages/engine-multi/src/worker/thread/run.ts
```diff
@@ -29,9 +29,25 @@ register({
...
+      sanitize,
+      publish
+    );
+
+    // Save the debug function so that we can use it
+    const debug = console.debug;
+
+    // override console.log
+    // any console.log statements will now get treated as adaptor logs
+    console = adaptorLogger;
```
josephjclark, line 44: This is the actual fix for #570: override `console` with the adaptor logger.

PR #583, opened 2024-02-01
Better adaptor errors

### LINKED ISSUE #582: CLI: Bad error output for http
Experimenting with the CLI and this job:
```
get({ url: 'https://api.chucknorris.io/jokes/random' });
```
Gives terrible error output.

Here's the result in 5.04:
image

It is better in 6.0.0:

image

The error is expected btw - I'd just like to be able to see what it actually is.

josephjclark:
Maybe closes #582

image

How am I going to test this?

Is the stacktrace really useful in lightning? Can we control this any how?

Is it really appropriate to graft the stack trace onto the error? Maybe it's better to return the stack to lightning and let lightning decide what to do.


josephjclark:
Away with you

Issue #556, opened 2024-01-10
Adopt new terminology (part 2)


josephjclark:
We need to update the the names of things to match Lightning:

- [ ] consider when we use "jobs" and when we use "steps"
- [x] replace "attempt" with "run"

Notes: 

- I think ExecutionPlan is still OK (although that's just an internal type), and workflow is still fine.
- Should we rename `state` to `input` in the CLI? `-i` is taken. Maybe we can add an `--input` alias.

This applies to all packages in `kit` (CLI, runtime, worker etc).

For each package, we particularly need to check:
* API interfaces
* Any options objects/keys
* All log output

Also note that docs need to change - CLI challenge

josephjclark:
See https://github.com/OpenFn/Lightning/issues/1248 for the changes made in Lightning

taylordowns2000:
part 1 is done via #564

taylordowns2000:
renaming `attempts` to `runs` is done in `0.8.0`,  here we still want to consider jobs, steps, and a few other items in the CLI

Issue #409, opened 2023-10-26
Backend terminology


josephjclark:
This issue is to tidy, consolidate and update the langauge/terminology used  throughout the backend stack:

* Expression: A string of Javascript of OpenFn DSL (ie, may or may not be compiled) to be excuted. I wonder whether, where and we we should differentiate compiled from uncompiled expressions (the runtime for example does not support uncompiled expressions ((although you can write an uncompiled expression that will work))
* Job: An expression PLUS the metadata that supports it - like the adaptor, initial state, maybe some options. The CLI can run a job directly if you pass an expression and adaptor. You can compile an expression as part of a job (but you can't really compile an expression without the job metadata). I think this is going to be renamed to a STEP.
* Workflow: A DAG of job nodes connected by edges. Also known as an execution plan. From Lightning's point of view it's an Attempt (soon I think a WorkOrder). I think Workflow is a general high level catch-all - this after all is a Workflow Automation platform. ExecutionPlan and Attempt then become like views or projections or representations of a Workflow, geared toward a specific context and backed by a formal type definition. 
* Operation Factory (?) (not really here) An adaptor exports operation factories, not operations. Technically speaking. This is horrible, subtle and confusing terminiology

Affects compiler, runtime, cli, worker, engine, everything really.

TODO: more changes to follow

PR #585, opened 2024-02-02
Backend renaming (1.0 version bumps plus the lexicon)

### LINKED ISSUE #233: Strict mode is wrong

### LINKED ISSUE #483: Worker: use job names from the attempt
When the worker generates an Execution Plan from an Attempt, it uses the job id for each job.

But a job may also have a human readable name (like "get from salesforce")

We should consider:

a) using the job name instead of the id (problematic if two job names happen to be equal)
b) including a name and an id on a job.

At the moment we use a job id in various places:
* If there's an error, we write it to state as `errors[jobId]`. Human readable job ids would help here
* If there are multiple leaf results, we write them all ot the final state as `state[jobId]: { data: {} ]`
* In logging, when a job starts execution

I think we should use the job NAME, not id, but ensure they are unique and add a number to the end if they are not. Or we could throw, I suppose. The changes of identical job names, in practice, are slim. If no job id or name is provided we can generate a name based on the index, as we do now.

Actually I might be even more radical and say that a job as a _name_, not an id. Because name implies human-readable and encouraging human readable job names makes everything better.

josephjclark:
This PR bumps the runtime up to 1.0 and introduces the Lexicon, a central repo for typings and namings.

I have started to enforce stronger type checking on the tests as part of this PR, which has helped to catch any issues in the  change of formats but also resulted in a large diff!

Closes #483 #556 #409 #233

Runtime changes:
- use new-style Execution Plan format
- rename jobs to steps (mostly, where it makes sense)
- remove strict state (closes #233)

CLI changes
- remove strict state (closes #233)
- Support but deprecate old workflow data structures
- Support execution plans
- Better use of step, job and expression in code and output #483 

Engine Changes
- Just an adjustment to match new typings

Worker Changes
- Just an adjustment to match new typings


josephjclark:
Hi @mtuchi when you have a spare half an hour would you mind taking a look at the CLI on this branch?

There's some stuff about terminology in the CLI readme - probably a good idea to look at that first.

You should look closely at the log output and make sure that it aligns with your documentation. Also if workflows and steps have names, they should be used in the log output now (instead of UUIDS), although that likely affects the worker more than the CLI

packages/lexicon/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+The OpenFunction stack is built on the concepts of Workflows, Runs, Jobs and Expressions (and more). Some of these terms can be used interchangable, or used differently in certain contexts.
+
+Here are the key concepts
+
+- An **Expression** is a string of Javascript (or Javascript-like code) written to be run in the CLI or Lightning.
+- A **Job** is an expression plus some metadata required to run it - typically an adaptor and credentials.
+  The terms Job and Expression are often used interchangeably.
+- A **Workflow** is a series of steps to be executed in sequence. Steps are usually Jobs (and so job and step are often used
+  interchangeably), but can be Triggers.
```
taylordowns2000, line 15: i think we don't technical refer to triggers as steps. right now, all available "steps" are (javascript-based) jobs. in the future, we might introduce a "notification" step which behaves differently. or even a "loop over array" step which does something else entirely. but for now, all steps are jobs.

packages/lexicon/core.d.ts
```diff
@@ -0,0 +1,137 @@
...
+export type ConditionalStepEdge = {
+  condition?: string; // Javascript expression (function body, not function)
+  label?: string;
+  disabled?: boolean;
+};
+
+/**
+ * A no-op type of Step
+ */
+export interface Trigger extends Step {}
```
taylordowns2000, line 115: might be removable, unless this is how it works in worker. (happy to consider triggers as steps if that's helpful)

packages/lexicon/lightning.d.ts
```diff
@@ -0,0 +1,161 @@
...
+/**
+ * Type definitions for Lightning and Worker interfaces
+ *
+ * This is the lightning-worker contract
+ *
+ * It is helpful to have these in the lexicon to avoid a circular dependency between lightning and the worker
+ * It's also kinda nice that the contract isn't in the worker itself, it's on neutral ground
+ */
+// An run object returned by Lightning
+export type Run = {
```
taylordowns2000, line 14: im a little surprised that this isn't called `Plan`, but i'm reading on to see what comes next!

packages/lexicon/lightning.d.ts
```diff
@@ -0,0 +1,161 @@
...
+export type STEP_COMPLETE = 'step:complete';
+
+export type ExitReason = {
+  reason: ExitReasonStrings;
+  error_message: string | null;
+  error_type: string | null;
+};
+
+export type GetPlanPayload = void; // no payload
+export type GetPlanReply = Run;
```
taylordowns2000, line 112: right, i remember being a little confused by this. when we "getPlan" i'd expect a plan to be returned... but a "Run" is returned. It's not technically a `Run` in lightning's world. To lightning, it's a `RunPlusSomeOtherThings` - is that not exactly what we want to call a `Plan`?

taylordowns2000 (commented):
i've reviewed the lexicon section and added comments inline. nothing is a deal-breaker for me, so happy for this to be the 1.0 and adjust if and when necessary. thanks for this work!

packages/lexicon/core.d.ts
josephjclark: It's useful in the worker for sure - we already recognise "execuable jobs" and "non-executable jobs" (which are triggers). Formalising it like this feels like a nice step to me, and it doesn't block any future step-types from being added

packages/lexicon/lightning.d.ts
josephjclark: Well - yes, you're right! I'll rename it to `LightningPlan` though to help distinguish Lighting's `RunPlusSomeOtherThings` structure from the runtime's `ExecutionPlan` structure.

But in the future Lightning will be sending the `ExecutionPlan` directly (I hope!)

stuartc (approved):
Looks good to me, obviously I didn't read everything - but yeah kudos on the lexicon stuff. 

Please merge it when you'd like!

josephjclark:
Thanks @stuartc !

PR #592, opened 2024-02-08
Release: 1.0 (worker, engine, cli, runtime)

### LINKED ISSUE #259: CLI 1.0.0
I've been thinking about this for a while.

When (and how) does CLI get to a 1.0.0 release?

## Why it matters

Well first of all it doesn't.

But a 1.0.0 release number is, to me, a strong vote of confidence in the stability and featureset of a package. Installing a 1.0.0 feels a lot better than installing a 0.0.56.

A 1.0.0 release can also imply that the package is in production.

I haven't been very good at versioning the CLI so far, I just patch bump everything. But really major features like magic, docs, workflows, should each have been a minor bump.

I did intend to publish a symbolic 0.1.0 release when we integrated with lightning. But I, er, didn't.

## Features

The CLI now supports:

* Running jobs standalone
* Running workflows
* Autoinstalling adaptors
* Running adaptors straight out of the monorepo
* Printing documentation for adaptors
* Generating magic metadata
* Highly configurable logging

I've also removed the strict mode stuff which was proving problematic

The features that are outstanding are a bit hand-wavy and shouldn't block 1.0:

* Credential management / Lightning integration
* Generate a workflow.json
* Typescript jobs (is this even a CLI issue?)
* a `config` command to describe and debug a config object #254

## Possible roadblocks

So what's stopping us bumping up to 1.0?

There are a couple of things which would change the interface or majorly affect stability. We should consider these:

* I'd like to change the `--log` argument (#190)
* Workflow plan.job.data isn't quite right #258
* Major issues in `describe-package` which mean that we still  don't always generate the right imports while compiling #238
* Maybe state globals are still a bit of an open question (I don't think state should ever be globally available and only accessible from a callback) #17 . But whatever we do I don't think we'll be strict about it (to support legacy jobs)

josephjclark:
The 1.0 release across the stack includes a number of polishes, tidyups, refactorings and renamings.

* [x] Major backend renaming #585
* [x] Better log output #585
* [x] Central lexicon for nice clear type defintions #585
* [x] Remove strict mode #585
* [x] Autoinstall by default
* [x] Send  version number to worker
* [x] Support zero persistence
* [x] Verify attempt token from lightning
* [x] Better error handling on fetch:attempt
* [x] Tidy version output 

This release uses a new data structure for wofklows in the CLI. Old structures are still supported, but you'll see log output like this:

image

Closes #259


mtuchi:
**Testing checklist** 
- [x] New workflow structure `{
  workflow: { steps: [] }
  options: { start }
}`
- [x] Adaptor auto-install without specifying `-i`
- [x] Test workflow with the old structure `jobs: [{}]`
- [x] Show `nodejs` and `cli` version only

josephjclark:
@mtuchi thank you!

Autoinstall still exists! It's just defaulted now. I'll double check the docs but I think this is OK.

I don't think I expect any changes to the CLI logs. Are the error messages you get OK or do they need addressing?

mtuchi:
@josephjclark the error message are OK 👍🏽

PR #593, opened 2024-02-08
Send worker versions

### LINKED ISSUE #563: worker: send version number when connecting to ws endpoint
### Details

In order to provide more details about how a step is being run and to give us room to accommodate API changes we should send the version number to Lightning.

### Implementation notes

I think the easiest way (at least from how to get it to Lightning) is to use query parameters. So adding `v: <VERSION>` to the params (`token` is already there).

josephjclark:
_Temporarily targeted at `lexicon` as that's where I branched from._

This PR encodes `api_version` and `worker_version` into the URL to connect to the worker queue.

API version represents the API spec/contract supported by this worker version. That uses a `major.minor` numbering scheme, where `major` is the actual spec version (`v1`), and `minor` is the implementation version (also `v1`).

The worker_version is the npm version of the actual worker.

Although I think `api_version` is the more valuable feature, it is cheap and easy to send both ¯\_(ツ)_/¯

Closes #563

### DIFF
--- a/packages/lexicon/index.d.ts
+++ b/packages/lexicon/index.d.ts
@@ -1,1 +1,2 @@
 export * from './core';
+export * as lightning from './lighting';
--- a/packages/lexicon/index.js
+++ b/packages/lexicon/index.js
@@ -1,0 +1,1 @@
+export * as lightning from './lightning';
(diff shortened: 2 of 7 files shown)


stuartc (approved):
Awesome!

PR #598, opened 2024-02-13
Verify run token

### LINKED ISSUE #399: ws-worker: verify the attempt token we get back from lightning
When claiming an Attempt, we get a token back from lightning. This JWT needs to be sent back to lightning to join the attempt channel and start processing the attempt.

At the moment we do no validation on the token we get back from lightning. Really we should verify it against the public key to make sure it's legit.

josephjclark:
The PR verifies the run token (JWT) that Lightning sends with each claim.

Verification means checking the signature against the public key and asserting the `iss` (and `id` TODO) claims.

Support has been added to the lightning mock to generate a real JWT token, which is then used in testing.

I don't have very good explicit testing around this - but some tests do now rely on the tokens being set properly. Integration tests now use tokens, and there is a test to ensure the server is destroyed after a token is rejected. So that's not bad.

Closes #399

### DIFF
--- a/packages/ws-worker/src/api/claim.ts
+++ b/packages/ws-worker/src/api/claim.ts
@@ -14,7 +14,6 @@ const verifyToken = async (token: string, publicKey: string) => {
 
   const { payload } = await jose.jwtVerify(token, key, {
     issuer: 'Lightning',
-    // TODO id
   });
 
   if (payload) {
--- a/packages/ws-worker/src/api/claim.ts
+++ b/packages/ws-worker/src/api/claim.ts
@@ -1,19 +1,13 @@
+import crypto from 'node:crypto';
 import * as jose from 'jose';
 import { Logger, createMockLogger } from '@openfn/logger';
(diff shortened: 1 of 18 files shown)


packages/ws-worker/src/api/claim.ts
```diff
@@ -1,14 +1,38 @@
...
+
 import { CLAIM } from '../events';
 
 import type { ServerApp } from '../server';
 
 const mockLogger = createMockLogger();
 
-// TODO: this needs standalone unit tests now that it's bene moved
-const claim = (app: ServerApp, logger: Logger = mockLogger, maxWorkers = 5) => {
+const verifyToken = async (token: string, publicKey: string) => {
```
josephjclark, line 12: Hi @stuartc - this is where we verify the run token based on the lightning public key. The key comes from `WORKER_LIGHTNING_PUBLIC_KEY` as a base64 encoded string - you can trace this through from `src/start.ts` if you wish

stuartc (approved):

PR #601, opened 2024-02-14
Worker: Better error handling for credential errors

### LINKED ISSUE #600: Worker: no error handling on fetch:credential
Just like #590, the Worker doesn't have very good error handling around `fetch:credential`. If for some reason a credential can't be returned (like with this Oath bug we're seeing), only bad things will happen.

At the time of writing that's manifesting as the Run being marked as Lost.

I am a bit concerned that the worker will get the `run:error` event before the `run:start` event. I think this is legit - we can't even start the run if we can't load the credentials.

(later, credentials and dataclips will be loaded on demand, rather than preloaded, so that becomes kinda moot)

This is all closely related to the dataclip stuff - see #595

josephjclark:
Closes #600 

This is actually a fix down in the engine.

When loading credentials from Lighting, if any credentials fail to load, we  catch and report the error, and throw in the main validation function.

Previously, because of how credentials are loaded in nested promises, this case resulted in an uncaught exception and everything would blow up. It's more stable now - and we should get a nice error message to explain the fail reason!

### DIFF
--- a/.changeset/twelve-ravens-listen.md
+++ b/.changeset/twelve-ravens-listen.md
@@ -1,0 +1,5 @@
+---
+'@openfn/engine-multi': patch
+---
+
+Add a CredentialLoadError
--- a/integration-tests/worker/test/integration.test.ts
+++ b/integration-tests/worker/test/integration.test.ts
@@ -304,6 +304,35 @@ test.skip('run a job with credentials', (t) => {
   });
 });
 
+test('run a job with bad credentials', (t) => {
+  return new Promise<void>(async (done) => {
+    const attempt = {
+      id: crypto.randomUUID(),
+      dataclip_id: 's1',
+      jobs: [
+        {
+          adaptor: '@openfn/language-common@latest',
+          body: 'fn((s) => s)',
+          credential: 'zzz',
+        },
+      ],
+    };
+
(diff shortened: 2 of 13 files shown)


josephjclark:
Hey @taylordowns2000 are you able to retest against this branch?

josephjclark:
I am unable to reproduce the original issue on my lightning.

The changes here are well worth merging IMO.

As a result of this branch, Taylor sees these logs (edited for clarity):
```

[RTE] ❯ Loading credential a671b76f-8173-4564-b354-64695aa60240 for step 4415c14f-bb32-4407-a38a-cacf1a7fc49a
[SRV] ✘ error connecting to run:9f9c471a-79f1-4627-bd2f-17b9183ff2c1 {
  "reason": "unauthorized"
}
[RTE] ❯ Error loading credential a671b76f-8173-4564-b354-64695aa60240
[SRV] ✘ ERROR in workflow-error handler: timeout
```

I think that when the credential fails to load, the worker gets kicked out of the websocket at the lightning end. It is no longer allowed into the run channel.

When the credential fails to load, the worker will try to send logs and events to Lightning. But these fail because the channel is dead - and so we get the timeout on the `workflow-error` handler (which really means `run-error`, which has been fixed on this branch)

Issue #606, opened 2024-02-20
CLI Codegen design


josephjclark:
This issue contains a living design for `openfn codegen` (or whatever we call it), the CLI command which today will generate an adaptor template, but tomorrow may generate other stuff.

## Inputs

The command will need to take the following inputs:

* openAPI Spec
* The endpoints you want to generate mappings for
* The folder to output the various generated assets to (actually is this just the monorepo root and adaptor name?)
* (optional) the model you want to use (name? url?)
* (optional) an API key for the model

## Outputs

The existing AI service will generate the following files:

* adaptor.d.ts
* adaptor.js
* adaptor.test.js

We need to extend it to generate a full adaptor template:

* src/adaptor.d.ts
* src/adaptor.js
* src/index.js
* test/adaptor.test.js
* readme.md
* package.json

What if you point to an existing adaptor? Can we merge the new code with the existing `Adaptor.js`?

## API Spec

Here's what the API might look like:

```
openfn generate adaptor spec.json
```

`spec.json` would contain all the inputs. Presumably they can all be overridden by CLI Args too.

Right now this is:
```
spec: <openAPI spec as JSON>,
instruction: "Create an OpenFn function that allows me to create new trackedEntityInstances via the /trackedEntityInstances endpoint"
model?: "gtp-4"
```

But I would probably like to change "instruction" to "endpoints", and just have a list

## Changing vs Generating

It's easy to generate a fresh adaptor because we're outputting fresh source code. No toes to step on.

But if this is going to work on an existing adaptor, we have to generate code inside existing code.

Either we do that by injecting generated code into an existing file, perhaps using static analysis tools, or we get the AI to do it (which feels like a risk to me).

I suppose we start by saying: here is the existing adaptor.js file. Please generate or amend this function. If that proves reliable then great, we can keep it. Bear in mind that everything is backed by git, so it's not like the codegen can make permanent breaking changes - also changes are easy to see via diff.

## Model Abstraction

Should we expose the model to the user?

I think we're building a highly specific service here. I think we should chose a model that works well for the task, maybe fine tune it (I think we had mixed results on that?) and optimise the prompt for that model.

Maybe we even find that we want to use multiple models in the pipeline.

So I kind of thing we should abstract the model, maybe expose it as an option if it helps, but otherwise don't put it it in the user's hands. This also implies that we own the model's API keys (if required). Maybe users need an API key to use the service generally?

## Python Server

The current AI spike is implemented as a python service behind a python http server.

[...]

PR #609, opened 2024-02-21
CLI: Adaptor codegen

josephjclark:
This PR adds a `generate` command to the CLI, which calls out to the AI codegen services in `gen`

See details in #606

At the moment this is a simplified form. It doesn't generate a full adaptor, it just generates adaptor.js with the generated code.

There are a whole host of backend changes required in `gen`, including:
* A single http server which we can deploy
* Generate code inside an existing adaptor
* Generate for multiple endpoints

Still to do in this PR:

* [ ] Generate from a list of endpoints in the spec (rather than passing an instruction)
* [ ] Validate the input spec
* [ ] Unit tests
* [ ] The signature `openfn generate adaptor dhis2 spec=./dhis2.json --endpoint=getTrackedEntityInstances` is not particularly friendly. I suppose you can encode the whole request into json `openfn generate adaptor ./spec.json`, but really you only want to fire it once. At least, if we get it to work.
* [ ] Add some kind of generic `openfn data <url>` command, or something, which basically posts a payload to an arbitrary endpoint on the python server. This is an extension hook which allows someone to easily add a new endpoint to the Python AI service and call it from the CLI without any development. The `data` command would load some JSON input, post it to the server (perhaps with an API key), and output the result somewhere on disk.
*  [ ] Create a template generator script in the adaptors monorepo. When generating code against the monorepo, this code  generator should run first, to ensure there's a decent stub adaptor, and then inject adaptor.js and adaptor.d.ts into it. This template generator is useful even the code generator here is no adopted (and will be the death of the confusing `template` folder). See https://github.com/OpenFn/adaptors/issues/19

Maybe to do later:

* Fetch openfnAPI schemas from a url (is that a thing? Do people host their schemas?(


josephjclark:
I'm gonna pause on this until we know what we want to do about the server  (continue to develop in python or shift to js?).

josephjclark:
Closing in favour of generic `apollo` command #682.

I might come back to this one day for a more user friendly adaptor generator

PR #610, opened 2024-02-22
Allow a workflow to support multiple adaptor versions

### LINKED ISSUE #268: Runtime: workflows can't handle two versions of the same adaptor
A workflow can technically specifiy two versions of the same adaptor:
```
{
jobs: [{ adaptor: 'common@1.0.0' }, { adaptor: 'common@2.0.0' }]
}
```
The runtime ought to be able to support this but it's a little tricky.

We can install and maintain two versions of the same adaptor in the repo, that itself isn't a problem.

It's the linker that gets a bit tricky.

The linker takes a map of imports to paths, like `{ '@openfn/language-common': 'repo/node_modules/@openfn/language-common_1.7.7` }`

And when it sees an import statement from a compiled job, it'll look up the package and load from that file.

But the import statement isn't versioned.

This all comes from the old pre-workflow idea of we run a single job and map the imports to particular paths. It breaks down in the workflow model.

So we either:
* Somehow update the linker's paths in between jobs in a workflow, so that versions match
* Introduce some kind of declaration of the version at the top of the file `// ADAPTOR @openfn/language-common 1.0.0`, which the runtime/linker can use (I want to do this ANYWAY because it makes the code more portable and declarative)

josephjclark:
Closes #268

Bit tricky this one! But I think I've got it.

Every step in a workflow gets compiled by the runtime manager (the CLI or the Engine) and is given an import statement `import { fn } from '@openfn/language-common'`.

It's up to the linker, deep inside the runtime, to decide how how resolve this import. Should we look in the repo? Were we given an explicit mapping? What version do we want?

What usually happens is that the runtime manager (the CLI or worker) will pre-parse the workflow and make a bunch of decisions about these imports. In the CLI, for example, the user might add an explicit path to `language-common`. Or if someone asked for `@latest`, the CLI will maybe autoinstall and replace `@latest` with an actual version (I think).

The problem is that one list of adaptor-version mappings is compiled for the whole workflow, and that one list is shared by the linker. So if we decide that `language-common` is 1.0, all steps will load 1.0 regardless of what was asked for.

What _this_ PR does is calculate this module map for each step inside the runtime based on the original input, and passes this "localised" map into the linker for each step.

It seems to work.

Sadly I can't get unit tests very close to the code without making major changes. It's actually very small tweak in a glue layer and I can't intercept in the right place.

## TODO

* [x] Add some integration tests around this
* [x] Ensure autoinstall works with multiple versions
* [x] Maybe move the adaptor version reporting so that it better reflects what is actually loaded

### DIFF
--- a/packages/engine-multi/src/api/execute.ts
+++ b/packages/engine-multi/src/api/execute.ts
@@ -50,7 +50,6 @@ const execute = async (context: ExecutionContext) => {
     const runOptions = {
       statePropsToRemove: options.statePropsToRemove,
       whitelist,
-      repoDir: options.repoDir,
     } as RunOptions;
 
     const workerOptions = {
@@ -108,6 +107,7 @@ const execute = async (context: ExecutionContext) => {
         jobError(context, evt);
       },
       [workerEvents.LOG]: (evt: workerEvents.LogEvent) => {
+        // console.log(evt.log.name, evt.log.message);
         log(context, evt);
       },
       // TODO this is also untested
--- a/packages/engine-multi/src/api/execute.ts
+++ b/packages/engine-multi/src/api/execute.ts
@@ -15,14 +15,13 @@ import {
 } from './lifecycle';
 import preloadCredentials from './preload-credentials';
 import { ExecutionError } from '../errors';
+import type { RunOptions } from '../worker/thread/run';
 
 const execute = async (context: ExecutionContext) => {
   const { state, callWorker, logger, options } = context;
   try {
-    // TODO catch and "throw" nice clean autoinstall errors
-    const adaptorPaths = await autoinstall(context);
+    await autoinstall(context);
 
-    // TODO catch and "throw" nice clean compile errors
     try {
       await compile(context);
     } catch (e: any) {
@@ -49,10 +48,10 @@ const execute = async (context: ExecutionContext) => {
     const whitelist = options.whitelist?.map((w) => w.toString());
 
     const runOptions = {
-      adaptorPaths,
-      whitelist,
       statePropsToRemove: options.statePropsToRemove,
-    };
+      whitelist,
+      repoDir: options.repoDir,
+    } as RunOptions;
 
     const workerOptions = {
       memoryLimitMb: options.memoryLimitMb,
@@ -116,6 +115,7 @@ const execute = async (context: ExecutionContext) => {
         error(context, { workflowId: state.plan.id, error: evt.error });
       },
     };
+
     return callWorker(
       'run',
       [state.plan, state.input || {}, runOptions || {}],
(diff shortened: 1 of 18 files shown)


josephjclark:
Likely issue: I think the module-overiding thing is removing path information. So if you specify an adaptor path in the linker options, it'll get blatted out and ignored. This broke the engine and will be a problem in some CLI edge cases.

I've also preferred repoDir to adaptorPaths in the engine, but I wonder if this is really the right thing to do. It might be a bit slower.

packages/engine-multi/src/api/execute.ts
```diff
@@ -49,10 +48,10 @@ const execute = async (context: ExecutionContext) => {
     const whitelist = options.whitelist?.map((w) => w.toString());
 
     const runOptions = {
-      adaptorPaths,
```
josephjclark, line 52: I am still not sure that replacing adaptorPaths with repoDir is the right answer.

First off, adaptorPaths will run more quickly. If we pass in the repoDir then the runtime has to do a lot more work to load the correct module (including parsing the package json).

Secondly, this fix is kind of a workaround a bug in the runtime, where paths get blatted.

I think a better approach here would be for autoinstall to write tothe plan a path to the correct, versioned adatpor. So the plan changes. The runtime receives neither adaptorpaths nor repodir, but absolute paths in the plan

packages/runtime/src/execute/expression.ts
```diff
@@ -125,11 +130,15 @@ export const wrapOperation = (
...
-  opts: Options = {}
+  opts: Options = {},
+  moduleOverrides: ModuleInfoMap = {}
 ): Promise<JobModule> => {
   if (typeof expression === 'string') {
     const exports = await loadModule(expression, {
       ...opts.linker,
+      // allow module paths and versions to be overriden from the defaults
+      // TODO I think this is too harsh and path information will be lost
+      modules: Object.assign({}, opts.linker?.modules, moduleOverrides),
```
josephjclark, line 141: This isn't a good solution because it totally blats the base modules object.

Ok, maybe it's academic because of how the CLI and Engine work. But really this should be a merge rather than a blat.

The problem is that if the linker options includes a path for a module, the path is lost if the override contains a version. It's silly really.

just needs a better merge algorithm. The hardest bit about that is that this is hard code to test right now...

packages/runtime/src/execute/compile-plan.ts
```diff
@@ -127,6 +128,12 @@ export default (plan: ExecutionPlan) => {
       'name',
     ]);
 
+    if ((step as Job).adaptor) {
```
josephjclark, line 131: Allow the incoming job to include a linker object

PR #643, opened 2024-03-28
Fix lazy state

josephjclark:
The lazy state operator `$` is causing trouble if you try to use `$` as a variable, a la:

```
parseXML(state.data, $ => {
    const inner = $.load(response.text()); // error
 });
```

This PR adds a bit of intelligence to the is-this-a-lazy-state-reference heuristic to ignore $ variable declared in the source. See the unit tests for details.

### DIFF
--- a/integration-tests/worker/CHANGELOG.md
+++ b/integration-tests/worker/CHANGELOG.md
@@ -1,5 +1,13 @@
 # @openfn/integration-tests-worker
 
+## 1.0.38
+
+### Patch Changes
+
+- @openfn/engine-multi@1.1.3
+- @openfn/lightning-mock@2.0.3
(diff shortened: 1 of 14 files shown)


josephjclark:
@mtuchi assuming this passes CI tests and it works for you, you can merge it and it'll be released! You'll have to ask Taylor to bump the lightning worker

mtuchi:
@josephjclark let's hold this, i would love to test before merging.

mtuchi:
Hiya @josephjclark i have tested this, it works as expected. Thanks for the fix 👍🏽

mtuchi (approved):
Tested locally on water-aid project, it works as expected 👍🏽

Issue #363, opened 2023-08-21
CLI: ReferenceError: dateFns is not defined


mtuchi:
### User story
When using `http` adaptor, The jobs that are using `common/src/dateFns.js` or `common/src/http.js` fails with reference-error:
```
[R/T] ✘ Failed job fetch-gc-submissions after 2.664s
[R/T] ✘ ReferenceError: dateFns is not defined
[R/T] ✘ Check state.errors.fetch-gc-submissions for details.
```
The version of CLI kit that i am using 
```
[CLI] ℹ Versions:
         ▸ node.js     18.12.1
         ▸ cli         0.2.4
         ▸ runtime     0.0.28
         ▸ compiler    0.0.34
```

josephjclark:
Adaptor docs is actually a bit broken too, as it lists the http functions of common at the root.

![image](https://github.com/OpenFn/kit/assets/7052509/2a671fc2-7240-4043-b97c-0dd769c0d6f9)

I sort of remember doing that - unsure if I misunderstood the structure of the package or if it was a deliberate workaround

PR #652, opened 2024-04-09
Fix alias exports

### LINKED ISSUE #238: describe-package: re-exported packages are not tracked
This is a new flavour of a familiar problem.

`describe-package` doesn't seem to be tracking the exports of `http` and `axios` out of common. The result is that the compiler won't try to import `http` from common.

So this:
```
fn(http.get('https://api.chucknorris.io/jokes/random'))
```
Compiles as this:
```
import { fn } from "@openfn/language-http";
export * from "@openfn/language-http";

export default [
  fn(http.get('https://api.chucknorris.io/jokes/random')),
];
```
And doesn't execute because `http` is not defined.

This is something to do with typescript not tracking the types of the re-exported `http` and `axios` packages, and rather than exporting an unknown name, it seems to just ignore the thing entirely. This has caused us problems before but we've worked around it by parsing the typedefs for `language-common`. That strategy seems to break down here.

Really as a quick fix I want TypeScript to report the names (but not types) of all exports. A longer fix would be to ensure deep type tracking of `common`'s exports.

## Workaround

As a workaround you should be able to add your own import statement to the top of your job
```
import { fn, http } from "@openfn/language-http";
```

## Prerequisites

Just to flag that we've got #197  outstanding - basically `describe-package` is in serious need of an overhaul. Ideally I'd take pn both of these together to provide a robust solution. But I may be able to crowbar something in without touching #197.

josephjclark:
This PR adds a little fix to `describe-package` which allows us to track `export * as http from './http'` style export aliases.

This has been a big problem and stops us using eg `http.get` in jobs.

The problem is that our `wrapped-symbol.ts`, which wraps up a TS symbol into something more manageable, is failing to recognise exports of the type  `export * as `.  Two small tweaks to `wrapped-symbol` fix it.

One thing to note is that the generated package description now includes a `namespaces` array as well as a `functions` array. The nice thing about this is that the new exports won't break anything unexpected - we're adding new information to an export, we're not changing existing information,

This fix will apply to the CLI and worker.

## Related issue

Fixes #238 #363


josephjclark:
This job now works:
```
fn(async (state) => {
  const response = await http.get({ url: 'https://jsonplaceholder.typicode.com/todos/1' })(state);
  return {
    ...state,
    data: response.data
  }
})
```
Output:
```
[CLI] ♦ Versions:
         ▸ cli                        1.1.4
         ▸ @openfn/language-common    latest
[CLI] ✔ Installing packages...
[CLI] ✔ Installation complete in 336ms
The common.http.¬get function has been deprecated. This adaptor should migrate to use common.util.http instead.
[R/T] ✔ Completed step job-1 in 231ms
[CLI] ✔ Result: 
[CLI] ♦ {
  "data": {
    "userId": 1,
    "id": 1,
    "title": "delectus aut autem",
    "completed": false
  }
}
[CLI] ✔ Finished in 650ms ✨
```


The same job against the production CLI fails (http not defined)

josephjclark:
This works with the new common helpers (but only if I edit the adaptor to support the helpers, see https://github.com/OpenFn/adaptors/issues/507 )

```
fn(async (state) => {
  const response = await util.request('GET', 'https://jsonplaceholder.typicode.com/todos/1');
  return {
    data: response.body
  }
})
```

Issue #638, opened 2024-03-21
compiler: maybe make AST visitors run in sequence


josephjclark:
At the moment, the compiler will walk the job AST once, and call out to each transformer function as it walks the tree.

This sort of means that all the visitors run at the same time. There's no sequencing.

That's been fine up until now. But the experiment lazy-state transformer give us a problem. Technically, it needs to replace the `$` references before import statements are generated -  otherwise the `$` (which is an unused variable) will be imported from the adaptor.

Solutions:
* Run each transformer once, in order, in its own parse of the tree, and pass the tree downstream. This means parsing the AST like 4 times (becasue we have 4 transformers right now)
* Add an `order` to the transformer, and for each order key, build a single visitor. This introduces multiple passes of the  AST, but not one per transformer

I prefer that second approach - it's a better compromise.

PR #658, opened 2024-04-10
Lazy state expressions

### LINKED ISSUE #649: Lazy State Expressions
The new lazy state `$` operator has an issue with expressions.

It works great for simple and nested state references:

```
post($.data.url, {
    data: $.data.nam`
}
```

But it doesn't work for any of these expressions:

```
// tempate litrals
post('www', {
    data: `hello ${$.data.name}`
})

// string concatenation
post('www', {
    name: $.data.first_name + " " + $.data.last_name
})

// dynamic property references
post('www', {
    city: mapCityForLocation[$.data.location]
})
```

Basically it works great if you use it as a property value, but it'll fail if you try to build it into any kind of more complex expression

josephjclark:
## Short Description

Support expressions in lazy state references.

What this really means is that instead of just dumbly expanding `$.data.x` to `(state) => state.data.x`, the compiler will hoist the wrapping arrow (the "open function" reference) higher up the AST. This allows a richer range of expressions to use the `$` symbol.

As a result of this work, I am convinced that this lazy state compilation idea scales robustly and adds a lot of value to users.

Docs PR (to be merged post-release): https://github.com/OpenFn/docs/pull/469

## Related issue

Fixes #649 #638

## Supported Cases

Here are some examples we can handle thanks to this PR:

```
// template literals
get(`hello ${$.data}`)  ->  get(state => `hello ${state.data}`)
get(`hello ${$.firstname} ${$.lastname}`) -> get(state => `hello ${state.firstname} ${state.lastname}`)

// binary expressions, like concatenation
get($.firstname + " " + $.lastname) -> get(state => state.firstname + " " + state.lastname)

// Lookups
get(city[$.location]) -> get(state => city[state.location])
```
## Implementation Details

There are two changes here:

1) The compiler will walk the AST once per transformer
2) The compiler will "hoist" the wrapping arrow function for a lazy state expression up to the top of the argument, allowing a much richer range of expressions to be handled

The first issue means that previously, we built a fancy ast visitor which visits each node exactly once, and applies multiple transformations to it.

This is actually problematic because a) it's just over-complicated; b) a visitor can decide to stop walking the tree ; and c) we can't control the order of visitors (so a hack was needed).

This PR sorts all that out.

Secondly, we use a bunch of rules now to decide where to hoist the wrapping arrow to:
- walk up the tree to find the parenting operation call (ie, the`get()`
- ensure the correct argument is wrapped in an arrow function (ie, make sure we have `get(state => /* do something */)`
- replace `$` with `state`

If for any reason we cannot safely generate an arrow function around the `$` ref, we throw, causing a syntax error  basically. Like maybe the top argument is already wrapped in an arrow like `(s) => {}`, which means our generated `state` references won't work.

## Changesets

Quick note on changesets!

This is the third compiler patch for $ operators. I'm tempted to call them stable now so I've also added minor version bumps to the CLI and Worker to acknowledge this syntactic improvement to job writing.

- [x] I have added unit tests

### DIFF
--- a/.changeset/afraid-seals-share.md
+++ b/.changeset/afraid-seals-share.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': minor
+---
+
+Initial release of "lazy state" operators ($)
--- a/.changeset/five-ligers-suffer.md
+++ b/.changeset/five-ligers-suffer.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': minor
+---
+
+Initial release of "lazy state" operators ($)
--- a/.changeset/nasty-frogs-join.md
+++ b/.changeset/nasty-frogs-join.md
@@ -1,0 +1,5 @@
+---
+'@openfn/compiler': patch
+---
+
+Allow lazy state functions to be hoisted further up the tree (lazy state expressions)
--- a/integration-tests/cli/test/execute-workflow.test.ts
+++ b/integration-tests/cli/test/execute-workflow.test.ts
@@ -148,7 +148,7 @@ test.serial(
   }
 );
 
-test.serial.only(
+test.serial(
   `openfn ${jobsPath}/wf-errors.json -iS "{ \\"data\\": { \\"number\\": 32 } }"`,
   async (t) => {
     const { err } = await run(t.title);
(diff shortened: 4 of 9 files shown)


josephjclark:
I am ready to release this. Maybe @mtuchi and @taylordowns2000 would like a quick look?

I expect some other heavy features to go out into the next release (likely tomorrow morning)

PR #666, opened 2024-04-16
Fixing Issue:#192 Allow auto-installing metadata for CLI

### LINKED ISSUE #192: CLI metadata: allow autoinstall
When you run the CLI metadata command:
```
openfn metadata -a salesforce -s tmp/state.json
```
The command will fail if the adaptor has not been previously  installed to the CLI's repo.

Adaptors can be autoinstalled in the CLI either by running a job/workflow, or by running `openfn repo install <adaptor>`

The work in this issue is to enable the metadataservice to autoinstall an adaptor. 

## Background

Metadata is used to drive "magic functions" in Lightning.

The idea is that each adaptor can provide a `metadata()` export which is responsible for loading a JSON-based description of the backend entities. In Salesforce for example that means logging into Salesforce and getting a description of all the sObjects on that deployment.

The metadata command is responsible for calling that metadata function with a credential and saving the result to disk.

At the time of writing, we only provide metadata for Salesforce and DHIS2.

## Implementation Notes

The `execute` service already has an autoinstall capability. I think you basically just need to import and call the autoinstall function from within the metadata service. 

Check out the implementation in `packages/cli/src/execute/handler.ts` for reference.

## Testing

Any solution must include unit tests. Unit tests should not run `npm install` or  fetch any external files, so you may have to work out a means of mocking out the autoinstall function (right now its imported from the runtime package, but perhaps you could pass a mock override - see the pattern in worker autoinstall at `packages/engine-multi/test/api/autoinstall.test.ts`)

We have a suite of metadata tests already in `integration-tests/cli/test/metadata.test.ts`. These use a local , pre-installed test adaptor to test the functionality.

I do not want to trigger an actual metadata run against a third party service in our unit tests. It would be acceptable to run autoinstall on the common adaptor. This should fail to generate metadata (because common has no metadata function), but it should do the autoinstall. So you could test that correct error is issued by the metadata command (something like `no metadata helper found`, as reported in `packages/cli/src/metadata/handler.ts`)

SatyamMattoo:
## Short Description

The code changes allow auto-installing metadata for CLI using flags in commands.

## Related issue

Fixes #192

- [ ] I have added unit tests

### DIFF
--- a/integration-tests/worker/package.json
+++ b/integration-tests/worker/package.json
@@ -32,8 +32,5 @@
   "files": [
     "dist",
     "README.md"
-  ],
-  "devDependencies": {
-    "@types/koa": "^2.15.0"
-  }
+  ]
 }
--- a/integration-tests/worker/package.json
+++ b/integration-tests/worker/package.json
@@ -32,5 +32,8 @@
   "files": [
     "dist",
     "README.md"
-  ]
+  ],
+  "devDependencies": {
+    "@types/koa": "^2.15.0"
+  }
 }
--- a/packages/cli/src/metadata/command.ts
+++ b/packages/cli/src/metadata/command.ts
@@ -3,7 +3,7 @@ import { build, ensure } from '../util/command-builders';
 import type { Opts } from '../options';
 import * as o from '../options';
 
-export type MetadataOpts = Required<Pick<Opts, 'adaptors' | 'repoDir'>> &
+export type MetadataOpts = Required<Pick<Opts, 'adaptors' | 'repoDir' | 'autoinstall'>> &
   Pick<
     Opts,
     | 'expandAdaptors'
@@ -19,13 +19,14 @@ const options = [
   o.expandAdaptors, // order is important
 
   o.adaptors,
+  o.autoinstall,
   o.force,
   o.log,
   o.logJson,
   o.repoDir,
   o.statePath,
   o.stateStdin,
-  o.useAdaptorsMonorepo,
+  o.useAdaptorsMonorepo
 ];
 
 export default {
--- a/packages/cli/src/metadata/handler.ts
+++ b/packages/cli/src/metadata/handler.ts
@@ -49,7 +49,7 @@ export const getAdaptorPath = async (
 };
 
 const metadataHandler = async (options: MetadataOpts, logger: Logger) => {
-  const { repoDir, adaptors } = options;
+  const { repoDir, adaptors, autoinstall } = options;
   const adaptor = adaptors[0];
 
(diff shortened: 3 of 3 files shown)


SatyamMattoo:
@josephjclark Could you please take a look at the changes I've made for this issue? Your feedback and recommendations would be greatly appreciated. Thanks in advance!

josephjclark (changes requested):
Hi @SatyamMattoo!

Thank you so much for reaching out with a contribution.

Unfortunately the solution here is incorrect. But that's not surprising becuase the issue doesn't give you much to work with! That's my fault and I'm sorry.

I've edited the issue and added the details you need to get started. You want to read up a bit about openfn and the CLI and our adaptors on docs.openfn.org.

In situations like this it's usually a good idea to reach out and ask for more details before getting started. If the requirements are not clear, you are unlikely to submit a valid solution.

I'd also suggest that you should not pick up a piece of work unless you can work out 
 how to test it. You absolutely must be able to manually test the issue, and you should include a unit test with any PR you open.
 
If you can't test it, that's a good sign that the requirements are too vague and you need to ask for help :)

Thanks again! I'd be delighted if you could take a look at the updated issue and have another go at this.

SatyamMattoo:
Hello @josephjclark,

Thank you for your feedback and guidance. I appreciate your suggestions on improving the quality of my contributions.

I've taken note of your advice regarding reaching out for clarification before starting work and ensuring that I can test the solution effectively. Moving forward, I'll make sure to thoroughly understand the requirements and include proper testing procedures in my contributions.

Regarding your suggestion to include unit tests. As I'm relatively new to testing, could you please provide some sample code or point me to some resources that I can reference while writing the tests to this issue? This would greatly help me in ensuring the quality of my code.

Once again, thank you for your support and I look forward to working on the updated issue.

Best regards

josephjclark:
Hi @SatyamMattoo,

We use [ava](https://github.com/avajs/ava) for our unit testing. You'll find many tests in this repo. We generally try to ensure that every file in a `src` package has a corresponding `*.test.ts` file in the `test` folder.

There are more detailed thoughts about testing this particular feature in the issue itself.

But outside of automated tests, my point here is that I don't think you've done anything to try this code out and see if it works. I don't think you've run `openfn metadata` on your machine :thinking: 

While I appreciate that the issue didn't give you enough context to be able to do that, you shouldn't ever submit code to a repo that you haven't run yourself. Otherwise, how do you know if it works? What confidence do you give to the reviewer that your solution works?

I will close down this pull request for now, but please do re-open it or create a new one when you're ready :pray:

PR #668, opened 2024-04-16
Fix worker death

### LINKED ISSUE #664: Worker Death
In production this afternoon we had a nasty case of worker death

image

[Link to logs](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.project_id%3D%22platform-test-267207%22%0Aresource.labels.location%3D%22europe-west6%22%0Aresource.labels.cluster_name%3D%22swiss-ap-1%22%0Aresource.labels.namespace_name%3D%22default%22%0Alabels.k8s-pod%2Fapp%3D%22global-worker%22%20severity%3E%3DDEFAULT;cursorTimestamp=2024-04-15T14:06:06.786281404Z;startTime=2024-04-15T13:07:03.623Z;endTime=2024-04-15T16:07:03.623884Z?authuser=0&organizationId=189089165982&project=platform-test-267207)

A couple of things to note:
* The worker is queuing requests internally. Like the worker accepts a job, hands it to the child process pool,  and the pool enqueues it. That shouldn't be happening
* It looks like we had some errors and then maybe timeouts after logs?
* Mutchi mentioned a postgres bug around this time

Suspicion 1: Child processes are dying and not being recreated (but maybe they;'re managing to send out an error to the worker so that the queue is getting freed up)

Suspicion 2: Postgres is erroring and calling process.exit(), and the exit is not properly handled

josephjclark:
## Short Description

This PR fixes an issue where the worker doesn't close down pooled child processes after uncaught exceptions, resulting in the worker refusing to work.

## Related issue

Fixes #664 

## Implementation Details

What's basically happening is:

- The engine is correctly catching uncaught exceptions (and I think process.exits) when they occur
- And it is correctly sending error events out (these errors may be rubbish but I do think they are sending)
- BUT the engine doesn't properly resolve the promise for that run, so the engine still thinks the workflow is executing
- That means the child process pool doesn't allocate new resources for future work, and so chokes up

The engine should eventually be timing out the runs, but by this point lightning thinks they're dead and isn't listening to events. But I think the backlog will very slowly clear.

Anyway, as result of the fix, the error is handled gracefully, the pool re-allocates the worker thread, and everyone is happy.

## QA Notes

I've added two integration tests, both of which  reproduce very similar errors to main. And they both fail on main.

- [x] I have added unit tests

### DIFF
--- a/packages/engine-multi/src/worker/pool.ts
+++ b/packages/engine-multi/src/worker/pool.ts
@@ -93,9 +93,9 @@ function createPool(script: string, options: PoolOptions = {}, logger: Logger) {
       // Note: Ok, now I have visibility on the stdout stream
       // I don't think I want to send this to gpc
       // This might be strictly local debug
-      child.stdout!.on('data', (data) => {
-        console.log(' >>>', data.toString());
-      });
+      // child.stdout!.on('data', (data) => {
+      //   console.log(data.toString());
+      // });
 
       logger.debug('pool: Created new child process', child.pid);
       allWorkers[child.pid!] = child;
--- a/packages/engine-multi/src/worker/pool.ts
+++ b/packages/engine-multi/src/worker/pool.ts
@@ -87,7 +87,14 @@ function createPool(script: string, options: PoolOptions = {}, logger: Logger) {
         env: options.env || {},
 
         // This pipes the stderr stream onto the child, so we can read it later
-        stdio: ['ipc', 'ignore', 'pipe'],
+        stdio: ['ipc', 'pipe', 'pipe'],
+      });
+
+      // Note: Ok, now I have visibility on the stdout stream
+      // I don't think I want to send this to gpc
+      // This might be strictly local debug
+      child.stdout!.on('data', (data) => {
+        console.log(' >>>', data.toString());
       });
 
       logger.debug('pool: Created new child process', child.pid);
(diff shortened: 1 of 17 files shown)


packages/engine-multi/src/worker/pool.ts
```diff
@@ -87,9 +87,16 @@ function createPool(script: string, options: PoolOptions = {}, logger: Logger) {
...
 
         // This pipes the stderr stream onto the child, so we can read it later
-        stdio: ['ipc', 'ignore', 'pipe'],
+        stdio: ['ipc', 'pipe', 'pipe'],
       });
 
+      // Note: Ok, now I have visibility on the stdout stream
+      // I don't think I want to send this to gpc
+      // This might be strictly local debug
+      // child.stdout!.on('data', (data) => {
```
josephjclark, line 96: This is incidental debugging code. I do want to leave it here but commented out for now.

packages/engine-multi/src/worker/thread/helpers.ts
```diff
@@ -66,17 +65,23 @@ export const createLoggers = (
...
       workflowId,
       // Map the error out of the thread in a serializable format
       error: serializeError(err),
-      stack: err?.stack
+      stack: err?.stack,
       // TODO job id maybe
     });
+
+    // Explicitly send a reject task error, to ensure the worker closes down
+    publish(workerEvents.ENGINE_REJECT_TASK, {
```
josephjclark, line 82: This is the actual fix and this is what WASN'T happening before.

It's not enough to tell the worker we've errored. We need to send an event to the parent process to tell it that we're dead.

I'm a little concerned about duplicating error messages here and so there's a little bit of gnarly code on the reporting side to handle that.

PR #670, opened 2024-04-18
Fixing:#192

### LINKED ISSUE #192: CLI metadata: allow autoinstall

SatyamMattoo:
## Short Description

The fix includes a function responsible for adaptor validation and installation utilizing the existing install function from the repo/handler module.

## Related issue

Fixes #192 

## Implementation Details

To prevent potential failures, I propose implementing an automatic adaptor installation feature. This would validate the presence of the required adaptor and install it if necessary, ensuring a seamless user experience. I've developed and tested the solution thoroughly, including unit tests and compatibility checks with existing integration tests.

- [x] I have added unit tests

### DIFF
--- a/packages/cli/src/metadata/handler.ts
+++ b/packages/cli/src/metadata/handler.ts
@@ -4,6 +4,8 @@ import loadState from '../util/load-state';
 import cache from './cache';
 import { getModuleEntryPoint } from '@openfn/runtime';
 import { ExecutionPlan } from '@openfn/lexicon';
+import { install } from '../repo/handler';
+import getAutoinstall from './get-auto-install';
 
 // Add extra, uh, metadata to the, uh, metadata object
 const decorateMetadata = (metadata: any) => {
@@ -81,7 +83,13 @@ const metadataHandler = async (options: MetadataOpts, logger: Logger) => {
   }
 
   try {
-    // Import the adaptor
+    //Install and import the adaptor
+    const autoInstall = getAutoinstall(adaptor);
+    if (autoInstall) {
+      logger.info('Auto-installing language adaptors');
+      await install({ packages: [adaptor], repoDir }, logger);
+    }
+
     const adaptorPath = await getAdaptorPath(adaptor, logger, options.repoDir);
     const mod = await import(adaptorPath!);
 
--- a/packages/cli/src/metadata/get-auto-install.ts
+++ b/packages/cli/src/metadata/get-auto-install.ts
@@ -1,0 +1,10 @@
+const getAutoinstall = (adaptor: string): boolean => {
+
+  if (adaptor.includes("=") || adaptor.startsWith("/") || !adaptor.length) {
+    return false;
+  }
+  return true
+};
+
+
+export default getAutoinstall;
--- a/packages/cli/test/metadata/get-autoinstall.test.ts
+++ b/packages/cli/test/metadata/get-autoinstall.test.ts
@@ -1,0 +1,38 @@
+import test from 'ava';
+import getAutoinstall from '../../src/metadata/get-auto-install';
+
+test('should return false when no adaptor is provided', (t) => {
+  const adaptor = '';
+  const result = getAutoinstall(adaptor);
+  t.is(result, false);
+});
+
+test('should return false for adaptors with "test=" prefix', (t) => {
+  const adaptor = 'test=/home/satyammattoo/kit/integration-tests/cli/modules/test';
+  const result = getAutoinstall(adaptor);
+  t.is(result, false);
+});
+
+test('should return false for adaptors that are full module paths', (t) => {
+  const adaptor = '/home/satyammattoo/kit/integration-tests/cli/modules/test';
+  const result = getAutoinstall(adaptor);
+  t.is(result, false);
+});
+
+test('should return true for adaptors with a name', (t) => {
+  const adaptor = 'common';
+  const result = getAutoinstall(adaptor);
+  t.is(result, true);
+});
+
+test('should return true for adaptors with a name and version', (t) => {
+  const adaptor = '@openfn/language-common@1.0.0';
+  const result = getAutoinstall(adaptor);
+  t.is(result, true);
+});
+
+test('should return false for adaptors that are full paths', (t) => {
+  const adaptor = '@openfn/language-common';
+  const result = getAutoinstall(adaptor);
+  t.is(result, true);
+});


packages/cli/src/metadata/handler.ts
```diff
@@ -81,7 +83,13 @@ const metadataHandler = async (options: MetadataOpts, logger: Logger) => {
   }
 
   try {
-    // Import the adaptor
+    //Install and import the adaptor
+    const autoInstall = getAutoinstall(adaptor);
+    if (autoInstall) {
+      logger.info('Auto-installing language adaptors');
```
josephjclark, line 89: This log line is redundant, the install function does all the logging we need.

For the record, here's the actual installl output:
```
[CLI] ℹ Auto-installing language adaptors
[CLI] ✔ Installing packages...
[CLI] ℹ Looked up latest version of @openfn/language-salesforce: found 4.6.2
[CLI] ℹ Will install @openfn/language-salesforce version 4.6.2
[CLI] ✔ Installed @openfn/language-salesforce@4.6.2
[CLI] ✔ Installation complete in 10.526s
```

josephjclark (changes requested):
Pretty good @SatyamMattoo!

I notice there's no integration test, but I'd really like to see one (as outlined in the issue). It's OK for the integration test to autoinstall but then fail to generate metadata. You can run it against the common adaptor.

You can test this locally by running `pnpm openfn metadata -a common -s tmp/state.json` from `packages/cli`.

You'll need to create an empty `tmp/state.json`, like this:
```
{
  "configuration": {
    "user": "satyam",
  }
}
```

Try adding `--log debug` to see a different level out output.

josephjclark:
I've noticed out CI tests haven't run on this, I think because it's targeted at main. I wonder if I can address this...

SatyamMattoo:
@josephjclark Sure, I'll look into these issues and include the integration tests for the same. Thankyou.

SatyamMattoo:
Hello @josephjclark,

I've implemented the requested changes, which include:

1. Implementation of the auto-install function within the handler.ts file.
2. Renaming of the function to shouldAutoinstall.
3. Removal of redundant logs.
4. Addition of an integration test to verify the auto-installation of the adaptor ("common").

Please review these modifications at your convenience and inform me of any further adjustments or additions you deem necessary.

Best regards

packages/cli/src/metadata/handler.ts
```diff
@@ -48,6 +49,14 @@ export const getAdaptorPath = async (
   return adaptorPath;
 };
 
+export const shouldAutoinstall = (adaptor: string): boolean => {
```
josephjclark, line 52: This still feels quite heavy and verbose. I would write it as:
```
export const shouldAutoinstall = (adaptor: string): boolean => adaptor?.length && !adaptor.startsWith('=') && !adaptor.includes('=')
```

packages/cli/src/metadata/handler.ts
```diff
@@ -81,7 +90,12 @@ const metadataHandler = async (options: MetadataOpts, logger: Logger) => {
   }
 
   try {
-    // Import the adaptor
+    //Install and import the adaptor
```
josephjclark, line 93: This comment isn't actually useful, would you mind just removing it?

packages/cli/src/metadata/handler.ts
```diff
@@ -81,7 +90,12 @@ const metadataHandler = async (options: MetadataOpts, logger: Logger) => {
...
+    const autoInstall = shouldAutoinstall(adaptor);
+    if (autoInstall) {
```
josephjclark, line 95: You could inline the condition for brevity. It'll still read really well:
```
if (shouldAutoinstall(adaptor)) {
   ...
}
```

packages/cli/test/metadata/should-autoinstall.test.ts
```diff
@@ -0,0 +1,38 @@
+import test from 'ava';
+import {shouldAutoinstall} from '../../src/metadata/handler'
+
+test('should return false when no adaptor is provided', (t) => {
+  const adaptor = '';
+  const result = shouldAutoinstall(adaptor);
+  t.is(result, false);
```
josephjclark, line 7: You should replace `t.is(result, false)` with `t.false(result)`. It's a bit more concise and we'll get a nicer error on fail.

packages/cli/test/metadata/should-autoinstall.test.ts
```diff
@@ -0,0 +1,38 @@
...
+});
+
+test('should return false for adaptors with "test=" prefix', (t) => {
+  const adaptor = 'test=/home/satyammattoo/kit/integration-tests/cli/modules/test';
```
josephjclark, line 11: Hmm, it's a silly aesthetic thing, but we don't need such a long path here.

Can we replace this (and the next path on line 17) with something shorter and simpler, like:
```
const adaptor = 'test=/repo/modules/test';
```

integration-tests/cli/test/metadata.test.ts
```diff
@@ -126,3 +126,18 @@ test.serial('does not log credentials', async (t) => {
...
+
+// Test for auto-installing adaptor
+test.serial(
+  `openfn metadata -f -S "${state}" -a common --log-json --log info`,
+  async (t) => {
+    const { stdout } = await run(t.title);
+
+    t.regex(stdout, /Generating metadata/);
+    t.regex(stdout, /Installing packages.../);
+    t.regex(stdout, /Installed @openfn\/language-common@1.13.1/);
```
josephjclark, line 138: This will fail later because the version of common will change.

You either need to use a more forgiving regex, like `/Installed @openfn\/language-common@` (which is fine), or lock the version in the command above, ie by adding `-a common@1.31.1`

I'm happy with both solutions - perhaps you can pick your favourite and tell me why?

integration-tests/cli/test/metadata.test.ts
```diff
@@ -126,3 +126,18 @@ test.serial('does not log credentials', async (t) => {
...
+    t.regex(stdout, /Installation complete in \d+\.\d+s/);
+    t.regex(stdout, /No metadata helper found/);
+    t.notRegex(stdout, /Returning metadata from cache/);
```
josephjclark, line 141: This line doesn't feel critical to the test story.  I would remove it.

integration-tests/cli/test/metadata.test.ts
```diff
@@ -126,3 +126,18 @@ test.serial('does not log credentials', async (t) => {
   t.regex(logString, /(pin_number)/i);
   t.notRegex(logString, new RegExp(sensitiveValue), 'i');
 });
+
+// Test for auto-installing adaptor
```
josephjclark, line 130: Can we add a line of comment along the lines of 

> This test should auto-install the common adaptor but then happily fail to generate metadata

It's a slightly strange test, conceptually, and I just want to clarify that we don't expect any metadata output

josephjclark (changes requested):
Thank you @SatyamMattoo, that integration test is great.

I've had a really close look and picked out a couple of tiny things that need adjusting, mostly just style, but then we should be good to go!

integration-tests/cli/test/metadata.test.ts
SatyamMattoo: Hey, I have decided to improve the regex syntax instead of specifying the version of the adaptor for the following reasons:
1. This approach is resilient to changes in the version of the common package. It looks for a general pattern rather than a specific version number, making it more future proof.
2. As new versions of the common package are released, as long as the installation message format remains consistent, the regex will still match.

SatyamMattoo:
Hi @josephjclark,

Thanks for your detailed review! Your suggestions really helped improve the code quality. I've made all the requested changes. Please take a look whenever you get a chance. Let me know if there's anything else that needs improvement.

Best regards

josephjclark:
This is great @SatyamMattoo, thank you for your patience!

josephjclark:
@SatyamMattoo Your fix is now live in [@openfn/cli@1.2.3](https://www.npmjs.com/package/@openfn/cli/v/1.2.3)  :tada:

PR #675, opened 2024-04-21
Fixing:#667 

### LINKED ISSUE #667: Allow steps/jobs in different workflows to have the same name
The current implementation of openfn pull returns an error when a job name is duplicated in two different workflows. The duplicate job name constraint should only apply on workflows i.e. No two jobs within a workflow should have the same name. 

See error message below:
Image

SatyamMattoo:
## Short Description

CLI: Allow steps/jobs in different workflows to have the same name.

## Related issue

Fixes #667

## Implementation Details

Changed the validator function to check for duplicates in a single workflow rather than the whole project, enabling same job/step names across different workflows. Created different sets for workflow and job names to return a valid error in both cases.

## QA Notes

I have added a test case for the issue and also changed one which has same job name in different workflows.

- [x] I have added unit tests

### DIFF
--- a/packages/deploy/src/validator.ts
+++ b/packages/deploy/src/validator.ts
@@ -13,23 +13,20 @@ export function parseAndValidate(input: string): {
   doc: ProjectSpec;
 } {
   let errors: Error[] = [];
-  let workflowKeys: Set<string> = new Set();
-  let jobKeys: Set<string> = new Set();
   const doc = YAML.parseDocument(input);
 
-  function pushUniqueKey(context: YAML.Pair, key: string, keyType: 'workflow' | 'job') {
-    const keyList = keyType === 'workflow' ? workflowKeys : jobKeys;
-    if (keyList.has(key)) {
+  function pushUniqueKey(context: YAML.Pair, key: string, arr: string[]) {
+    if (arr.includes(key)) {
       errors.push({
         context,
-        message: `duplicate ${keyType} key: ${key}`,
+        message: `duplicate key: ${key}`,
       });
     } else {
-      keyList.add(key);
+      arr.push(key);
     }
   }
 
-  function validateJobs(workflow: YAMLMap) {
+  function validateJobs(workflow: YAMLMap, jobKeys: string[]) {
     const jobs = workflow.getIn(['jobs']);
 
     if (jobs) {
@@ -37,7 +34,7 @@ export function parseAndValidate(input: string): {
         for (const job of jobs.items) {
           if (isPair(job)) {
             const jobName = (job as any).key.value;
-            pushUniqueKey(job, jobName, 'job');
+            pushUniqueKey(job, jobName, jobKeys);
           }
         }
       } else {
@@ -55,14 +52,15 @@ export function parseAndValidate(input: string): {
       // map to avoid errors downstream
       doc.setIn(['workflows'], {});
     } else if (isMap(workflows)) {
+      const workflowKeys: string[] = [];
       for (const workflow of workflows.items) {
         if (isPair(workflow)) {
           const workflowName = (workflow as any).key.value;
-          pushUniqueKey(workflow, workflowName, 'workflow');
+          const jobKeys: string[] = [];
+          pushUniqueKey(workflow, workflowName, workflowKeys);
           const workflowValue = (workflow as any).value;
           if (isMap(workflowValue)) {
-            jobKeys = new Set(); // Reset job keys for each workflow
-            validateJobs(workflowValue);
+            validateJobs(workflowValue, jobKeys);
           } else {
             errors.push({
               context: workflowValue,
--- a/packages/deploy/src/validator.ts
+++ b/packages/deploy/src/validator.ts
@@ -13,17 +13,19 @@ export function parseAndValidate(input: string): {
   doc: ProjectSpec;
 } {
   let errors: Error[] = [];
-  let keys: string[] = [];
+  let workflowKeys: Set<string> = new Set();
+  let jobKeys: Set<string> = new Set();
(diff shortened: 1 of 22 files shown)


josephjclark:
Hi @SatyamMattoo ! I'm sorry I only just saw your PR this morning. Thank you for another contribution!

I'll take a look today and get back to you.

packages/deploy/src/validator.ts
```diff
@@ -13,17 +13,19 @@ export function parseAndValidate(input: string): {
...
 } {
   let errors: Error[] = [];
-  let keys: string[] = [];
+  let workflowKeys: Set<string> = new Set();
+  let jobKeys: Set<string> = new Set();
   const doc = YAML.parseDocument(input);
 
-  function pushUniqueKey(context: YAML.Pair, key: string) {
-    if (keys.includes(key)) {
+  function pushUniqueKey(context: YAML.Pair, key: string, keyType: 'workflow' | 'job') {
```
josephjclark, line 20: Instead of using two top-level variables and selecting them based on as string, can we just pass the correct collection into this function?
```
function pushUniqueKey(context: YAML.Pair, key: string, arr: string[]) {
  ...
}
```

So `pushUniqueKey` means: "add a key to THIS array BUT only if it doesn't exist"

This change also makes the function a bit more pure, which is nice.

Once you've done this, I also think you should declare the two arrays further down, inside the  validation loop. I'll comment where I mean.

packages/deploy/src/validator.ts
```diff
@@ -13,17 +13,19 @@ export function parseAndValidate(input: string): {
   doc: ProjectSpec;
 } {
   let errors: Error[] = [];
-  let keys: string[] = [];
+  let workflowKeys: Set<string> = new Set();
```
josephjclark, line 16: It is strange to have a set here. Normally a set would ensure that a value is only added once.

But here, we have to check for membership of the set before we add it anyway. It's no different to using an array - and an array is a simpler and more commonly used structure.

So I'd prefer both of these key sets to actually be arrays - it's just simpler.

packages/deploy/src/validator.ts
```diff
@@ -50,14 +53,23 @@ export function parseAndValidate(input: string): {
...
       for (const workflow of workflows.items) {
         if (isPair(workflow)) {
-          pushUniqueKey(workflow, (workflow as any).key.value);
-
-          validateJobs((workflow as any).value);
+          const workflowName = (workflow as any).key.value;
+          pushUniqueKey(workflow, workflowName, 'workflow');
+          const workflowValue = (workflow as any).value;
+          if (isMap(workflowValue)) {
+            jobKeys = new Set(); // Reset job keys for each workflow
```
josephjclark, line 64: Instead of "resetting" the jobKeys, we should create a new array as a const. In other words, declare the array here, rather than higher up

packages/deploy/src/validator.ts
```diff
@@ -50,14 +53,23 @@ export function parseAndValidate(input: string): {
     if (typeof workflows === 'undefined') {
       // allow workflows to be unspecified, but ensure there is an empty
       // map to avoid errors downstream
-      doc.setIn(['workflows'], {})
-    }
-    else if (isMap(workflows)) {
+      doc.setIn(['workflows'], {});
+    } else if (isMap(workflows)) {
```
josephjclark, line 57: Please declare the `workflowKeys` array here, inside this if statement (but outside of the for, obviously).

That scopes the variable to where it's actually used, which feels neater

packages/deploy/test/validator.test.ts
```diff
@@ -40,11 +40,29 @@ workflows:
...
   results = parseAndValidate(doc);
 
-  t.truthy(findError(results.errors, 'duplicate key: workflow-one'));
-
-  t.truthy(findError(results.errors, 'duplicate key: foo'));
+  t.truthy(findError(results.errors, 'duplicate workflow key: workflow-one')); 
 
   t.truthy(findError(results.errors, 'jobs: must be a map'));
+  
+  doc = `
```
josephjclark, line 47: I had never seen these unit tests before and they are not well structured :scream: 

I won't ask you to fix that - you've done the right thing by sticking to the established pattern.

But could you please add a little comment at the top of this `doc` declaration, something like `disallow two jobs of the same name in the same workflow`. Just to make it clear to the next developer what's actually being tested.

josephjclark (changes requested):
Thank you for this @SatyamMattoo! It looks like a good solution, well done.

I've left a couple comments - mostly I'd like to remove the sets and simplify `pushUniqueKey` by passing the collection straight into it.

Once you've done this I'll get this merged next week.

Sorry again for the delayed response :pray:

SatyamMattoo:
Thank you @josephjclark for the detailed review. I'll make the required changes as soon as possible and push it for further consideration.

packages/deploy/src/validator.ts
SatyamMattoo: Thanks for the suggestion! Simplifying pushUniqueKey by removing keyType makes it more focused. Declaring arrays inside validateWorkflows enhances organization. I initially added keyType for clearer error messages to users.

packages/deploy/test/validator.test.ts
SatyamMattoo: Yes, I think there should have been different tests for each case. For now, I have followed the previous pattern, but I'll be happy to help if there are any changes needed.

SatyamMattoo:
Hello @josephjclark I have pushed the requested changes. You can review the changes whenever you get a chance. Let me know if there's anything else that needs improvement. 

Best regards!

josephjclark (approved):
Thank you @SatyamMattoo :pray:

PR #685, opened 2024-05-10
Worker: ensure workers are released after an error on run:complete

josephjclark:
## Short Description

If the run:complete event error happens to throw, the worker will not close out a Run properly, ie will not free up capacity and claim another run.

This is causing the worker to "jam" silently.

I do not know WHY `run:complete` is timing out on us right now (see [these logs](https://console.cloud.google.com/logs/query;cursorTimestamp=2024-05-10T00:06:42.140836231Z;query=labels.%22k8s-pod%2Fapp%22%3D%22global-worker%22%0Atimestamp%3D%222024-05-10T00:06:42.140836231Z%22%0AinsertId%3D%22pgvd1lgqsoczh9gr%22?organizationId=189089165982&project=platform-test-267207)). And we'll have to work that out. But at least the worker shouldn't die silently any more.

I've also added a little bit more logging around this stuff. Most of the time it's junk but it may help us diagnose when the worker isn't releasing threads from the pool.

## Related issue

None raised, see [slack](https://openfn.slack.com/archives/C0394D0GBUN/p1715326454971499).

### DIFF
--- a/.changeset/tidy-gorillas-own.md
+++ b/.changeset/tidy-gorillas-own.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Fix an issue where workers may not be returned to the pool if run:complete throws
--- a/.changeset/tidy-gorillas-own.md
+++ b/.changeset/tidy-gorillas-own.md
@@ -1,5 +1,0 @@
----
-'@openfn/ws-worker': patch
----
-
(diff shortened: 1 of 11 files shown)


josephjclark:
@taylordowns2000 This is ready to go and the image is built. Can we make sure we run the end to end tests on it before going live?

PR #696, opened 2024-05-21
Runtime: adding more validation to workflow

### LINKED ISSUE #681: CLI: add better validation on workflows
When loading a workflow JSON, the CLI should do more validation work, warning users when a hand-made workflow is incorrect.

The runtime actually has a `validate-plan.ts` file - it should be acceptable to just add extra validation rules here. They should bubble up to the CLI (for example, a workflow.json which specifies a start step not defined in the workflow will throw a validation error, ie `{ options: { start: "x" }, workflow: { steps: [ { id: "a" }] } }`

Validation rules should include:

* Check the overall structure: a workflow must have a `workflow` key  at the root, and the `workflow` must have a `steps` key, which should be an array. 
* Log a warning if the steps array is empty
* A step with an adaptor should also have an expression (steps without expressions are fine)
* A step should only have the valid set of keys defined - any keys not found in the type reference should trigger a fail.
* Log a warning if any unrecognised options are a passed on the `options` object

Plus anything else that may be useful

All  validation rules should be thoroughly unit tested.

A big risk is that adding extra validation will break unit tests, which often delight in using invalid workflows. I think there's already a `skipValidation` option which we can pass to the runtime and we may need to add this. If this is breaking lots and lots and lots of tests we may want to re-think (perhaps validation should be opt-in, and we need to change the CLI to always pass a `validate: true` flag)

SatyamMattoo:
## Short Description

Added functions to validate the workflow, steps and options and return errors or warnings accordingly

## Related issue

Fixes #681 

## Implementation Details

Added more validation functions to `validate-plan.ts` returning errors or warnings for a bad workflow structure.

## QA Notes

The following approach is not breaking any existing test cases. But we can add new test
cases where we can check the warning logs are printed successfully

- [x] I have added unit tests


josephjclark:
Hi @SatyamMattoo , just to check, this and other issues aren't covered by the DMP programme. This has no relation to c4gt and does not influence  any DMP application you make.

If you want to take on a project under DMP - and I hope you will! - you need to apply directly through the C4GT portal. I don't know anything about that, I'm not involved in the process at all.

SatyamMattoo:
Hey @josephjclark 
Sorry for the misunderstanding, I forgot to change the branch name here because I was working on a different issue as well. I have applied for the DMP program as well and this isn't related to that. You can have a look and suggest any changes that might be needed here. 

Best regards!

packages/runtime/src/util/validate-plan.ts
```diff
@@ -20,6 +24,54 @@ export default (plan: ExecutionPlan) => {
...
+  if (workflow.steps.length === 0) {
+    logger.warn('Warning: The workflow.steps array is empty.');
+  }
+
+  workflow.steps.forEach((step, index) => {
+    assertStepStructure(step, index);
+  });
+
+  if (options) {
+    assertOptionsStructure(options);
```
josephjclark, line 47: You can remove this if statement if you default `options` to an empty object inside `assertOptionStructure`, resulting in more elegant code :)

```
const assertOptionsStructure = (options: WorkflowOptions = {}) => { ... }
```

packages/runtime/src/util/validate-plan.ts
```diff
@@ -20,6 +24,54 @@ export default (plan: ExecutionPlan) => {
...
+
+  if ('adaptor' in step && !('expression' in step)) {
+    throw new ValidationError(`Step ${index} with an adaptor must also have an expression.`);
+  }
+};
+
+const assertOptionsStructure = (options: WorkflowOptions) => {
+  const allowedKeys = ['timeout', 'stepTimeout', 'start', 'end', 'sanitize'];
+
+  Object.keys(options).forEach((key) => {
```
josephjclark, line 68: Try using `for in`:

```
for (const key in options ) {
  if (!allowedKeys.includes(key)) {
      logger.warn(`Warning: Unrecognized option "${key}" in options object.`);
    }
}
```

packages/runtime/src/util/validate-plan.ts
```diff
@@ -20,6 +24,54 @@ export default (plan: ExecutionPlan) => {
...
+    if (!allowedKeys.includes(key)) {
+      logger.warn(`Warning: Unrecognized option "${key}" in options object.`);
```
josephjclark, line 70: Please don't include full stops after log messages. They're unusual and unnecessary in log output

josephjclark (changes requested):
Hi @SatyamMattoo 

I've left a few comments on particular details.

But after thinking about it for a few days, I really think this level of validation needs to move into the `cli` package. Having these kind of heavyweight validation steps in the runtime is just uneccessary

In `packages/cli/src/execute/handler.ts` there is a call to `loadPlan` inside `executeHandler. I would add a call to `validatePlan` after this, and copy your validation rules into new file in `cli/src/util/`

The existing validation in the runtime, which validates the logical consistency of the execution plan, needs to stay where it is. But these new structural/syntactic validations need to move out and into the CLI.

Sorry for the inconvenience - this is on my for specifying bad requirements in the first place. I just didn't know until I saw the code.

packages/runtime/src/util/validate-plan.ts
```diff
@@ -1,16 +1,20 @@
...
+import { ExecutionPlan, Job, Step, Trigger, WorkflowOptions } from '@openfn/lexicon';
 import { ValidationError } from '../errors';
+import createLogger from '@openfn/logger';
 
 type ModelNode = {
   up: Record<string, true>;
   down: Record<string, true>;
 };
 
+const logger = createLogger(undefined, { json: true });
```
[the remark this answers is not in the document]
SatyamMattoo: Do we need to create a logger in the main function and then pass it to the functions like 
`const assertWorkflowStructure = (plan: ExecutionPlan, logger: Logger)` ?

Additionally, should we assign a name to the logger when creating it?
```
export default (plan: ExecutionPlan) => {
  const logger = createLogger(undefined, { json: true });
  assertWorkflowStructure(plan, logger);
```

SatyamMattoo:
Hello @josephjclark,

I have made the requested changes. Please review them at your convenience and let me know if there are any additional modifications needed.

Best regards.

josephjclark:
Thank you @SatyamMattoo - I'll get back to you next week when I have time :)

josephjclark:
Hi @SatyamMattoo actually I've found a spare 20 minutes so  I'm taking a look now.

I don't think you've seem my [main comment](https://github.com/OpenFn/kit/pull/696#pullrequestreview-2070233256) on the last review. We need to move these new validation steps out of the runtime and into the CLI.

SatyamMattoo:
Hi @josephjclark I have moved the validation from the runtime to the CLI and also added a few more tests for the logs. You can have a look at your convenience and let me know if there are any more changes needed.

Best regards.

packages/cli/test/util/validate-plan.test.ts
```diff
@@ -0,0 +1,118 @@
...
+import validate from '../../src/util/validate-plan';
+
+const logger = createMockLogger('', { level: 'debug' });
+
+test.afterEach(() => {
+    logger._reset();
+})
+
+test('throws for missing workflow', (t) => {
+    //@ts-ignore
```
josephjclark, line 13: These `@ts-ignore`s are pretty awkward.

What I usually do is force a type assertion:
```
 const plan = {
        options: {
            start: 'a',
        }
    } as ExecutionPlan;
```

This usually does the trick. Sometimes you need to hack it with `as unknown as ExecutionPlan`

packages/cli/test/util/validate-plan.test.ts
```diff
@@ -0,0 +1,118 @@
...
+            }
+        },
+    };
+
+    t.throws(() => validate(plan, logger), {
+        message: 'The workflow.steps key must be an array',
+    });
+});
+
+test('throws for adaptor without an expression', (t) => {
```
josephjclark, line 45: This title should be: `throws for a step with an adaptor but no expression`

packages/cli/test/util/validate-plan.test.ts
```diff
@@ -0,0 +1,118 @@
...
+test('throws for missing workflow', (t) => {
+    //@ts-ignore
+    const plan: ExecutionPlan = {
+        options: {
+            start: 'a',
+        }
+    };
+
+    t.throws(() => validate(plan, logger), {
+        message: 'Missing or invalid workflow key in execution plan',
```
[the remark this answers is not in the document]
SatyamMattoo: This error message was actually related to the 'workflow' key, so I have changed the message to 
`throw new Error("Missing or invalid 'workflow' key in execution plan");`

SatyamMattoo:
Hey @josephjclark, I have committed the requested changes. You can review them at your convenience and let me know if there are any further changes required.

Sorry for the late response, I was occupied with my exams recently.

Best regards.

packages/cli/test/util/validate-plan.test.ts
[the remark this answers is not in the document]
josephjclark: Ah I see! It would be better english to say `Missing or invalid key "workflow" in execution plan` (double quotes is deliberate)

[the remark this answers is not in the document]
SatyamMattoo: Done!

josephjclark (approved):

josephjclark:
Thank you @SatyamMattoo - this will be released shortly!

PR #697, opened 2024-05-23
deploy: improve error messages

josephjclark:
I'm trying to improve the error messaging when validating project.yaml files

So far I've not been able to make much improvement actually. I think the problems are coming out of the native yaml parser and we're not doing a good job of reporting those.

I've also managed to introduce lots of random errors by indenting the yaml at various places. these trigger blowups without any useful context. Not really sure what to do about these yet.

This might be a much bigger deal than I first thought

Maybe closes OpenFn/lightning#2140

Implementation notes:

* I've refactored the validation tests
* I've refactored "duplicate key" to "duplicate id", which is more user friendly
* I'm ensuring a simple string path on an error object (this may wrong, I may have to revert to an array for consistency with the yaml validator)

### DIFF
--- a/packages/deploy/src/index.ts
+++ b/packages/deploy/src/index.ts
@@ -111,7 +111,7 @@ export async function deploy(config: DeployConfig, logger: Logger) {
 
   logger.debug('spec', spec);
   if (spec.errors.length > 0) {
-    spec.errors.forEach((e: any) => logger.warn(e.message));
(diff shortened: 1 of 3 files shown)


josephjclark:
Ok so while this PR does have improvements, the vague error messages in question are not coming from the CLI, they're coming straight outta lightning.

I will merge this with the next CLI release but otherwise won't push it any further

PR #704, opened 2024-05-28
Engine: support multiple inputs to a step

### LINKED ISSUE #698: Runtime: support multiple inputs
The runtime needs to be able to support  multiple inputs to a node.

This doesn't need to do anything particularly clever - if multiple edges lead to a node, that node gets called multiple times. We just need to ensure the correct input state is used - but I think we've already solved that.

We need to remove the validation that enforces that the workflow must be a tree.

We currently have loop detection - I suggest we keep that in place.

I am hoping that because this isn't technically hard, and we've got solid and simple execution rules, that this won't be more than day's work.

josephjclark:
This PR enables the runtime call each step multiple times.

Each time a step is called, it passes its output to any downstream steps, and so the downstream steps are run multiple times.

It basically treats a repeated step as if it were a unique step in the workflow. So if step `x` is run twice, the runtime basically runs `x-0` and `x-1` as if there were defined on the plan. Note that actually call `x-0` `x`, because otherwise 8 million tests break. And I think it's a bit cleaner?

## Issue

Closes #698

## Changes

* Removed a validation step that looks for steps with multiple inputs and throws
* Adjust the core runtime to ensure that a step can have multiple inputs AND outputs
* Update unit tests

## Important Behavioural notes

These are the rules of this feature:

* Execution is in SERIES and BREADTH FIRST. If start goes to a and b, and a and b go to x, the execution order is `start -> a -> b -> x -> x`
* State is isolated between steps. So when x is called twice, it is with two different state objects. This is what you want, but might not be what you expect. You cannot aggregate stuff in x.
* If a leaf node runs multiple times, each output will be given a unique id
* A leaf node which returns no state is ignored
* If run ends with only a single leaf step returning state, that state will be the run output. So the output will be like `{ data: {} }`
* If a run end with multiple leaf steps returning state, each step will report its own output (this includes one step returning multiple outputs). So the output will be like `{ a: { data: {}, b: { data: {}, 'b-1': { data: {} } }`

### DIFF
--- a/integration-tests/cli/test/errors.test.ts
+++ b/integration-tests/cli/test/errors.test.ts
@@ -106,21 +106,6 @@ test.serial('circular workflow', async (t) => {
   t.regex(error.message[0].message, /circular dependency: b <-> a/i);
 });
 
-test.serial('multiple inputs', async (t) => {
-  const { stdout, err } = await run(
-    `openfn ${jobsPath}/multiple-inputs.json --log-json`
-  );
-  t.is(err.code, 1);
-
-  const stdlogs = extractLogs(stdout);
-
-  assertLog(t, stdlogs, /Error validating execution plan/i);
-  assertLog(t, stdlogs, /Workflow failed/i);
-
-  const error = stdlogs.find((l) => l.message[0].name === 'ValidationError');
-  t.regex(error.message[0].message, /multiple dependencies detected for: c/i);
-});
-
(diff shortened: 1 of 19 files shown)


mtuchi:
@josephjclark my initial test passed 👍🏽, i want to test important behaviors that you described in depth 
i will share more feedback later today

PR #705, opened 2024-05-28
Run prettier

josephjclark:
This PR runs prettier (2.8.8) on all source files.


josephjclark:
And this looks fine to me. Assuming CI passes it can merge right into main (no need for a release)

PR #707, opened 2024-05-28
Prettier tooling

### LINKED ISSUE #671: Ensure prettier is automatically set up for new developers
It's really important that prettier runs so that we have a consistent code style.

But new developers coming in to contribute to the repo may not have prettier configured.

My expectation is:
* New users clone the repo
* They run pnpm install
* They edit a source file
* Sometime between saving and commiting, prettier runs automatically and reformats the edit. Ideally on save, so that they can see it.

We may need to add documentation to the readme, but ideally this stuff should happen automatically.

I have my VSC setup to format on save, which uses prettier. But this won't work for other editing environments.

We should look at how prettier recommends this to be setup on their officlal pages and be guided by that.

I would prefer not to format pre-commit. The idea of changing code, invisibly, right before commit it to the repo, makes my blood run cold.

josephjclark:
Add prettier tooling

Closes #671

Running a few final tests here before merging to main

No release needed.


josephjclark:
Formatting failed in CI here: https://github.com/OpenFn/kit/pull/707/commits/8b78b275726674d09cc2e20ec3d7795ffd82f7ea

josephjclark:
Hi @River-unknown, your work has been integrated into our pipeline and merged. Thanks!

River-unknown:
@josephjclark thanks to you too. I wouldn't have completed the task without that last push. Really thankful.

PR #715, opened 2024-06-13
CLI: allow file paths for state and data in a workflow

### LINKED ISSUE #639: CLI: cannot set a file path to data in workflow.json
Looks like it just isn't supported in the code. We can load expressions and config but not data. Silly really.

Example use-case:
```
{
    "workflow": {
        "steps": [
            {
                "id": "mapping",
                "adaptor": "common",
                "data": "./sampleData/data.json",
                "expression": "fn(s => ({...s, firstname: s.data.name.split(' ')})"
            }
        ]
    }
}
```

SatyamMattoo:
## Short Description

Allows the user to add file paths to the `state` and `data` in a workflow. 

## Related issue

Fixes #639 

## Implementation Details

I have added a function that checks if the `state` is a file path. If it is, the function reads the file. Additionally, if the `data` (or any key inside `state`) is a file path, it will read that file and update the key with the file's content.

For example, if we have
`workflow.json`:

```
{
    "workflow": {
        "steps": [
            {
                "id": "1",
                "adaptor": "common",
                "state": "./state.json",
                "expression": "./expression.js"
            }
        ]
    }
}
```
`state.json`:

```
{
    "data":"./data.json"
}
```
`data.json`:

```
{
    "x": 1
}
```

It will compile the workflow as:

```
{
  "workflow": {
    "steps": [
      {
        "id": "1",
        "adaptor": "@openfn/language-common",
        "state": {
          "data": {
            "x": 1
          }
        },
        "expression": "import { fn } from \"@openfn/language-common\";\nexport * from \"@openfn/language-common\";\nexport default [fn((state) => state)];"
      }
    ],
    "name": "input"
  },
  "options": {}
}
```
## QA Notes

Added a test to verify that the file paths are identified, and the file content is successfully copied to the required key. Also tested it locally for different cases.

- [x] I have added unit tests

### DIFF
--- a/packages/cli/src/util/load-plan.ts
+++ b/packages/cli/src/util/load-plan.ts
@@ -19,7 +19,6 @@ const loadPlan = async (
     | 'adaptors'
     | 'baseDir'
     | 'expandAdaptors'
-    | 'statePath'
   >,
   logger: Logger
 ): Promise<ExecutionPlan> => {
@@ -224,7 +223,6 @@ const importExpressions = async (
       );
       job.configuration = JSON.parse(configString!);
     }
-
     if (stateStr && isPath(stateStr)) {
       const stateString = await fetchFile(
         job.id || `${idx}`,
--- a/packages/cli/src/util/load-plan.ts
+++ b/packages/cli/src/util/load-plan.ts
@@ -19,6 +19,7 @@ const loadPlan = async (
     | 'adaptors'
     | 'baseDir'
     | 'expandAdaptors'
+    | 'statePath'
   >,
   logger: Logger
 ): Promise<ExecutionPlan> => {
@@ -203,6 +204,9 @@ const importExpressions = async (
       typeof job.expression === 'string' && job.expression?.trim();
     const configurationStr =
       typeof job.configuration === 'string' && job.configuration?.trim();
+    const stateStr = 
+      typeof job.state === 'string' && job.state?.trim();
+
     if (expressionStr && isPath(expressionStr)) {
       job.expression = await fetchFile(
         job.id || `${idx}`,
@@ -219,6 +223,27 @@ const importExpressions = async (
         log
       );
       job.configuration = JSON.parse(configString!);
+    }
+
+    if (stateStr && isPath(stateStr)) {
+      const stateString = await fetchFile(
+        job.id || `${idx}`,
+        rootDir,
+        stateStr,
+        log
+      );
+      job.state = JSON.parse(stateString!);
+    }
+
+    const state = job.state
+    if (typeof state === 'object' && state !== null) {
+      const keys = Object.keys(state);
+      for (const key of keys) {
+        if (typeof state[key] === 'string' && isPath(state[key])) {
+          const fileContent = await fetchFile(job.id || `${idx}`, rootDir, state[key], log);
+          state[key] = JSON.parse(fileContent);
+        }
+      }
     }
   }
 };
--- a/packages/cli/src/util/load-plan.ts
+++ b/packages/cli/src/util/load-plan.ts
@@ -232,17 +232,6 @@ const importExpressions = async (
       );
       job.state = JSON.parse(stateString!);
     }
-
-    const state = job.state
-    if (typeof state === 'object' && state !== null) {
-      const keys = Object.keys(state);
-      for (const key of keys) {
-        if (typeof state[key] === 'string' && isPath(state[key])) {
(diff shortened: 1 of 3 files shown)


josephjclark:
Thank you @SatyamMattoo  - I'll take a look at this next week :pray:

packages/cli/src/util/load-plan.ts
```diff
@@ -220,6 +223,26 @@ const importExpressions = async (
...
+      );
+      job.state = JSON.parse(stateString!);
+    }
+
+    const state = job.state
+    if (typeof state === 'object' && state !== null) {
+      const keys = Object.keys(state);
+      for (const key of keys) {
+        if (typeof state[key] === 'string' && isPath(state[key])) {
+          const fileContent = await fetchFile(job.id || `${idx}`, rootDir, state[key], log);
```
josephjclark, line 241: @SatyamMattoo What is the use-case for loading nested paths in `state.data`? It seems over-complicated and a bit weird to me

packages/cli/test/util/load-plan.test.ts
```diff
@@ -48,7 +48,7 @@ test.serial('expression: load a plan from an expression.js', async (t) => {
     plan: {},
   };
 
-  const plan = await loadPlan(opts as Opts, logger);
+  const plan = await loadPlan(opts as unknown as Opts, logger);
```
josephjclark, line 51: This typing is pretty awkward. I think you can just remove the explicit casting entirely?
```
const plan = await loadPlan(opts, logger);
```

packages/cli/test/util/load-plan.test.ts
```diff
@@ -271,3 +271,35 @@ test.serial('old-workflow: load a plan from workflow path', async (t) => {
     expression: 'x()',
   });
 });
+
+test.serial('step: allow file paths for state and data', async (t) => {
```
josephjclark, line 275: For the record this test is testing two things at once: loading `state` as a file path and loading `state.data` as a file path.

That really ought to be two different tests.

Good unit tests are as localised and focused as possible

packages/cli/src/util/load-plan.ts
SatyamMattoo: Hey @josephjclark, this was for loading file paths available within the state (like `state.data`). The issue asked for loading paths for `data` as well, so if `state.data` or any key of the `state` is a path it will load that file for us.

josephjclark: Hmm, I don't think the issue asks for that at all :thinking: 

The issue just asks to load `data` as a single path. That's wrong, as you've pointed out, it's actually the `state` key that needs loading as single path. But it says nothing about recusing through the data to load additional files. That's not a behaviour I think we need.

josephjclark:
This will also need testing and maybe implementing in the `deploy` command @SatyamMattoo, which is probably the hard bit

packages/cli/src/util/load-plan.ts
SatyamMattoo: I think the code might be misleading, what it actually does is that if the `state.data` is a file path like `/tmp/data.json` it will load that file for us. This function will load the data (or any key inside the state) for us. 
Are you suggesting we only load paths for the `state` and not for `data` inside the state? If so, I will remove this.

For further clarification this is what the function does:
If any key like data inside the state.json is a file path like:

```
{
    "data": "./data.json"
}
```
And the `data.json`:

{
    "x": 1
}

It will load the content from the `data.json` for us as well.

SatyamMattoo:
Okay sure, I will try doing that as well and push the changes for the review. For the testing part do you mean the integration tests or adding more tests here?

josephjclark:
Actually @SatyamMattoo forget about the deploy thing - I think that's actually a separate concern. I'll merge this and raise another issue for deploy if I need to (but on reflection we may not)

packages/cli/src/util/load-plan.ts
josephjclark: > Are you suggesting we only load paths for the state and not for data inside the state? 

Yes :)

Forget about data - that's a mistake in the original issue. A typo basically. The only key we care about is `state`.

If `step.state` is as string, try to load it as a file and set the resulting JSON to the value of `step.state`.

SatyamMattoo: Okay thank you! I have pushed the changes.

josephjclark (approved):

PR #722, opened 2024-06-25
Support promises

### LINKED ISSUE #721: compiler: support `.then()` and `.catch()`
Tricky one this.

Every operation is a promise*. So we should be able to do this in our job code:

```
fn(s => s).then(s)
```

This pattern eliminates the need for callbacks. Which is great, because operations with an optional options object and optional callback (which happens a lot) have really ugly signatures.

It also enables this:
```
fn(s => s).catch(s)
```
Which is badass! Suddenly I can catch any error I want! Which is covered by https://github.com/OpenFn/adaptors/issues/496 (an issue we have with this is that really I want to pass state and error into the catch, like `catch( (e, s) => {})` - but I'm not quite sure how to do that technically. Maybe a custom promise?)

The PROBLEM is that the compiler hates this.

The compiler wants to move all top-level operations into an array, like this:

```
export default [fn(s => s).then(s)]
```
First off I think this breaks everything in the runtime, because that function must be deferred. The .then() will immediately invoke it. So I think the compiler needs to do this?

```
export default [() => fn(s => s).then(s)]
```

I don't know, I need to wrap my brain around that.

But secondly, and what this issue is about, is the compiler doesn't recognise `x.then()` as an operation. It recognises it as a method call and ignores it.

I think basically the compiler needs to treat `x().then()` and `x().catch()` specially and do some kind of magic wrapping around them to make them work properly.

* Every job compiles to an array of functions, where each function returns an Operation (a function which takes state and returns state and may itself be a promise). Each of those operations is actually wrapped by our own Promise. So even if an operation is fn(), which is not a promise, that will be wrapped in a promise at execution time. Hurrah!

josephjclark:
This PR enables all operations in job code to be treated like promises. 

This confers the following benefits:

* `fn.catch()` is a really useful thing to be able to do
* We can drop most callbacks from adaptor APIs

Closes #721 

Documentation in #https://github.com/OpenFn/docs/pull/518 (this docs PR should be a great non-technical explanation of what this PR actually does)

## Motivations and examples

It occurred to me a little while ago that adaptor API design would be way cleaner if we didn't need a callback option as the last argument. It gets very hard to manage optional arguments AND an optional callback.

I also think that the callback function is semantically a little difficult. Why is is there? In most cases, you can rewrite `get(www, {}, callback)` as `get(www, {}); callback()`. It's exactly the same thing.

About the only time you need the callback is if you're nested inside an each (or maybe some other operation): `each($.data, post('/upsert', $.data, (s) => s))`. The final callback here needs to be invoked for each item in the array - an `fn()` block won't do it.

But if you could treat each operation as a promise, make it thenable and catchable, you wouldn't need the callback.

You can do stuff like this:

```
each($.data, post('/upsert', (state) => state.data).then((s) => s))
```

You can also use a catch on any operation now, if for any reason you want to do your own error handling (#496):
```
get('www').catch(e => { ... })
```

## Compiler magic

It turns out that making operations into promises isn't easy.

I don't want to re-write every adaptor to return a promise. Even if I did, that wouldn't work with the runtime (which expects to execute an array of functions, not an array of promises). Plus the promise would execute immediately, whereas we want to defer execution until later.

I also don't want magic in the runtime to handle this. I think it's important that compiled code can be executed in a simple node.js app without our runtime. That keeps code portable. If the runtime is doing  magic to promisify all operations, then the expression isn't very portable.

So the solution goes in the compiler. We have to transform the users's code into something portable.

The tl;dr is, we take  a promise `fn().then(x)` and compile it into tthis:

```
_defer(fn(), p => p.then(x))`
```
Where defer is a function imported from the runtime itself, so really the compiled code is like
```
import { defer as _defer } from '@openfn/runtime';
export default [_defer(fn(), p => p.then(x))]
```

It's kinda hard to explain, I struggle with the language and this reflects in the implementation. But I'll have a go.

Some important things to understand here:
- fn() returns a function (maybe an async one) but not a promise
- The `fn()` is compiled into an array of functions, which will be passed to the runtime
- The runtime expects an array of functions to execute, not an array of promises
- Even if I wanted to change every adaptor to return a promise, that wouldn't really work with the runtime
- It is critical that the operation created by `fn()` is not executed immediately. This is the whole thing. Each operation must create a function to be executed later.
- And obviously when you do promise.then(), the promise will execute right away

My solution is to break up the chain into two parts
1) The original operation `fn()`, untouched.
2) The promise bit, `then(whatever)`. Which might actually be a chain of promises.

I do this by creating a function called `defer(operation, thenFn, catchFn)`
- `defer` is just am operation really. It returns a function that takes state and returns state.
- When `defer` is invoked, it'll call the operation `fn()` and pass the result into `Promise.resolve`
- This gives us a Promise that takes and returns state
- Then we have to trigger the promise bit

The compiler code to break up the promise into a defer function is a little gnarly. I've tested it as well as I can.

Catch also complicates things.

We compile `catch()` and `then()` a bit differently. The catch is designed to trap any exception thrown by the operation - even though the operation may not be a promise. So the defer function needs to handle this right.

If for some reason the user does `fn().catch().then()`, then we have to handle the catch function AND invoke the promise with state.

## New Integration Tests

I've added a new suite of integration tests specifically for job writing.

These just compile and run against an adaptor - no CLI or Worker in the way. So it should be a nice clean environment to write, test, debug and experiment with pure job code.

## TODO

At the time of writing I've still got a couple of things to look at:

* [x] catch needs to pass state to the callback
* [x] The defer function is actually quite long (it pretty prints at like 12 lines) and it distracts quite heavily from the job code. I don't want to minify the function because I want compiled code to be readable. So either I need a prettier/simpler function, or we need to import the defer function from a library (I don't think it's totally unreasonable to import from @openfn/runtime in job code)
* [x] Compiler stuff is scary and I want to test this so damn hard
* [x] Docs

### DIFF
--- a/.changeset/flat-emus-shout.md
+++ b/.changeset/flat-emus-shout.md
@@ -1,6 +1,0 @@
----
-'@openfn/compiler': minor
----
-
-Add promises transformer
-Don't try and import variables declared in other import statements
--- a/.changeset/flat-emus-shout.md
+++ b/.changeset/flat-emus-shout.md
@@ -1,0 +1,6 @@
+---
+'@openfn/compiler': minor
+---
+
+Add promises transformer
+Don't try and import variables declared in other import statements
--- a/.changeset/late-plants-juggle.md
+++ b/.changeset/late-plants-juggle.md
@@ -1,0 +1,5 @@
+---
+'@openfn/runtime': patch
+---
+
+Allow the linker to directly import some whitelisted packages
--- a/.changeset/late-plants-juggle.md
+++ b/.changeset/late-plants-juggle.md
@@ -1,5 +1,0 @@
----
-'@openfn/runtime': patch
----
-
-Allow the linker to directly import some whitelisted packages
--- a/integration-tests/execute/CHANGELOG.md
+++ b/integration-tests/execute/CHANGELOG.md
@@ -1,0 +1,10 @@
+# @openfn/integration-tests-execute
+
+## 1.0.1
+
+### Patch Changes
+
+- Updated dependencies [40fd45b]
+- Updated dependencies [40fd45b]
+  - @openfn/compiler@0.2.0
+  - @openfn/runtime@1.4.1
--- a/integration-tests/execute/package.json
+++ b/integration-tests/execute/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@openfn/integration-tests-execute",
+  "private": true,
+  "version": "1.0.0",
+  "description": "Job execution tests",
+  "author": "Open Function Group <admin@openfn.org>",
+  "license": "ISC",
+  "type": "module",
+  "scripts": {
+    "test": "pnpm ava"
+  },
+  "dependencies": {
+    "@openfn/compiler": "workspace:^",
+    "@openfn/language-common": "1.7.7",
+    "@openfn/runtime": "workspace:^",
+    "@types/node": "^18.15.13",
+    "ava": "5.3.1",
+    "date-fns": "^2.30.0",
+    "rimraf": "^3.0.2",
+    "ts-node": "10.8.1",
+    "tslib": "^2.4.0",
+    "typescript": "^5.1.6"
+  },
+  "files": [
+    "dist",
+    "README.md"
+  ],
+  "devDependencies": {
+    "@types/rimraf": "^3.0.2"
+  }
+}
--- a/integration-tests/execute/package.json
+++ b/integration-tests/execute/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@openfn/integration-tests-execute",
   "private": true,
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Job execution tests",
   "author": "Open Function Group <admin@openfn.org>",
   "license": "ISC",
--- a/integration-tests/execute/package.json
+++ b/integration-tests/execute/package.json
@@ -12,6 +12,7 @@
   "dependencies": {
     "@openfn/compiler": "workspace:^",
     "@openfn/language-common": "1.7.7",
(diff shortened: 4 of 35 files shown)


josephjclark:
Taylor :+1: to .then() and .catch() on everything

Taylor :+1: to fn() being the primary canonical means of doing a "callback", but the promise is an option

josephjclark:
One blocker on removing callbacks:

```
each($.data.ids, get('www, (s) => s))
```

`then()` will save us on v2, but v1 will not have this support

josephjclark:
Phew, ok, this needs writing up better, but I've got something working in the compiler

basically we take a promise chain `fn().then(x)` and restructure it into a special `defer(fn, x)` call. The defer call ensures taht that promise doesn't run until the right time, and state is passed through the promise properly.

It's not looking TOO bad, but a major caveat right now is that I can only handle a single `.then` or `.catch` - I can't handle a whole chain. The defer function is a bit limiting in this regard.

Still to do:
* I want the compiled `defer` function to be way shorter
* What can we do about promise chains? If we're not gonna support them, I wanna throw a compile error at leastd
* Lots of tests

josephjclark:
Hey I think we're ready to go here!

I'd really like Stu to take a look at this next week, because the compiler changes are a bit aggressive.

I'd also love @mtuchi and maybe @taylordowns2000 to have a play with openfnx against this branch and see what you think! Feedback very much apprecaited. Check the docs PR for end-user sort of documentation

mtuchi:
State is undefined in this example 👇🏽 
```
bulkQuery('SELECT Name, Email FROM Contaca').catch((error, state) => {
  console.log({ error, state });
  return state;
});
```

PR #732, opened 2024-07-16
Add snapshots option to cli pull command

midigofrank:
## Short Description

Adds `--snapshots` option to cli pull command

## Related issue

Fixes #729 

## QA Notes

- `openfnx pull {projectId}` should work as it was before
- adding `--snapshots` option appends the snapshots to the query params. i.e `?snapshots[]=1&snapshots[]=2`

## Checklist before requesting a review

- [x] I have performed a self-review of my code
- [ ] I have added unit tests
- [x] Changesets have been added (if there are production code changes)

## Release branch checklist

Delete this section if this is not a release PR.

If this IS a release branch:

- [ ] Run `pnpm changeset` from root to bump versions
- [ ] Run `pnpm install`
- [ ] Commit the new version numbers
- [ ] Run `pnpm changeset tag` to generate tags
- [ ] Push tags `git push --tags`

Tags may need updating if commits come in after the tags are first generated.

PR #734, opened 2024-07-17
Add snapshots to the url for fetching project spec

### LINKED ISSUE #729: CLI: add an option to `pull` that targets given snapshots
### User story
GitHub actions rely on `openfn pull` to fetch project state and .yaml files from the app to a GitHub repository (as it would when called in a local environment). When this command is called, the latest version of the project (workflows and settings) is fetched and saved in a directory specified in the config.json file. 

While this works for most use cases, there's a challenge when openfn pull is triggered by the Sync to GitHub feature on Lightning due to lag introduced by steps between when the Sync to GitHub event was initiated and when the GitHub action triggers the `openfn pull` command. With snapshots, we want to be able to attach a snapshot id to openfn pull so that newer changes are not deployed until the user syncs them by clicking "Sync to GitHub". 

 Check this issue to read more: https://github.com/OpenFn/lightning/issues/1827

### Details
Maybe have `openfn pull --snapshots snapshot1 snapshot2` ??

- If the `snapshots` option is included, the cli should include them in the get requests to lightning.  i.e.
For fetching the YAML:
```
get(`/api/provision/yaml?id=${project_id}&snapshots[]=${snapshot1}&snapshots[]=${snapshot2}`)
```
For fetching the project state: 

```
get(`/api/projects/${project_id}?snapshots[]=${snapshot1}&snapshots[]=${snapshot2}`)
```
- When the `snapshots` option is **NOT** included, the `pull` command works exactly as it is right now.

midigofrank:
## Short Description

This is a follow up to https://github.com/OpenFn/kit/pull/732
The cli makes 2 API requests when you run `openfn pull {projectID}` . 

This PR adds the provided snapshot ids to the URL for fetching the `project` spec

## Related issue

Fixes #729 

## Implementation Details
- I'm a little worried that the logic for building the query params has been duplicated. At the same time, I think it's okay to duplicate it because they are 2 different endpoints and they "could" require different logic.

- [ ] I have added unit tests

### DIFF
--- a/.changeset/four-wolves-tease.md
+++ b/.changeset/four-wolves-tease.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': minor
+---
+
+Add snapshot ids to the url for fetching project spec
--- a/.changeset/four-wolves-tease.md
+++ b/.changeset/four-wolves-tease.md
@@ -1,5 +1,0 @@
----
-'@openfn/cli': minor
----
-
-Add snapshot ids to the url for fetching project spec
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -1,5 +1,11 @@
 # @openfn/cli
 
+## 1.6.1
+
+### Patch Changes
+
+- 505d60b: Add snapshot ids to the url for fetching project spec
+
 ## 1.6.0
 
 ### Minor Changes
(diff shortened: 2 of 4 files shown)


josephjclark (approved):

.changeset/four-wolves-tease.md
```diff
@@ -0,0 +1,5 @@
+---
+'@openfn/cli': minor
```
josephjclark, line 2: I'd probably call this a patch actually @midigofrank  - it doesn't change the external API. Let's treat it as a fix to the previous.

I'll go ahead and do this now

Issue #571, opened 2024-01-25
Worker: Support MAX_DATACLIP_SIZE


josephjclark:
The worker should check (estimate?) the seize of a serailzed dataclip and not bother to send it if exceeds `MAX_DATACLIP_SIZE`.

It should log that it did this.

josephjclark:
The CLI could log the size of the serialized state object at the end of each step. This would be nice feedback at devtime. 

Also, in the CLI, if the state object is > 10mb (which I presume is the default `MAX_DATACLIP_SIZE`), we should console.warn it

PR #740, opened 2024-07-30
Worker: limit payload size

### LINKED ISSUE #739: worker: restrict size of emitted dataclips
Sending ~40mb json dataclips is blowing up lightning

We always planned to do something about this. It's just started being a problem.

- 10mb is probbaly the target size env var but and option controlled
- we're tallking about the size of the serialized JSON, the size on the wire
- streaming encoder?

What if the limit is blown?
- kill the run
- exit reason dataclip too large

Limit will have to apply to log messages too.

So it's a generic payload limit

logs don't need to be a kill, but run:complete does

josephjclark:
* [x] Don't send a dataclip on step:complete if the payload  is too big
* [x] Don't send a log message if the payload is too big
* [x] Send `output_dataclip_id` to lightning  if the dataclip is removed

Closes #739 #571

Log output with limit printed, log redacted and dataclip not sent:
```
RTE Memory limit: 500mb
RTE Timeout: 300s
RTE Payload limit: 10mb
VER Versions:
    ▸ worker                     1.4.1
    ▸ @openfn/language-common    1.15.1
R/T Executing 4b9cf63e-82c0-46c3-beca-ad84d8d2daab
R/T Starting step New job
R/T [linker] loading module @openfn/language-common
R/T [linker] Loading module @openfn/language-common from /tmp/openfn/worker/repo/node_modules/@openfn/language-common_1.15.1/dist/index.cjs
R/T Resolved adaptor @openfn/language-common to version 1.15.1
R/T Executing expression (1 operations)
R/T Starting operation 1
JOB (Log message redacted: exceeds 10mb memory limit)
R/T Operation 1 complete in 59ms
R/T Expression complete!
R/T Completed step New job in 346ms
R/T Final memory usage: [step 81mb] [system 234mb]
R/T Dataclip too large. This dataclip will not be sent back to lighting.
R/T Run complete with status: success
```

### DIFF
--- a/packages/ws-worker/src/util/convert-lightning-plan.ts
+++ b/packages/ws-worker/src/util/convert-lightning-plan.ts
@@ -43,6 +43,7 @@ const mapTriggerEdgeCondition = (edge: Edge) => {
 export type WorkerRunOptions = ExecuteOptions & {
   // Defaults to true - must be explicity false to stop dataclips being sent
   outputDataclips?: boolean;
+  payloadLimitMb?: number;
 };
 
 export default (
@@ -57,6 +58,12 @@ export default (
   if (run.options) {
     if (run.options.run_timeout_ms) {
       engineOpts.runTimeoutMs = run.options.run_timeout_ms;
+    }
+    if (run.options.payload_memory_limit_mb) {
+      engineOpts.payloadLimitMb = run.options.payload_memory_limit_mb;
+    }
+    if (run.options.run_memory_limit_mb) {
+      engineOpts.memoryLimitMb = run.options.run_memory_limit_mb;
     }
     if (run.options.sanitize) {
       engineOpts.sanitize = run.options.sanitize;
--- a/packages/ws-worker/src/util/convert-lightning-plan.ts
+++ b/packages/ws-worker/src/util/convert-lightning-plan.ts
@@ -54,21 +54,20 @@ export default (
 
   // But some need to get passed down into the engine's options
   const engineOpts: WorkerRunOptions = {};
-
   if (run.options) {
-    if (run.options.run_timeout_ms) {
+    if ('run_timeout_ms' in run.options) {
       engineOpts.runTimeoutMs = run.options.run_timeout_ms;
     }
-    if (run.options.payload_memory_limit_mb) {
+    if ('payload_memory_limit_mb' in run.options) {
       engineOpts.payloadLimitMb = run.options.payload_memory_limit_mb;
     }
-    if (run.options.run_memory_limit_mb) {
+    if ('run_memory_limit_mb' in run.options) {
       engineOpts.memoryLimitMb = run.options.run_memory_limit_mb;
     }
-    if (run.options.sanitize) {
+    if ('sanitize' in run.options) {
       engineOpts.sanitize = run.options.sanitize;
     }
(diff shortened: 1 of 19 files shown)


taylordowns2000 (approved):
looks great. thanks!

@stuartc , happy with this if you are. from an `ENV` perspective, we currently use `MAX_DATACLIP_SIZE_MB` to control the webhook payload size limit. In an ideal world, it would be the _same_ as this `WORKER_MAX_PAYLOAD_MEMORY_MB` or, even if it's technically _not_ the same, we'd provide some way to make it the same... i.e., to not have to managed two separate values.

If an administrator wants to limit the max dataclip size to `10mb` they should be able to control that with a single environment variable, instead of keeping two in sync.

100% happy to merge this as is and add a new ENV to the mix, but want to avoid duplication if/when possible.

stuartc:
> looks great. thanks!
> 
> @stuartc , happy with this if you are. from an `ENV` perspective, we currently use `MAX_DATACLIP_SIZE_MB` to control the webhook payload size limit. In an ideal world, it would be the _same_ as this `WORKER_MAX_PAYLOAD_MEMORY_MB` or, even if it's technically _not_ the same, we'd provide some way to make it the same... i.e., to not have to managed two separate values.
> 
> If an administrator wants to limit the max dataclip size to `10mb` they should be able to control that with a single environment variable, instead of keeping two in sync.

I hear you, I'm hesitant to break the pattern we have - but what we can do on Lightning - is have the env `WORKER_MAX_PAYLOAD_MEMORY_MB` be set to the same as `MAX_DATACLIP_SIZE_MB` if the former isn't provided. This only applies if the RTM is running via Lightning. So an administrator would need to set `WORKER_MAX_PAYLOAD_MEMORY_MB` on the worker container.

But the worker has the a `payloadLimitMb` option for the run exposed now, so once we start sending that to the worker, the worker arguably doesn't need an env variable since Lightning will set it per-run. And in that case I'm happy to default to `MAX_DATACLIP_SIZE_MB` for the run options until we have per-run options.

@josephjclark what is the scale of the env variable? Does `...=10` mean 10 megabytes. Also doesn't matter too much, but is it Megabit or MegaByte?

packages/ws-worker/src/util/convert-lightning-plan.ts
```diff
@@ -53,15 +54,20 @@ export default (
...
-
   if (run.options) {
-    if (run.options.run_timeout_ms) {
+    if ('run_timeout_ms' in run.options) {
       engineOpts.runTimeoutMs = run.options.run_timeout_ms;
     }
-    if (run.options.sanitize) {
+    if ('payload_memory_limit_mb' in run.options) {
+      engineOpts.payloadLimitMb = run.options.payload_memory_limit_mb;
+    }
```
stuartc, line 63: What's the naming thought here? It's not really a memory limit right? What about `dataclip_limit_mb` or something like that?

josephjclark: Good catch, there are too many words here! I'll fix it shortly 

I'm preferring payload to dataclip because it affects logs too

josephjclark:
The env var has been renamed to `WORKER_MAX_PAYLOAD_MB`

That's measured in bytes @stuartc

PR #742, opened 2024-08-09
Support file path in job body

### LINKED ISSUE #285: Support paths to job files in project.yaml
For users who work locally or for certain projects in their GitHub repos, users want to keep their jobs in different .js files and then have the files referenced in their project.yaml files. When OpenFn deploy is called, we need to make sure we get the content of the files and add them to the right section on the project.yaml. 
We should advise users to create a job.js file for every job such that for a workflow with 5 steps, we will need 5 jobs.js files.  that are referenced in the .yaml file.

  - When using `cli deploy` and detecting a file path in the yaml, you merge in the real content before sending it to the lightning server
  - When using `cli pull` you get a new copy of the `project.yaml` and the `projectState.json` and compare it to the local copy of the `project.yaml`... you notice that the local `project.yaml` has file paths, and instead of committing the entirely new `project.yaml`  to GitHub, you interpret the paths and write to those files.
  - If someone adds a new workflow or new steps to lightning directly, then syncs to GitHub for a project that already has relative paths for other workflows or jobs, they will retain the relative paths for things that are already set up locally, but the new workflow or job will be returned inline. (they can move it out later by hand if they want.)
  - on clicking export, keep it as a monolith _for now_. (out of scope to make export use _either_ relative paths or monolith.)
  
### Implementation Note

Instead of putting a path directly in the `body` key (which would also be a string value), we use this pattern instead:

```yaml
  job-xyz:
    ...
    body:
      path: ../my-job.js
```
Note that this path is **relative** to the project.yaml, and not the current working directory.

This way we can tell that the body is _not_ available directly under the key much more easily, and leaves room for other features.

For instance, a step that looks like this in the project.yaml
```
 Notify-CHW-upload-successful:
       name: Notify-CHW-upload-successful
       adaptor: @openfn/language-http@latest
       enabled: true
       credential:
       globals:
       body: |
         fn(state => state);
``` 
We'd expect users to configure their step this way and cli deploy still works.
```
 Notify-CHW-upload-successful:
       name: Notify-CHW-upload-successful
       adaptor: @openfn/language-http@latest
       enabled: true
       credential:
       globals:
       body: 
         path: ../my-job.js
```

This does not apply to exports at the moment. All project exports will be monolithic i.e. job code and other project settings are kept in the same file.

### User acceptance criteria
- [ ] If the `body` is an object, and doesn't have a `path` key then we throw an error.
- [ ] Users can easily make a change to a job expression, and redeploy without errors if everything is structurally correct
- [ ] Running cli deploy for a project with relative file paths in the project.yaml should copy and paste the job code in the project.yaml when deploying to lightning. 
- [ ] Running cli pull for a project on lightning whose local project.yaml file uses a relative path should automatically copy and paste the job code into the right files without errors.  
- [ ] The CLI pull/deploy criteria above should also apply to GitHub sync.

midigofrank:
## Short Description

Support file paths in job body

## Related issue

Fixes #285 

## Implementation Details
When the `path` is present in the body, we load the file's content into another key `content`. So essentially, it becomes:
`{ path: './bodyPath.js', content: 'content fron bodyPath.js' }`

The `YAML.vistor` function does all the magic here. 
While traversing the nodes, we check if the current node is the `job body`. We know that it's the job body by moving up the `path tree` until we find a node with the key `jobs`. Here's how the path tree looks like:
```bash
[
  Document {
    ....
    contents: YAMLMap { items: [Array], range: [Array] },
    range: [ 0, 972, 972 ]
  },
  YAMLMap { items: [ [Pair], [Pair] ], range: [ 0, 972, 972 ] },
  Pair {
    key: Scalar {
      value: 'workflows',
      ...
    },
    value: YAMLMap { items: [Array], range: [Array] }
  },
  YAMLMap { items: [ [Pair] ], range: [ 74, 972, 972 ] },
  Pair {
    key: Scalar {
      value: 'DHIS2-to-Sheets',
      ...
    },
    value: YAMLMap { items: [Array], range: [Array] }
  },
  YAMLMap {
    items: [ [Pair], [Pair], [Pair], [Pair] ],
    range: [ 95, 972, 972 ]
  },
  Pair {
    key: Scalar {
      value: 'jobs',
      ...
    },
    value: YAMLMap { items: [Array], range: [Array] }
  },
  YAMLMap { items: [ [Pair], [Pair] ], range: [ 133, 526, 526 ] },
  Pair {
    key: Scalar {
      value: 'Upload-to-Google-Sheet',
      ...
    },
    value: YAMLMap { items: [Array], range: [Array] }
  },
  YAMLMap {
    items: [ [Pair], [Pair], [Pair] ],
    range: [ 363, 526, 526 ]
  }
]
```

When you `pull`, we merge the `StateJob.id` into the `SpecJob.id` of the current state and spec in place. We then keep track of all the jobs in a single place. The `id` here is guaranteed to be unique.
These "old spec jobs" are what we use to match the "new spec job". 

## QA Notes

- [x] I have added unit tests

### DIFF
--- a/packages/deploy/src/index.ts
+++ b/packages/deploy/src/index.ts
@@ -1,8 +1,8 @@
 import { confirm } from '@inquirer/prompts';
 import { inspect } from 'node:util';
-import { DeployConfig, ProjectState } from './types';
+import { DeployConfig, ProjectState, SpecJob } from './types';
 import { readFile, writeFile } from 'fs/promises';
-import { parseAndValidate } from './validator';
+import { parseAndValidate, addSpecJobBodyPath } from './validator';
 import jsondiff from 'json-diff';
 import {
   mergeProjectPayloadIntoState,
@@ -91,7 +91,7 @@ function writeState(config: DeployConfig, nextState: {}): Promise<void> {
 export async function getSpec(path: string) {
   try {
     const body = await readFile(path, 'utf8');
-    return parseAndValidate(body);
+    return await parseAndValidate(body, path);
   } catch (error: any) {
     if (error.code === 'ENOENT') {
       throw new DeployError(`File not found: ${path}`, 'SPEC_ERROR');
@@ -102,6 +102,55 @@ export async function getSpec(path: string) {
 }
 
 // =============================================
+
+async function getAllSpecJobs(
+  config: DeployConfig,
+  logger: Logger
+): Promise<SpecJob[]> {
+  const jobs: SpecJob[] = [];
+
+  try {
+    const [state, spec] = await Promise.all([
+      getState(config.statePath),
+      getSpec(config.specPath),
+    ]);
+
+    for (const [workflowKey, workflow] of Object.entries(spec.doc.workflows)) {
+      if (workflow.jobs) {
+        for (const [jobKey, specJob] of Object.entries(workflow.jobs)) {
+          const stateJob = state.workflows[workflowKey]?.jobs[jobKey];
+          stateJob &&
+            jobs.push({
+              id: stateJob.id,
+              name: specJob.name,
+              adaptor: specJob.adaptor,
+              body: specJob.body,
+            });
+        }
+      }
+    }
+  } catch (error: any) {
+    logger.debug(`Could not read the spec and state: ${error.message}`);
+  }
+
+  return jobs;
+}
+
+export async function updatePulledSpecJobBodyPath(
+  newSpecBody: string,
+  newState: ProjectState,
+  config: DeployConfig,
+  logger: Logger
+): Promise<string> {
+  try {
+    const oldSpecJobs = await getAllSpecJobs(config, logger);
+
+    return await addSpecJobBodyPath(newSpecBody, newState, oldSpecJobs, config);
+  } catch (error: any) {
+    logger.warn(`Could not update spec job body paths: ${error.message}`);
+    return newSpecBody;
+  }
+}
 
 export async function deploy(config: DeployConfig, logger: Logger) {
   const [state, spec] = await Promise.all([
--- a/packages/deploy/src/index.ts
+++ b/packages/deploy/src/index.ts
@@ -1,8 +1,8 @@
 import { confirm } from '@inquirer/prompts';
 import { inspect } from 'node:util';
-import { DeployConfig, ProjectState, SpecJob } from './types';
+import { DeployConfig, ProjectState } from './types';
 import { readFile, writeFile } from 'fs/promises';
-import { parseAndValidate, addSpecJobBodyPath } from './validator';
+import { parseAndValidate } from './validator';
 import jsondiff from 'json-diff';
 import {
   mergeProjectPayloadIntoState,
@@ -10,6 +10,7 @@ import {
   toProjectPayload,
   getStateFromProjectPayload,
 } from './stateTransform';
+import { syncRemoteSpec } from './pull';
 import { deployProject, getProject } from './client';
 import { DeployError } from './deployError';
 import { Logger } from '@openfn/logger';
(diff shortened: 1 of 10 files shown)


taylordowns2000 (changes requested):
looks great @midigofrank , can you:

1. update `test/fixtures.ts` to include relative paths? (and ensure the tests work properly)
2. edit https://github.com/OpenFn/docs/blob/main/docs/deploy/portability.md?plain=1#L31 to show the new format?

don't worry about shifting the current spec to "old versions"—i'll clean that up myself once your PR is open on docs

midigofrank:
> looks great @midigofrank , can you:
> 
> 1. update `test/fixtures.ts` to include relative paths? (and ensure the tests work properly)
> 2. edit https://github.com/OpenFn/docs/blob/main/docs/deploy/portability.md?plain=1#L31 to show the new format?
> 
> don't worry about shifting the current spec to "old versions"—i'll clean that up myself once your PR is open on docs

@taylordowns2000 the `test/fixtures.ts` are only used in testing the `stateTransform` functionality. State transform happens waay after the relative paths have been used to load the body. 

What I can test however is that if the body is in the form:
```yaml
body: 
  content: "some content"
```

Then the tests would still pass

packages/deploy/test/validator.test.ts
```diff
@@ -1,27 +1,39 @@
 import test from 'ava';
 import { parseAndValidate } from '../src/validator';
+import fs from 'fs/promises';
+import path from 'path';
+
+// Helper to create and clean up temporary test files
+const createTempFile = async (t, content: string, ext = 'txt') => {
```
josephjclark, line 7: hmm. There is a very clean file system mock in use in this repo which would make these tests a lot nicer IMO.

We use [mock-fs](https://www.npmjs.com/package/mock-fs)

It can get a bit hairy when it blocks out node modules, but that can be worked around. See the `mockFs` helper in `cli/test/util.ts` (you can't import that into `deploy` but feel free to duplicate it )

There's some simple example usage in `runtime/test/repo/ensure-repo.test.ts`

Not neccessary if you're struggling for time, but strongly recommended

packages/deploy/src/validator.ts
```diff
@@ -8,12 +10,16 @@ export interface Error {
   range?: [number, number, number];
 }
 
-export function parseAndValidate(input: string): {
+export async function parseAndValidate(
+  input: string,
+  specPath: string
```
josephjclark, line 15: I think you should default `specPath` to `'.'` - that way you don't break unit tests which don't use it

packages/deploy/src/index.ts
```diff
@@ -103,6 +103,55 @@ export async function getSpec(path: string) {
...
+      }
+    }
+  } catch (error: any) {
+    logger.debug(`Could not read the spec and state: ${error.message}`);
+  }
+
+  return jobs;
+}
+
+export async function updatePulledSpecJobBodyPath(
```
josephjclark, line 139: Can I ask for a better function name here? And maybe on `addSpecJobBodyPath` ?

This is really complex stuff and it's taken me ages to get any kind of feeling for what's happening here :sweat_smile:  

Basically what these functions are doing is pulling out job bodies from the newly pulled spec and writing them to disk if the state (?) includes a path. Right?

So something like `extractJobsToDisk`  might be a better name for `addSpecJobBodyPath`

Probably `updatePulledSpecJobBodyPath` should be moved to a new file `packages/deploy/src/pull.ts`, since it only affects pulls the context helps understandability. A name like `syncRemoteSpec` or something might be a good general name for what this is trying to do

josephjclark (commented):
Hi @midigofrank, sorry I've been so late getting back to you on this. A few comments from me!

I'm getting more and more worried about complexity in `deploy`. If we have time I'd love to see some fresh unit tests on the new functions.

packages/deploy/test/validator.test.ts
midigofrank: Niice, I'll change it

stuartc (approved):

PR #752, opened 2024-08-27
Worker: timestamp events

### LINKED ISSUE #751: Send start and end times to lightning
After a chat with @stuartc , we're wondering if it makes sense to send timestamps along with the key start/end events for runs and steps?

Right now, we don't include timestamps in run:started, step:started, step:complete, or run:complete: https://github.com/OpenFn/kit/blob/main/packages/lexicon/lightning.d.ts#L151-L189

Instead, we need to set the here in Lightning: https://github.com/OpenFn/lightning/blob/main/lib/lightning/runs/run.ex#L163

This could be the cause for the differences we're seeing in reported durations in the worker and in lightning.

josephjclark:
## Short Description

This PR adds timestamps to `run:start`, `run:complete`, `step:start` and `step:complete`

Timestamps are generated as hr timestamps in the _engine_, then filtered up to the worker and sent on to Lightning as microsecond strings.

The API version has been bumped up as the worker-lightning contract has changed

## Related issue

Fixes #751

### DIFF
--- a/.changeset/new-meals-accept.md
+++ b/.changeset/new-meals-accept.md
@@ -1,5 +1,0 @@
----
-'@openfn/engine-multi': patch
----
-
-Add high resolution timestamps to key events
--- a/.changeset/new-meals-accept.md
+++ b/.changeset/new-meals-accept.md
@@ -1,0 +1,5 @@
+---
+'@openfn/engine-multi': patch
+---
+
+Add high resolution timestamps to key events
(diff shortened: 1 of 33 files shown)


taylordowns2000 (approved):
happy with the timestamps sent back to lightning (just viewed in `dbg` for now).

probably wait until @stuartc has implemented the lightning side (he might have change requests) before merging

taylordowns2000:
i'm testing this with the OLD lightning api on staging right now. if all good there (additive only, we'll just ignore the timestamps) then i will merge and @stuartc can propose changes if/when he wants as he's doing the lightning side

taylordowns2000:
tests all good. merging.

Issue #759, opened 2024-09-05
Worker: looking a bit fragile if execute throws


josephjclark:
The worker's top level execute function has a few places where it might potentially throw.

This would result in the worker crashing and all runs in progess getting lost.

This is obviously quite severe but it's not entirely clear what we should do. The worker has an obligation to try and finish all outstanding work and report back to lightnig.

For example: failure to join the run channel will result in an unhandled exception. At this point something is very wrong. I don't think we can report back to Lightning yet because we've not even been able to join the comms channel (or we were kicked out). So the run is probably lost at this point. We should probably kill the whole worker (but give existing runs a chance to finish)

There are probably a few other equally nefarious places where an error will result in the whole worker crashing. It's a very sensitive bit of code with a lot of asyncrhonicity and dependencies on other systems.

PR #760, opened 2024-09-05
Worker: don't blow up if a run doesn't have a start node

### LINKED ISSUE #758: Worker: bad crash if a workflow is invalid
Still investigating this

If the worker receives a bad  workflow (basically no start node), it'll blow up very early in the execute pipeline. Probably before we have good guards.

image

I'm not quite sure what the consequences of this crash are. Am investigating. Will add guard code.

The problem is occurring because Lightning is sending us "bad" jobs, likely as a result of sync issues

josephjclark:
We're seeing an error where the worker will blow up if a run does not have a start node (or specifies a start node which does not exist).

This causes the worker to immediately exit and all outstanding runs will be lost. Not a great look.

Fixes #758 

The fix is easy enough and I've added unit tests. The runtime will catch the inconsistent workflow structure and raise an exception for us - we just need the worker to not blow up before this happens.

There are a few other areas around this code where an exception will blow up the worker. I've raised an issue to track them #759  but I don't think I want to take them on here. Let's just fix the immediate issue.

### DIFF
--- a/.changeset/curvy-onions-tease.md
+++ b/.changeset/curvy-onions-tease.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Fix an issue where a run with a missing start node caused the server to crash
--- a/.changeset/curvy-onions-tease.md
+++ b/.changeset/curvy-onions-tease.md
@@ -1,5 +1,0 @@
----
-'@openfn/ws-worker': patch
----
-
-Fix an issue where a run with a missing start node caused the server to crash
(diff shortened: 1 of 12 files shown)


josephjclark:
Kinda blocked by #752 - need to know I can merge safely

taylordowns2000 (approved):
looks good. thank you. can be merged after rebase

josephjclark:
@taylordowns2000 worker 1.6.1 is built and the PR is merged

PR #785, opened 2024-09-25
Release/next

josephjclark:
Some tidyups and logging improvements


josephjclark:
@taylordowns2000 can we try worker 1.6.5 on staging? The image is built

PR #795, opened 2024-10-15
CLI: Support kafka trigger type

### LINKED ISSUE #796: CLI: Support `kafka` trigger type
### Details

Currently, only `webhook` and `cron` trigger types are supported in the `cli`. Lightning recently added ` kafka` trigger type.

This issue aims to support the newly added `kafka` type, as needed by lightning

midigofrank:
## Short Description

This pull request adds `kafka` trigger type. 

Fixes #796 

## Implementation Details

- The structure maintains the structure as it is in lightning
- There is a validation that uses regex to ensure that the host is in the form `hostname:number`

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/deploy/src/stateTransform.ts
+++ b/packages/deploy/src/stateTransform.ts
@@ -168,6 +168,7 @@ function mergeTriggers(
 
         // prefer spec, but use state if spec is missing, or default
         const trigger = {
+          id: stateTrigger!.id,
           type: pickValue(specTrigger!, stateTrigger!, 'type', 'webhook'),
           enabled: pickValue(specTrigger!, stateTrigger!, 'enabled', true),
         };
--- a/packages/deploy/src/stateTransform.ts
+++ b/packages/deploy/src/stateTransform.ts
@@ -141,7 +141,7 @@ function mergeTriggers(
     splitZip(stateTriggers, specTriggers!).map(
       ([triggerKey, stateTrigger, specTrigger]) => {
         if (specTrigger && !stateTrigger) {
-          const trigger = {
+          const trigger: any = {
             id: crypto.randomUUID(),
             ...pickKeys(specTrigger, ['type', 'enabled']),
           };
@@ -154,7 +154,7 @@ function mergeTriggers(
             trigger.kafka_configuration = {
               ...specTrigger.kafka_configuration,
               hosts: transformSpecKafkaHost(
-                specTrigger.kafka_configuration?.hosts ?? []
+                specTrigger.kafka_configuration?.hosts
               ),
             };
           }
@@ -167,7 +167,7 @@ function mergeTriggers(
         }
 
         // prefer spec, but use state if spec is missing, or default
-        const trigger = {
+        const trigger: any = {
           id: stateTrigger!.id,
           type: pickValue(specTrigger!, stateTrigger!, 'type', 'webhook'),
           enabled: pickValue(specTrigger!, stateTrigger!, 'enabled', true),
--- a/packages/deploy/src/stateTransform.ts
+++ b/packages/deploy/src/stateTransform.ts
@@ -31,9 +31,9 @@ function stringifyJobBody(body: SpecJobBody): string {
   }
 }
 
-function transformSpecKafkaHost(hosts: string[]): StateKafkaHost[] {
+function transformSpecKafkaHost(hosts: string[] = []): StateKafkaHost[] {
   function isValidHost(value: string): boolean {
-    const regex = /^([a-zA-Z0-9.-]+):(\d+)$/;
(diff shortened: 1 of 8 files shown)


packages/deploy/src/stateTransform.ts
```diff
@@ -130,6 +149,16 @@ function mergeTriggers(
               ...(specTrigger.type === 'cron'
                 ? { cron_expression: specTrigger.cron_expression }
                 : {}),
+              ...(specTrigger.type === 'kafka'
```
josephjclark, line 152: This is getting way to complicated and I'm going to insist on a refactor here :(

I've just scribbled this and haven't tested it, but something along these lines would be a lot more readable and scalable:
```
 if (specTrigger && !stateTrigger) {
    const trigger = {
      id: crypto.randomUUID(),
      ...pickKeys(specTrigger, ['type', 'enabled']),
    };
    
    if (specTrigger.type === 'cron') {
      trigger.cron_expression = specTrigger.cron_expression
    } else if (specTrigger.type === 'kafka') {
      trigger.kafka_configuration = {
        ...specTrigger.kafka_configuration,
        hosts: transformSpecKafkaHost(
          specTrigger.kafka_configuration?.hosts ?? []
        ),
      },
    }

    return [triggerKey, trigger]
  }
```

packages/deploy/src/stateTransform.ts
```diff
@@ -149,6 +178,16 @@ function mergeTriggers(
               ...(specTrigger!.type === 'cron'
                 ? { cron_expression: specTrigger!.cron_expression }
                 : {}),
+              ...(specTrigger!.type === 'kafka'
```
josephjclark, line 181: Same again - this block needs to be refactored for readability

packages/deploy/src/stateTransform.ts
```diff
@@ -149,6 +178,16 @@ function mergeTriggers(
...
+                ? {
+                    kafka_configuration: {
+                      ...specTrigger!.kafka_configuration,
+                      hosts: transformSpecKafkaHost(
+                        specTrigger!.kafka_configuration?.hosts ?? []
```
josephjclark, line 186: I would default the argument of `transformSpecKafkaHost` inside the function declaration, rather than passing an empty array here:

```
function transformSpecKafkaHost(hosts: string[] = []): StateKafkaHost[] { ... }
```

josephjclark (changes requested):
Hi @midigofrank, I'm recommending a couple of refactors because some of the code is getting stretched very thin here.

Is there anyway we can add a test on this? We should be able to at least add unit tests on `transformSpecKafkaHost`

packages/deploy/src/stateTransform.ts
```diff
@@ -30,6 +31,24 @@ function stringifyJobBody(body: SpecJobBody): string {
   }
 }
 
+function transformSpecKafkaHost(hosts: string[]): StateKafkaHost[] {
+  function isValidHost(value: string): boolean {
+    const regex = /^([a-zA-Z0-9.-]+):(\d+)$/;
```
josephjclark, line 36: I'm a little confused by `.` and `-` at the end of the host regex here? 

`.` means any character, so it kind of means the first pad means "match any character" and renders the rest obsolete.

I'm not sure why we need to support `-`

josephjclark: Another pattern  I see for this sort of thing is to use the [`URL`](https://developer.mozilla.org/en-US/docs/Web/API/URL) object to parse a url string, and then test for `host` and `port` key.

I'm not convinced this approach is better than regex, but it's definitely better than a _bad_ regex :sweat_smile:

josephjclark: Huh and maybe a dumb question - I'm nitpicking and anyway I don't know much about kafka... but does it _definitely_  need a port in the URL? It can't run off a default port?

midigofrank: Yeah, it is AI generated 😅 . I did a couple of tests and it was able to capture all the bad cases I could think of.

I tried using the `URL` object but it proved troublesome because the protocol is missing from the required input.  `new URL('localhost:9092')` parses `localhost` as the `protocol`.  And yes, you MUST specify the port when connecting to a kafka cluster

midigofrank: Let me see if I can simplify this. Maybe we don't even need a regex, the intention is to make sure that there is a `hostname:port_number` in the string. AI decided to take it a notch higher by trying to ensure that the hostname is a valid ip address

josephjclark (approved):

Issue #777, opened 2024-09-25
Support multiple adaptors in CLI and worker


josephjclark:
The Collections API requires us to support multiple adaptors in the Worker (because we use an adaptor to provide the collections API).

The runtime and compiler should support multiple adaptors, they don't much care. And I think the CLI right now is cool with it.

Worker and Engine probably just assume a single adaptor.

So what I think we need to do is:
* Lightning Runs can include an array of adaptors (although they won't, apart from unit tests)
* ExecutionPlans should use an array of adaptors for each step (even if it's an array of 1)
* Ensure that this all works OK
* Ensure that the compiler can correctly generate imports with multiple adaptors (this could be a gotcha - didn't we switch off "smart imports"?)
* Create multi-adaptor integration tests in CLI and Worker

josephjclark:
Done in #801

PR #797, opened 2024-10-18
Support multiple adaptors in the compiler

josephjclark:
## Short Description

This PR enables the compiler to generate import statements for two adaptors.

This basically just means turning the `adaptor` option (which days which adaptor to use) into `adaptors` and making it an array. This turns out  to have a huge ripple effect in the typings but it's sorted now: the "entry points" tend to ensure that `adaptors` is an array, massing the original input in to the right format.

One issue which arises out of this work is the possibility of conflict. If you import `http` and `fhir`, for example, do we import `fn` from `http` or `fhir`? Fortunately with collections we don't really need to address this issue. But if somehow it ever does come up, it'll result in  compiled javascript code being invalid. So we can deal with this later.

This is part of #777

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/neat-lies-begin.md
+++ b/.changeset/neat-lies-begin.md
@@ -1,0 +1,5 @@
+---
+'@openfn/engine-multi': minor
+---
+
+Support multiple adaptors in job structures
--- a/.changeset/sharp-needles-peel.md
+++ b/.changeset/sharp-needles-peel.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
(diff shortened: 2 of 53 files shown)


josephjclark:
Ah ok - we need to add support for multiple adaptors in the engine as part of this PR, because the engine's call into the compiler is now wrong

josephjclark:
The size of the diff here is really escalating! The work is fairly trivial but far reaching and requires close thought. Hopefully get it wrapped on Monday

josephjclark:
Phew! I think this is done. Changed my strategy on how to do this several times :blush:

PR #798, opened 2024-10-18
compiler: don't extract common exports

josephjclark:
## Short Description

There's a very gnarly bit of compiler code called `preloadAdaptorExports` which tries to work out the list of valid exports from an adaptor.

We've had a lot of trouble around this historically - it doesn't actually work terrible well, and relies on some ropey, poorly tested code in `describe-adaptor`

The problem right now is that the list of exports for an adaptor is very generous. We basically list every export listed in the d.ts (regardless of how or where it's exported). And for some reason we also list everything exported from common, regardless of whether it's actually exported from the adaptor. Frankly I forget what this is all about.

Now this is a problem when we have multiple adaptors, a la collections, because the compiler things that both `http` and collections export `fn`, so it'll try to import it from both packages. Resulting in all kinds of trouble.

My solution is the lowest effort, lowest risk thing I can see:  only work out the exports for `index.d.ts` and `adaptor.ts`, because those two files are all that really matter for job code.

Now we're cheating a bit on the adaptors.d.ts thing, it's not a generic solution and makes an assumption about adaptor structure. But it's a fair one. And working out the exports properly, even with typescript, is actually really hard.

One day, when we re-write `describe-package`, we'll have a much more robust and generic solution for this sort of thing. But for now, I'll settle for this.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/compiler/src/util.ts
+++ b/packages/compiler/src/util.ts
@@ -1,5 +1,5 @@
 import { readFileSync } from 'node:fs';
-import { readFile, readdir } from 'node:fs/promises';
+import { readFile } from 'node:fs/promises';
 import path from 'node:path';
 import { Project, describeDts } from '@openfn/describe-package';
 import type { Logger } from '@openfn/logger';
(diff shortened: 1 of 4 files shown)


packages/compiler/src/util.ts
```diff
@@ -24,7 +24,7 @@ export const isRelativeSpecifier = (specifier: string) =>
 // But we may relax this  later.
 export const preloadAdaptorExports = async (
   pathToModule: string,
-  useMonorepo?: boolean,
+  _useMonorepo?: boolean,
```
josephjclark, line 27: TODO I need to remove this option now

PR #801, opened 2024-10-22
Collections

### LINKED ISSUE #799: integration tests: write a test against collections
Once the first build of the collections adaptor is released, we need at a least one unit test which uses it in mock mode.

I can't do it right now because the adaptor isn't released, so it's a bit of a catch-22...

josephjclark:
Release branch for collection stuff

* https://github.com/OpenFn/kit/pull/798
* #803
* #804
* Fixes #799
* #808

### DIFF
--- a/.changeset/eleven-cougars-exist.md
+++ b/.changeset/eleven-cougars-exist.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': minor
+---
+
+Support collections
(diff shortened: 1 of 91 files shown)


josephjclark:
@stuartc I think this is done.

First job tomorrow is to try it against lightning and see how it holds up.

I don't know if you want to review anything. If you do, check the merged PRs referenced in the description - you'll go insane trying to understand this PR all in one go!

josephjclark:
Two problems have crept in today:
1) `fn` is being imported twice by the compiler. But I fixed this last week! 
2) `TypeError: import_undici2.default.client is not a constructor` when the mock is disabled

josephjclark:
Tempted to drop in a fix for #806 here too, because it'll really help testing

PR #824, opened 2024-11-14
Node20.8.1

Angeth-Herjok:
## Short Description

This should get the non-working compilers to work

Fixes #

## QA Notes

You mention that no changes available on  tsconfig.common.json now it should  "types": ["ava"] // Ensure this line is included
  },
  "include": ["**/*.ts"]
}

run pnpm install
pnpm build
pnpm ava


josephjclark:
Hi @Angeth-Herjok 

Running this locally I have node `20.8.1`, but my tests fail because they cannot load typescript types.

I do not see anything in this PR which would address the issue.

Did you test locally?

tsconfig.common.json
```diff
@@ -21,6 +18,7 @@
     "noUnusedParameters": true,
     "resolveJsonModule": true,
     "skipLibCheck": true,
-    "strict": true
+    "strict": true,
+    "types": ["ava"] // Add this line
```
josephjclark, line 22: This line doesn't feel significant. This just gives the typescript compiler a hint as to where ava type definitions are.

I don't think fixes our problem - we need to tell ava how to load typescript files

Issue #761, opened 2024-09-05
Flaky test in engine


josephjclark:
This seems to fail in CI occasionally?

`trigger workflow-log for job logs`

In `packges/engine-multi/test/integration.test.ts`

https://app.circleci.com/pipelines/github/OpenFn/kit/2492/workflows/b7831333-7977-4ea9-ab47-dd4fee9a4160/jobs/5678

doc-han:
We need a single long running parent issue for tracking all flaky tests.

doc-han:
Fixed in #832

PR #832, opened 2024-11-26
Runtime: warn when an expression doesn't return state

### LINKED ISSUE #737: runtime: crash occurs if an expression doesn't return state
This should suffice:
```
fn() => undefined)
```

The runtime blows up with a exception if no state is returned at the end. We need to be more robust.

Also, warn if a operation does not return state

doc-han:
## Short Description

This PR
1. warns when an operation doesn't return a state. Warning happens when an operation realizes that the one before it didn't return state. #737

Fixes #737 #761

#### Before
<table><thead>
  <tr>
    <th>Operation</th>
    <th>Result</th>
  </tr></thead>
<tbody>
  <tr>
    <td>
Screenshot 2024-11-26 at 2 25 35 AM</td>
    <td>
Screenshot 2024-11-26 at 2 27 03 AM</td>
  </tr>
<tr><td colspan="2">the following operations just fail without us knowing why</td></tr>
</tbody>
</table>

#### After

<table><thead>
  <tr>
    <th>Operation</th>
    <th>Result</th>
  </tr></thead>
<tbody>
  <tr>
    <td>
Screenshot 2024-11-26 at 2 25 35 AM</td>
    <td>
Screenshot 2024-11-26 at 2 27 03 AM</td>
  </tr>
<tr><td colspan="2">we warn that the previous operation didn't return a state & might cause the current to fail</td></tr>
</tbody>
</table>

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/engine-multi/test/integration.test.ts
+++ b/packages/engine-multi/test/integration.test.ts
@@ -260,7 +260,7 @@ test.serial('run without error if no state is returned', (t) => {
     ]);
 
     api.execute(plan, emptyState).on('workflow-complete', ({ state }) => {
-      t.falsy(state);
+      t.deepEqual(state, {});
 
       // Ensure there are no error logs
       const err = logger._find('error', /./);
--- a/packages/engine-multi/test/integration.test.ts
+++ b/packages/engine-multi/test/integration.test.ts
@@ -151,19 +151,20 @@ test.serial('trigger workflow-log for job logs', (t) => {
 
     let didLog = false;
 
-    api.execute(plan, emptyState).on('workflow-log', (evt) => {
-      if (evt.name === 'JOB') {
-        didLog = true;
-        t.deepEqual(evt.message, JSON.stringify(['hola']));
-        t.pass('workflow logged');
-      }
-    });
-
-    api.execute(plan, emptyState).on('workflow-complete', (evt) => {
-      t.true(didLog);
-      t.falsy(evt.state.errors);
-      done();
-    });
+    api
+      .execute(plan, emptyState)
+      .on('workflow-log', (evt) => {
+        if (evt.name === 'JOB') {
+          didLog = true;
+          t.deepEqual(evt.message, JSON.stringify(['hola']));
+          t.pass('workflow logged');
+        }
+      })
+      .on('workflow-complete', (evt) => {
+        t.true(didLog);
+        t.falsy(evt.state.errors);
+        done();
+      });
   });
 });
 
--- a/packages/engine-multi/test/integration.test.ts
+++ b/packages/engine-multi/test/integration.test.ts
@@ -261,7 +261,7 @@ test.serial('run without error if no state is returned', (t) => {
     ]);
 
     api.execute(plan, emptyState).on('workflow-complete', ({ state }) => {
-      t.deepEqual(state, {});
+      t.falsy(state);
 
       // Ensure there are no error logs
       const err = logger._find('error', /./);
--- a/packages/runtime/src/execute/expression.ts
+++ b/packages/runtime/src/execute/expression.ts
@@ -18,7 +18,11 @@ import {
 } from '../errors';
 import type { JobModule, ExecutionContext } from '../types';
 import { ModuleInfoMap } from '../modules/linker';
-import { checkAndClearNullState, nullState } from '../util/null-state';
+import {
+  clearNullState,
+  isNullState,
+  createNullState,
+} from '../util/null-state';
 
 export type ExecutionErrorWrapper = {
   state: any;
@@ -52,7 +56,7 @@ export default (
       // Create the main reducer function
       const reducer = (execute || defaultExecute)(
         ...operations.map((op, idx) =>
-          wrapOperation(op, logger, `${idx + 1}`, opts.immutableState, `${idx}`)
+          wrapOperation(op, logger, `${idx + 1}`, opts.immutableState)
         )
       );
 
@@ -101,29 +105,25 @@ export const wrapOperation = (
   fn: Operation,
   logger: Logger,
   name: string,
-  immutableState?: boolean,
-  prevName?: string
+  immutableState?: boolean
 ) => {
   return async (state: State) => {
     logger.debug(`Starting operation ${name}`);
     const start = new Date().getTime();
-    if (checkAndClearNullState(state)) {
-      logger.warn(`Operation ${name} might fail!`);
+    if (isNullState(state)) {
+      clearNullState(state);
       logger.warn(
-        `The previous operation ${prevName} didn't return a state. did you forget?`
+        `WARNING: No state was passed into operation ${name}. Did the previous operation return state?`
       );
     }
     const newState = immutableState ? clone(state) : state;
 
-    if (typeof fn !== 'function') {
-      logger.warn(`Are you sure ${name} is an operation?`);
-      logger.debug(`Operation ${name} isn't a valid operation`);
-      const duration = printDuration(new Date().getTime() - start);
-      logger.debug(`Operation ${name} skipped in ${duration}`);
-      return newState;
-    }
+    let result = await fn(newState);
 
-    const result = (await fn(newState)) || nullState();
+    if (!result) {
+      logger.debug(`Warning: operation ${name} did not return state`);
+      result = createNullState();
+    }
 
     // TODO should we warn if an operation does not return state?
     // the trick is saying WHICH operation without source mapping
--- a/packages/runtime/src/execute/expression.ts
+++ b/packages/runtime/src/execute/expression.ts
@@ -18,6 +18,7 @@ import {
 } from '../errors';
 import type { JobModule, ExecutionContext } from '../types';
 import { ModuleInfoMap } from '../modules/linker';
+import { checkAndClearNullState, nullState } from '../util/null-state';
 
 export type ExecutionErrorWrapper = {
   state: any;
@@ -51,7 +52,7 @@ export default (
       // Create the main reducer function
       const reducer = (execute || defaultExecute)(
         ...operations.map((op, idx) =>
-          wrapOperation(op, logger, `${idx + 1}`, opts.immutableState)
+          wrapOperation(op, logger, `${idx + 1}`, opts.immutableState, `${idx}`)
         )
       );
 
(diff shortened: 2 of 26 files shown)


packages/runtime/src/util/null-state.ts
doc-han, line 1: This is the special object we return from an operation that didn't return a state. 
It's the same as returning `{}` but with an invisible marker by setting a Symbol property.

Hence, by checking the symbol property, we know it's a marked empty object from an operation with no state returned.

packages/runtime/src/execute/expression.ts
```diff
@@ -100,13 +101,30 @@ export const wrapOperation = (
   fn: Operation,
   logger: Logger,
   name: string,
-  immutableState?: boolean
+  immutableState?: boolean,
+  prevName?: string
```
doc-han, line 105: @josephjclark 
This was needed to know the name of the previous operation.

So we can say. `The previous operation ${prevName} didn't return a state. did you forget?`.
should the wording include the previous name?

packages/engine-multi/test/integration.test.ts
```diff
@@ -151,19 +151,20 @@ test.serial('trigger workflow-log for job logs', (t) => {
...
+          didLog = true;
+          t.deepEqual(evt.message, JSON.stringify(['hola']));
+          t.pass('workflow logged');
+        }
+      })
+      .on('workflow-complete', (evt) => {
+        t.true(didLog);
+        t.falsy(evt.state.errors);
+        done();
+      });
```
doc-han, line 167: @josephjclark 
Is this a good fix to the flaky test?

I think so. we basically want to check whether workflow-log happens before workflow-complete. Having two different execute meant we couldn't determine the order in which the events were fired because we didn't know which one executes faster. with a single execute, it's quite deterministic.

packages/runtime/src/execute/expression.ts
```diff
@@ -100,13 +101,30 @@ export const wrapOperation = (
...
+    if (checkAndClearNullState(state)) {
+      logger.warn(`Operation ${name} might fail!`);
+      logger.warn(
+        `The previous operation ${prevName} didn't return a state. did you forget?`
+      );
+    }
     const newState = immutableState ? clone(state) : state;
-    const result = await fn(newState);
+
+    if (typeof fn !== 'function') {
```
doc-han, line 118: @josephjclark 
This function check is all we need!

1. it makes sure that fn can be called. prevent error. skipping operation
2. this also works for async functions, because, they are functions that basically return a promise.
3. this also takes care of unwanted top-level code. 🥳

#### Takes care of unwanted top-level code. 
This is because the `fn` value we're validating here isn't directly the function call at the top-level but rather what that function call returns. 

So if you have eg.
```js
map("$.[*]",
  create("SObject",
    field("FirstName", sourceValue("$.firstName"))
  )
)
```
then that `fn` check in our code is for the function returned by our top-level code `map`. 

Note: All of our top-level functions are high-order functions with some signature
```js
function map(...args){
	return (state) => {
		// do some work here
	}
}
```

Hence checking whether `fn` is a function is actually checking whether some top-level code is a high-order function. which most unwanted top-level code fail to satisfy. 
eg. 
`console.log("")`
`setTimeout()`
...

doc-han:
yet to add some additional tests!

packages/engine-multi/test/integration.test.ts
josephjclark: Hey - oh , yes! Man this test was so dumb :joy: 

Yeah you're absolutely right. This is an editing error or just a straight up fail by me.

Thank you!

packages/runtime/src/execute/expression.ts
```diff
@@ -100,13 +101,30 @@ export const wrapOperation = (
...
+      logger.warn(`Are you sure ${name} is an operation?`);
```
josephjclark, line 119: Logging on this one is tricky! :thinking: 

The "skipped" function is actually executed, right? It's not run as part of the pipeline. So we're in a weird space.

Let's not print a duration. Let's also print just one debug line and one warning.

This will get better when I've got source mapping plugged in so let's not worry about it too much

So can I suggest:
```
logger.warn(`WARNING: operation ${name} is not a valid operation! Make sure to only use functions provided by an adaptor at the top scope.`)
logger.debug(`Skipping operation {$} as invalid`)
```

This isn't very well explained by the docs so I'll add some more docs and then come back and put a link in the error later.

josephjclark: Actually I'm super tempted to throw in this case and refuse to run :thinking: 

If the user has done this, then their code is not executing like they think it is. Functions will not execute properly, or they'll run out of order.

It's relatively OK if it's just a console.log - it'll log nonsense but nothing will blow up. But it's extremely unlikely that the user will get the behaviour they want, so maybe we need to force them to remove the offending line.

Let me think about this. Maybe I'll accept the logging for this PR but I'll raise an issue to think about being stricter.

josephjclark (commented):
This is fantastic @doc-han, thank you so much! It's a really good, clean PR and you've done all the right things.

Having had a little play with this and the output, I'm having some thoughts about the structure. 

At the end of an operation, if no state was returned, let's **always** `logger.debug(`warning:  operation ${name} did not return state`). That gives us a nice log in-context, and because it's debug level we can easily hide it.

So the output will be like:
```
[R/T] ❯ Starting operation 1
[R/T] ❯ Warning: operation 1 did not return state
[R/T] ❯ Operation 1 complete in 0ms
```

Plus there are some other comments

packages/runtime/src/util/null-state.ts
```diff
@@ -0,0 +1,21 @@
+const NULL_STATE = Symbol('NullState');
+
+// The good thing about using a Symbol is that even if we forget to clean the object.
+// it's still represented as {}, because symbols aren't visible as keys
+export function nullState() {
```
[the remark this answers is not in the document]
doc-han: either works. I'll pick `createNullState` since I don't see classes.

packages/runtime/src/util/null-state.ts
```diff
@@ -0,0 +1,21 @@
+const NULL_STATE = Symbol('NullState');
+
+// The good thing about using a Symbol is that even if we forget to clean the object.
```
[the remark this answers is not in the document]
doc-han: Alright

packages/runtime/src/util/null-state.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export function isNullState(state: any) {
+  return typeof state === 'object' && state[NULL_STATE] === true;
+}
+
+export function clearNullState(state: any) {
+  if (typeof state === 'object') delete state[NULL_STATE];
+}
+
+export function checkAndClearNullState(state: any) {
```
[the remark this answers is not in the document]
doc-han: yeah that's right. I was over-writing 😂
I wanted a single function that does both checking whether an object is nullstate and also at the same time removing the symbol from the object. 

The symbol has to be removed after detection. else if it keeps getting passed down then all subsequent operations will warn on state not returned.

packages/runtime/test/execute/expression.test.ts
```diff
@@ -139,17 +140,20 @@ test.serial('async operations run in series', async (t) => {
   t.is(result.data.x, 12);
 });
 
-test.serial('jobs can return undefined', async (t) => {
-  // @ts-ignore violating the operation contract here
-  const job = [() => undefined] as Operation[];
+test.serial(
+  'jobs return null-state instead of undefined or null',
```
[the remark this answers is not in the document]
doc-han: So currently, this PR makes sure that, `null` or `undefined` aren't possible values returned by an operation. because they get replaced with a `null-state`.

When an operation is executed, 
1. we don't let null or undefined to be passed to the next operation. 
2. we replace that with a marked empty object. 
3. the next operation check whether its input state is marked and get's to know that the previous state actually didn't return state.

packages/runtime/src/execute/expression.ts
```diff
@@ -100,13 +101,30 @@ export const wrapOperation = (
...
-  immutableState?: boolean
+  immutableState?: boolean,
+  prevName?: string
 ) => {
   return async (state: State) => {
     logger.debug(`Starting operation ${name}`);
     const start = new Date().getTime();
+    if (checkAndClearNullState(state)) {
+      logger.warn(`Operation ${name} might fail!`);
+      logger.warn(
```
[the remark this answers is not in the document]
doc-han: Alright. makes sense 👍🏽

packages/runtime/src/execute/expression.ts
doc-han: Yup that's right. 
1. just as you said, we are actually supposed to error and abort. Because such failure probably will affect the subsequent operations.
2. I was then thinking, the only things that will fail here are top-level blocks of code that actually aren't imported functions from our adaptors.

doc-han:
I think most of our user errors will be reduced extremely when they have development time help(proper static & semantic validations)

packages/runtime/test/execute/expression.test.ts
```diff
@@ -139,17 +140,20 @@ test.serial('async operations run in series', async (t) => {
...
-  const context = createContext();
+    const state = createState() as TestState;
+    const context = createContext();
 
-  const result = (await execute(context, job, state, {})) as TestState;
+    const result = (await execute(context, job, state, {})) as TestState;
 
-  t.assert(result === undefined);
-});
+    t.assert(isNullState(result));
```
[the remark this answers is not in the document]
doc-han: function `isNullState` does an object check. 
```js
export function isNullState(state: any) {
  return typeof state === 'object' && state[NULL_STATE] === true;
}
```
Because nullState is also an object.

packages/runtime/src/execute/expression.ts
doc-han: I'm removing the function check out of this PR. probably needs a new one.

doc-han: will be in a different PR!

packages/runtime/test/execute/step.test.ts
```diff
@@ -285,10 +285,10 @@ test.serial('log memory usage', async (t) => {
   t.regex(memory?.message, /\d+mb(.+)\d+mb/i);
 });
 
-test.serial('warn if a non-leaf step does not return state', async (t) => {
+test.serial("warn if a previous step don't return state", async (t) => {
```
josephjclark, line 288: Hey @doc-han this test isn't quite right :thinking: 

Your edit makes this a single step with two operations in the expression.

But this test is making sure that:
a) when a step in a workflow is finished, check if it returned state
b) if it does not, log a warning

I'd forgotten that we were already doing b), but it's actually broken in this PR!

Here's some log output from production for a workflow where one step doesn't return state:

```
[R/T] ✔ Completed step a in 124ms
[R/T] ❯ Final memory usage: [step 57mb] [system 167mb]
[R/T] ⚠ WARNING: step a did not return a state object. This may cause downstream jobs to fail.
[R/T] ℹ Starting step b
```

Note that these are STEPS, not operations!

What we actually need to do is add an `isNullState` test to `execute/step.ts` line 424, to restore this statement:
```
  if (next.length && !didError && !result) {
    logger.warn(
      `WARNING: step ${stepId} did not return a state object. This may cause downstream jobs to fail.`
    );
  }
```

Then you should be able to revert the changes to this test and it should pass.

josephjclark (changes requested):
See comment!

packages/runtime/test/execute/step.test.ts
doc-han: Currently, null-state doesn't escape the expression execution context. Hence, at the step level, you'll never have it. 
due to:
https://github.com/OpenFn/kit/blob/4376f7b3e615c0b0144deecf6d65d4a03dbb3abf/packages/runtime/src/execute/step.ts#L101-L102

Hence, if the last result before step is null-state. I change that back to undefined so step get's what it expects.

packages/runtime/src/execute/step.ts
```diff
@@ -94,12 +94,13 @@ const prepareFinalState = (
...
-      `Cleaning up state. Removing keys: ${removedProps.join(', ')}`
-    );
+    if (removedProps.length)
+      logger.debug(
+        `Cleaning up state. Removing keys: ${removedProps.join(', ')}`
+      );
 
-    const cleanState = stringify(state);
-    return JSON.parse(cleanState);
+    if (isNullState(state)) return undefined;
```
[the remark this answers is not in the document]
doc-han: done.

josephjclark (approved):
Thanks! I'll do a final check and release in the morning :rocket:

Issue #617, opened 2024-03-01
Update node versions


josephjclark:
* Ensure everything works in node 18.19
* Ensure everything works in 20.latest
* Ensure everything works in 21.x

Ideally we should just work across all vaguely modern node versions.

18.19 did something I think to break the loader vm arg, which causes all sorts of problems.

Remember to update the circle config so that CI is also running in the latest version. I'd actually love to run in a matrix with 18,20,22 (at least for a little while)

- Ava tests fail
- the describe-package esbuild script fails

Some constraints on the work:
- The fix should be minimal - ideally this is just a tweak to a couple of config files
- I am hoping that we can continue to test in memory in a single command, rather than compiling tests into .js files and leaving them on disk

Akshanshkaushal:
hi @josephjclark  @christad92  To better understand the codebase, I want to solve this issue. Could you please assign me this?

josephjclark:
So the problem we have is that node 20 changes the way loaders work, which breaks our ava config. This will affect adaptors as well - see https://github.com/OpenFn/adaptors/issues/187

The first problem is that the ts node loader has changed and doesn't work with ava. It works if I use a thing called `tsimp`, but it builds all the test files to disk. Not essential I suppose but it annoys me. Also the docs say it does everything virtually, so that's kinda confusing.

The second problem is some changes about importing folders, which we can't do any more. I suppose we'll just have re-write the affected code, although that could be a big job

josephjclark:
Running ava with workerThreads off (forcing it to use child processes) works fine, at the cost of using more system resources (and so being slower in CI).

What I'm confused at with this is that I don't get import errors for importing folders and stuff, which I do get when using other loaders. Is that just a config thing?

josephjclark:
See also #542

PR #833, opened 2024-11-29
update node version (v18, v20, v22)

### LINKED ISSUE #542: node 20: ava fails in node 18.19.0
Node 18.18.2 is absolutely fine.

Node 18.19.0 breaks ava.

What's happening SOMEHOW is the ts-node esm loader, which we use to run our ava tests, breaks against the very latest node. At least for the worker-threaded approach (forcing ava to run with child processes works fine).

I think we SHOULD be using `@ava/typescript` and not `ts-node/esm` to handle the typescript tests. But `@ava/typescript` fails for me, even in a minimal testing project, and I can't work it out. See #541 for notes on that.

The build, so far as I can tell, is unaffected.

As a wider note our node version is quickly drifting out of date and we really ought to get on a 20x version (if not 21 soon). I wonder too if we should run our integration test suite against LTS 18 and 20 as well as the latest node build - just to ensure stability.

doc-han:
## Short Description

This PR ensures that **Openfn/kit** tooling works on several node versions. (v18, v20 & v22)

## Fixes
#542
#617

### 1. chore: move from ts-node to swc-loader (f857884aedd6f6b0367eef6300e3d38dcaba3f9e)
Removes `ts-node` from the codebase and adds `@swc-node/register` as the new loader for running ava tests.

### 2. ci: update CI to run tests for v18, v20 & v22 (33f777da959aeac9fc990d2fa4387e1bb2ca6cdb)
Updates the CI to run `unit_tests` and `integration_tests` under the node versions 18,20, 22.

### 3. fix: several tests & mocks (1222c9cabcbeb7f4d28ee5b3c9eab3c5c8af31e6)
This commit fixes several mocks that seem to function in v18 but not in other node versions. Some of them have to do with functionalities that don't exist across versions and so on.

1. Change implementation of mockSocket to use a class instead of a function.
2. Update the import of `package.json` from import with assert to use `fs`
3. Update wording of expected error `cannot read properties of undefined` to consider `cannot destructure property`
4. update test that waits for workerprocess to be killed to actually be killed. Because `workerProcess.kill('SIGTERM')` is async

### 4. fix: resolve mock-fs (fc2e5e12ea799d865b9628e461c90669f0d5f429)
1. `readFileSync` with `utf-8` was broken by an optimization in node. **An update to `mock-fs@5.4.1` resolved that**. (https://github.com/tschaub/mock-fs/issues/377)
2. dynamic import from a mocked filesystem stopped working with `mock-fs`. No good issue around that. **Using the real filesystem resolved that**. (https://github.com/tschaub/mock-fs/issues/130)

### 5. test: collection tests & flaky test (4cfd33913335be391df23c5c91a2b98dc5b43eb6)
1. A test issue introduced by tests accompanying collections.
2. skipping the test `"allow a job to complete after receiving a sigterm"`. This is because it's flaky and `.kill('SIGTERM')` is async.
## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


doc-han:
@josephjclark 
We have some progress here. 

1. Tests run on node20 when we're using the swc loader. 
2. Hence, we currently have 2 separate CI tests for v18 and v20
3. Most of the issues at the moment are code specific. some things that work in v18 but seem troublesome in v20. Mostly some mocks that need to be updated.

Work in progress!

josephjclark:
Thank you for the update @doc-han ! Appreciate it

If anything in the core code is seriously breaking please let me know as soon as possible

doc-han:
> Thank you for the update @doc-han ! Appreciate it
> 
> If anything in the core code is seriously breaking please let me know as soon as possible

Currently, the only thing not working in `mock-fs`. 
It had some issues due to optimizations introduced to the `fs.readfilesync function`. 

In node 20. `mock-fs` fails to provide a mock when the mocked fs path is imported instead of read of `fs`.

This is because. with custom adaptor paths. even tho we mock the adaptor path. the runtime doesn't use `fs` to read the path. we rather dynamically import he `index.js` file at that adaptor path.

Been on it for a while now.

josephjclark:
I see. So we either need to change how the runtime imports modules (not keen on that!) or we need to find a better way to mock modules on the file system?

Or we need to work out a different way to test. We could use the actual file system to load modules, although the mock is very convenient 🤔

So you have a link for mock-fs and node 20? Does this fix itself it node 22?

josephjclark:
Hey @doc-han I think this is going to get complicated, can we divide-and-conquer a bit?

- raise an issue for each isolated node 20 problem (ie node loaders, mock fs)
- create an epic branch to hold all the fixes ready for release
- create a fix branch for each specific issue, targeting the epic branch 

This will make code review much easier and help me understand what's going on.

I can help tidy up the issues on Monday, just so long as I have a clear list of the problems

doc-han:
> I see. So we either need to change how the runtime imports modules (not keen on that!) or we need to find a better way to mock modules on the file system?
> 
> Or we need to work out a different way to test. We could use the actual file system to load modules, although the mock is very convenient 🤔
> 
> Di you have a link for the mock-fs and node 20 issue? Does this fix itself it node 22?

1. readFileSync Issue: https://github.com/tschaub/mock-fs/issues/377 (fixed via a version bump to 5)

2. don't see a good issue around `mock-fs` not working with dynamic import in node v20. closest issue for `mock-fs@v3`https://github.com/tschaub/mock-fs/issues/130

3. Node has an issue where loader arguments are not passed down to workers. Hence. other loaders like tsx/esm actually work with ava but due to this you'll have to not spawn workers for ava test when using them. https://github.com/nodejs/node/issues/47747
```js
// ava.config.js
...
  workerThreads: false,
...
```
____

## Our current headache
when `mock-fs` mocks `/modules/`
https://github.com/OpenFn/kit/blob/4376f7b3e615c0b0144deecf6d65d4a03dbb3abf/packages/cli/test/commands.test.ts#L60-L72
we later want to import from this mocked filesystem.
https://github.com/OpenFn/kit/blob/4376f7b3e615c0b0144deecf6d65d4a03dbb3abf/packages/runtime/src/modules/linker.ts#L151

Works in v18 but not in v20

doc-han:
Hence, our only failing tests are in `packages/cli/test/commands.test.ts`

doc-han:
> Hey @doc-han I think this is going to get complicated, can we divide-and-conquer a bit?
> 
> * raise an issue for each isolated node 20 problem (ie node loaders, mock fs)
> * create an epic branch to hold all the fixes ready for release
> * create a fix branch for each specific issue, targeting the epic branch
> 
> This will make code review much easier and help me understand what's going on.
> 
> I can help tidy up the issues on Monday, just so long as I have a clear list of the problems

Yeah we can do that. 
My plan was to get things working and then with rebase, rearrange and squash some of the commits into meaningful bits.

josephjclark:
Ok cool: so let's make this PR solely about https://github.com/OpenFn/kit/issues/542, and any other node 20 issues we'll handle  in different PRs

packages/engine-multi/test/integration.test.ts
```diff
@@ -151,19 +151,20 @@ test.serial('trigger workflow-log for job logs', (t) => {
...
-        t.pass('workflow logged');
-      }
-    });
-
-    api.execute(plan, emptyState).on('workflow-complete', (evt) => {
-      t.true(didLog);
-      t.falsy(evt.state.errors);
-      done();
-    });
+    api
```
josephjclark, line 154: If we get the the other PR merged and released (the one about null state), we don't have to duplicate this fix here. Generally a good idea to prioritise closing down open PRs :+1:

packages/ws-worker/src/mock/sockets.ts
```diff
@@ -74,39 +74,48 @@ export const mockChannel = (
 
 type ChannelMap = Record<string, ReturnType<typeof mockChannel>>;
 
-export const mockSocket = (
-  _endpoint?: string,
-  channels?: ChannelMap,
-  connect: () => Promise<void> = async () => {}
-) => {
-  const allChannels: ChannelMap = channels || {};
+export class MockSocket {
```
josephjclark, line 77: I appreciate that this branch is a work in progress but is this a necessary change? Should it be branched out differently to the ava/loader thing?

ava.config.js
```diff
@@ -1,3 +1,6 @@
+const semver = require("semver")
+
+const loader_arg = semver.lte(process.version, 'v20.5.0') ? '--loader=ts-node/esm' : '--import=@swc-node/register/esm-register'
```
josephjclark, line 3: If we're going to migrate to swc (and generally I'm happy to) are there other uses of ts-node or other loaders that we can get rid of?

We already have too many compilers and builders. Most for agood reason, but this might be a good time and opportunity to consolidate a little

packages/ws-worker/src/mock/sockets.ts
doc-han: It was necessary given that. it was initially defined using an arrow function instead of a normal classic function which utilises `this` to grant it a constructor. 
It was then failing in v20 because there were several calls to it using the `new` operator.

ava.config.js
doc-han: Yeah, we can get rid of one. once, one(probably swc or tsx) works across the versions.

josephjclark: Any insight as to which is better? Swc looks pretty cool, but then they all look pretty cool 🤔

doc-han:
@josephjclark 
The last commit [48285ba](https://github.com/OpenFn/kit/pull/833/commits/48285baa88f06a84348a21962e806217fb476457) is cherry-pick of #836 
will be removed after #836 gets merged and this branch will be rebased.

Resolves our test issues and as you can see. tests are passing on v18 & v20. hurray 🎉 (bed-time)

doc-han:
> Are we saying there's no fix to this? We straight up can't use ava + typescript + workers in node 20+? That would be a huge disappointment - are more people talking about this somewhere?

We currently don't have a problem with that. SWC seems to work fine. That issue was happening with other loaders like `tsx`. 
You'll notice that flag isn't set in `ava.config.js`

.circleci/config.yml
```diff
@@ -124,5 +135,8 @@ workflows:
           requires:
             - build
       - integration_test:
+          matrix:
+            parameters:
+              node_version: ["18.18", "20.12"]
```
josephjclark, line 140: Can we test on: 18.12.1, 12 LTS, 20 LTS, 22

ava.config.js
josephjclark, line 3: Let's check and see if SWC works on all node versions :crossed_fingers:

doc-han:
@josephjclark 
I've cleaned up the PR to have just 5 commits. Each explaining a concise chunk of work done to get this working. 
The lock file issue was due to changing the order of commits in the PR before squashing them. I've spent some time to go back on the first commit and procedurally fix our lock-file by redoing the actions that change that file in the new order of the commits.

packages/deploy/src/stateTransform.ts
```diff
@@ -472,7 +472,8 @@ function idKeyPairs<P extends { id: string }, S extends { id: string }>(
   stateItems: Record<keyof S, S>
 ): [key: string, projectItem: P, stateItem: S][] {
   let pairs: [string, P, S][] = [];
-  for (const projectItem of projectItems) {
+  for (let i = 0; i < projectItems.length; i++) {
```
josephjclark, line 475: This is a curious diff?

Does for/of cause something to fail in node 22 here?

josephjclark:
This is looking great. Apart from one question, I'm pretty happy that:
* Changes to runtime code are minor
* Any diffs to runtime code are necessary
* Changes to the build should be safe

Basically this does a great job of fixing and updating the test framework and improving the robustness of the builds.

I will take over this PR now in order to:
* Update versions. Technically it's a patch with no interface changes. Since nothing here breaks in node 18, I don't think it's a major.  On the other hand it feels slightly risky and think emotionally it's a minor.
* Update one integration test to change date formatting (which for some reason is  failing for me locally, probably time zones)
* Set the local node version to 22
* Fix the unit tests to node 22 (no matrix)
* Update the integration test matrix to node 18.12. 18 LTS, 20 LTS and 22 latest.


I also want to run some manual tests and will report back
- Test lightning with prod worker on node 18, 20 and 22. I expect a breakage because of the import asseration change
- Re-test with this build of the worker on 18, 20 and 22
- Maybe look at the built diffs across the version dumps, see if anything major has changed in transpilation

Once I've finished all that, I'm happy to release.

Next, we need to update any node compatibility/version docs we have lying around (particularly for lightning deployments), and soon we should deprecate node 18 in the worker with 6 months notice.  I might deprecate node 18 in the CLI too, just to encourage people to update.

packages/deploy/src/stateTransform.ts
doc-han: Changed it because not all list data passed to the `idKeyPairs` function is an `iterable` (had some errors on that). might be an array but not an `iterable`. Hence the `for..of` syntax will error but our classic looping works on both `iterables` and `non-iterable` lists.

doc-han: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols

josephjclark:
Only a simple test but this is working fine in locally node 22

image

And also in node 18:

image


I've also been running some tests in CLI and I think we're looking good.

josephjclark:
Starting to see some strange fails on particular node versions. Investigating - not sure how concerned to be yet. Also seeing some strange log output in CI whici I need to get to the bottom of.

Also, I think unit tests have been running wrongly? I think forever? They should be running inside a docker container to use the global openfn command  - but they're not, they're running from a local build. This is probably something that needs to be looked at separately

doc-han:
> Starting to see some strange fails on particular node versions.

The `uncaught exception` error in unit-test is already documented under #844 
I've updated the description to elaborate on what's making it error.

josephjclark:
I've caught a couple of things:

1) Incredibly, the runtime was missing an `await` on the module loader. It's clearly working fine synchronously (somehow!), but when there's an error it's not getting caught. 
2) This is resulting in some tests appearing to pass even though they're throwing errors internally.

I've fixed the runtime, fixed some tests, and skipped some low-value ones

packages/ws-worker/src/events/run-start.ts
```diff
@@ -31,7 +31,7 @@ export default async function onRunStart(
   };
 
   const versions = {
-    worker: pkg.version,
+    worker: loadVersions().engine,
```
josephjclark, line 34: Ooh this isn't right - the worker and engine versions are different

doc-han: I had to rename it to version. It actually takes the version from package.json of the current scope

packages/ws-worker/test/events/run-start.test.ts
```diff
@@ -91,12 +91,18 @@ test('run:start should log the version number', async (t) => {
   const input = 'abc';
   const jobId = 'job-1';
 
+  // explicitly load package.json her
+  // because load-version does it a bit differently
+  const { default: pkg } = await import('../../package.json', {
+    with: { type: 'json' },
+  });
```
doc-han, line 98: this seem to only work for 20+

import with assertion doesn't seem to be stable at the moment.

josephjclark: Yes, this is fine - I'm only running the unit tests in node 22. Runtime code will be absolutely fine.

I've chosen import over readFile because it doe a better job of resolving the module relatively. `fs` was returning the wrong package.json.

Saying that, I've just noticed that everything is failing...

doc-han: I think the aim wasn't to only make the tests pass but also be sure that when someone with other node supported version runs the code it just works.

josephjclark: Not for this repo, no. Devs should be using asdf to manage versions and running on node 22.

The npm packages should all be flexible, but the repo itself should not

josephjclark:
Still some flaky tests in docgen which I'll have to keep skipping. But otherwise this is ready go. I'll test the build on the staging server.

I'm feeling pretty confident about this but the last two hours have got me a little worried. So I'l retest with a cool head tomorrow.

josephjclark:
Thank you so much for this @doc-han :pray:

Issue #648, opened 2024-04-03
Runtime: allow re-usable functions across a workflow


josephjclark:
## The Problem

It's common in workflows to want to re-use the same helper function in multiple steps.

For example, let's say we create our own id-generator or date-formatter, or data converter.

We used to allow functions to be written to state and used in multiple steps. I think that might still work? [citation needed]. It should break in-between steps really. Even if it's works its not a great approach.

### `workflow.functions`

Declare a function somewhere on the workflow body. Maybe we can have `workflow.fns` as a peer of `workflow.steps`. It probably  references a file, rather than defines functions inline.

I'm not keen on this idea because it need s special UI support in Lightning and I just don't think we have the resources.

### Make functions global

We could say: if you declare a function in a step, that function will automatically be available to the rest of the workflow.

This would either be through a function declaration `function generateId() {}` or possibly an exported function declaration (are arrows allowed?). Exported functions would open a door to unit testing too. Of course, we could use compiler magic to auto-export top-level functions.

I have no idea how we'd implement this technically.

Do we have to pre-parse the workflow, find all functions, extract their source, and set them into the runtime context? Sheesh, that sounds like a lot of work. 

Alternatively, when a function is declared, we call a special runtime API to register it, and it's then available to all other functions downstream. This would be alright.

### Use an API

We could provide a special API like `declare()`  which takes a function declaration and a name as an argument. Then, just like the global functions thing, this function would be made available to all downstream steps.

We'd have to inject the `declare` function from the runtime, just like we do with the `defer` function.

This could be compiler magic again - if the `declare` function is a useful hook for implementation, then maybe the compiler can wrap all top level function declarations into declare cals.

josephjclark:
Quick brainstorm of options:

* Allow the runtime to safely serialise functions. OR some functions. You should be able to decompile a function down to its source code, right? How does that work and can we exploit it? Then we can write functions to state and pass them downstream. This is a bit of a fiddly UX though because if you return state downstream, you could lose the functions entirely. Also I think writing logic to  a state object feels like an antipattern
* That means we have to import the custom functions into the job.This might make the compiler's life harder because it has to distinguish adaptor functions from user functions (and what if there's a conflict?)
[...]

PR #835, opened 2024-12-03
Compiler: allow function calls on lazy state

### LINKED ISSUE #684: Lazy state fails on function calls
This should work:
```
fn((state) => {
   state.callMeMaybe = () => {}
  return state
});

fn($.callMeMaybe())
fn((state) => {
   $.callMeMaybe()
)
```

But it does not - $ is undefined.

This is likely just a compiler issue. I think we should allow the usecase

doc-han:
## Short Description

Currently function calls on lazy-state do not work. 
eg. 
```js
fn((state) => {
   state.callMeMaybe = () => {}
  return state
});
// does not work before this PR
fn($.callMeMaybe())
```
This is because, we fail to wrap the state callback well. This update fixes that.
#### How it worked before and after
Given the source code `fn($.callMeMaybe())`
| Before this PR                             | After this PR                            |
|--------------------------------------------|------------------------------------------|
| ``` fn((state => state.callMeMaybe)()) ``` | ``` fn(state => state.callMeMaybe()) ``` |

Fixes #684 

#### Steps
1. after finding the MemberExpression with $, we start walking up the tree to find a CallExpression to wrap. 
2. when we find this CallExpression, a simple check is done to find out whether the parent of this CallExpression is another CallExpression with callee being state. (that's how we know it's a function called on a lazy-state)
3. we then want to wrap the state stuff at an additional level up for it to work!

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/integration-tests/execute/CHANGELOG.md
+++ b/integration-tests/execute/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/integration-tests-execute
 
+## 1.0.11
+
+### Patch Changes
+
+- Updated dependencies
+  - @openfn/compiler@0.4.2
+
 ## 1.0.10
 
 ### Patch Changes
--- a/integration-tests/execute/ava.config.js
+++ b/integration-tests/execute/ava.config.js
@@ -1,0 +1,13 @@
+export default {
+  extensions: {
+    ts: 'module',
+  },
+
+  environmentVariables: {
+    TS_NODE_TRANSPILE_ONLY: 'true',
+  },
+
+  nodeArguments: ['--loader=ts-node/esm', '--no-warnings', '--experimental-vm-modules'],
+
+  files: ['test/**/*test.ts'],
+};
--- a/integration-tests/execute/package.json
+++ b/integration-tests/execute/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@openfn/integration-tests-execute",
   "private": true,
-  "version": "1.0.10",
+  "version": "1.0.11",
   "description": "Job execution tests",
   "author": "Open Function Group <admin@openfn.org>",
   "license": "ISC",
--- a/integration-tests/execute/test/lazy-state.test.ts
+++ b/integration-tests/execute/test/lazy-state.test.ts
@@ -1,0 +1,51 @@
+import test from 'ava';
+import execute from '../src/execute';
+
+test.serial(
+  'use lazy-state in template literal & property access',
+  async (t) => {
+    const state = {
+      keyName: 'someKey',
+      firstName: 'John',
+      lastName: 'Doe',
+    };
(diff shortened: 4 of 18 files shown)


doc-han:
@josephjclark I want to add a good runtime/cli test for this. where do you think would be best to add that. 
```js
fn((state) => {
    state.callMeMaybe = (value) => {
        state.greetings = `hello ${value}`
		return state; // crazy! seems every function body that does something with state has to return it
    }
    return state
});

fn(state => {
    state.name = "John"
    return state;
})

fn($.callMeMaybe($.name)) // note: if callMeMaybe doesn't return state, next operation wouldn't have state passed.

```
Then the value at `state.greetings` will be `"hello John"`


### Note
We can say that. every function body that has something to do with state should return it.

josephjclark:
> every function body that has something to do with state should return it.

Uh, well not really. It depends on how it's used. We talk about Operations - a function which takes state and returns state. If you're using a function as an _operation_ - then yeah, the function body must return state.

But I can re-write your code like this:
```
fn((state) => {
    state.callMeMaybe = (value) => {
        state.greetings = `hello ${value}`
    }
    return state
});

fn(state => {
    state.callMeMaybe(); // no need to return state in this case
    return state;
})
```

But I think you're hitting on something important. Since we can't use lazy state inside a callback, the use of this is limited.

If I do:
```
fn($.callMeMaybe($.name))
```
Then the function must indeed be an Operation.

We can do stuff like this though
```
fn((state) => {
  state.buildUrl = (resource) => `api/v1/${resource}`
  return state;
});
get($.buildUrl(resourceName))
```

But thinking this through is making me really de-value the utility of functions on state. I think it's just creating confusion.

 #648 would be a much more useful facility (although the issue needs a lot of love!)

josephjclark:
@doc-han let me think about this but this does feel low priority now :thinking:  If it feels safe I'll merge it

josephjclark:
Hiya @doc-han - I want to merge this even though I don't think it's very valuable in the end.


We have a suite of tests in `integration-tests/execute` which test the compiler and runtime on job code. This is quite new and we don't have many - but it's nice to be able to test job code without the overhead of the CLI (or indeed the multi-process worker engine).

Can I ask you to create a file called `lazy-state.test.ts` in there and add a few simple lazy state tests and then a test or two on this specifically?

Don't go too overboard - just plug in a nice starting point :)

Thanks!

doc-han:
> Can I ask you to create a file called `lazy-state.test.ts` in there and add a few simple lazy state tests and then a test or two on this specifically?

Done

josephjclark (approved):

josephjclark:
Thanks @doc-han ! Appreciate it

PR #839, opened 2024-12-06
CLI: Collections support

### LINKED ISSUE #819: CLI: Support collections
It would surely be useful for the CLI to support the new Collections feature directly

A particularly strong  use-case would be to upload mapping data into a collection. I have 1000 mappings in my GH repo and want to upload them to lightning for use in my workflow. I don't really want to have to create a workflow step and copy and paste the values in for this. I want to call the collections API directly.

```
openfn collections get/count/keys/set/delete <name> <pattern> --input  --output --token --limit --url
```

* `get` would get values from a collection and return a JSON object of key/value pairs. Uses --output to write the result to disk (if unset, returns to stdout). Pass --json to parse values as JSON.
* `count` would return the number of values which match a particular pattern. Returns to stdout only.
* `keys` would return a list of matching keys without downloading any data (we don't even support this!)
* `set` would accept data through a file. CSV or JSON. If JSON, it must expect and object of key value pairs, and it should probably stringify values
* `delete` would delete keys according to the pattern and return the count of deleted keys.

## Auth

We need to think carefully about auth.

The request needs to include a personal access token.

But we probably don't want to take this as a flat env var because access tokens are scoped per project, so users will have several. Which makes it a bit fiddly to submit to the CLI.

Options:

- accept an env var name in the command, ie, `--token MY_OPENFN_TOKEN` (but surely `--token $MY_OPENFN_TOKEN` also works?)
- If a token isn't provided, prompt the user to paste it in (annoying)
- Allow tokens to be saved by the CLI (which must encrypt and write them to the repo somewhere). Now we only need to take a token once and the CLI can remember for that collection name

## Implementation details

Should this use the collections adaptor under the hood? Probably not, because it behaves very differently. But we might copy some code across to handle streaming efficiently.

josephjclark:
## Short Description

Adds collections support to the CLI

Closes #819

Docs are in https://github.com/OpenFn/docs/pull/613

## Usage Examples

```
openfn collections get my-collection \*

openfn collections get my-collection some-key

openfn collections get my-collection some-key --token $MY_OPENFN_PAT

openfn collections set my-collection some-key path/to/value.json

openfn collections set my-collection --items path/to/many-values.json

openfn collections remove my-collection 2024*

openfn collections remove my-collection 2024* --dry-run
```

Note: escaping `*` in the CLI is super important!

## Implementation Details

The most important things to note are these:

* OPENFN_PAT is used by default to load the user's PAT
* Otherwise, pass `--token` with a different PAT
* Set is deliberately split into "set one" and "set many", with the `--items` flag. It's a little cumbersome but designed, believe it or not, to reduce confusion.
* If getting some items, you should be able to directly set them from the result. So the formats have to be compatible
* Uploaded values are always loaded from a file
* If getting or setting a single item (no pattern), that item will be uploaded and downloaded verbatim, as a value. This works GREAT for mappings, which is the #1 use-case for CLI IMO
* If getting or setting multiple items, we use a JSON obejct to capture key/value pairs (not the `[{ key, value }]` format used on the wire)
* remove --dry-run is cool, but crudely implemented. Would be improved by https://github.com/OpenFn/lightning/issues/2758
* Gets will fetch all items in batches (controlled by `--page-size`), unless `--limit` is passed

## TODO

* [x]  Error handling
* [x] Add a suite of unit tests
* [x] Support  `created-since` etc flags
* [x] Proper support for limit

Other commands like `keys` and `count` will be implemented much later, when we have support form the backend

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


josephjclark:
Maybe `get collection-name` should get everything if you don't pass a key name? That dodges around the `*` problem a bit

josephjclark:
The API is very strongly built around JSON at the moment - all values are assume to JSON.

That's probably fine, but shouldn't we also be able to handle non-json values? I think so? We do say the collection stores string values.

It's probably reasonable to try and parse all value as JSON, and if they fail, just return the verbatim value.

doc-han:
> The API is very strongly built around JSON at the moment - all values are assume to JSON.
> 
> That's probably fine, but shouldn't we also be able to handle non-json values? I think so? We do say the collection stores string values.
> 
> It's probably reasonable to try and parse all value as JSON, and if they fail, just return the verbatim value.

Hmmm, for non-json values(binary data like images, ...) it would be a bit tricky to just encode them and store in the same place.

doc-han:
> Maybe `get collection-name` should get everything if you don't pass a key name? That dodges around the `*` problem a bit

I personally side with returning nothing when nothing is specified. give exactly the data they asked for. so if they asked for nothing, we give nothing.

josephjclark:
@doc-han Well 90% of our use-cases are going to be JSON based. I'm not sure why anyone would be saving binary data on our system - I'm not even sure why you'd store an arbitrary string really.

I might raise a `wontfix` issue to add support for `--json` (default true) and `--no-json` flags. This would give users and override and force JSON parsing off.

Meanwhile I also ought to add a catch for when JSON parsing fails, and we'll fallback to the raw value (perhaps with a warning)

Regarding the key thing, I'll leave it as it is now: a required field

Thanks for the input!

PR #848, opened 2024-12-19
Map errors positions against the original source (sourcemapping)

### LINKED ISSUE #230: Runtime: Report error positions against a sourcemap
I don't think this will be easy, but it should be possible. And it's not tracked here, so:

When an expression throws, we should report an error against its sourcemapped source, refering to the line and column number, job name, and if appropriate the file name.

This is a bit tricky right now because errors are reported from the the VM  inside SourceTextModule. I'm deliberately hiding stacktrace information because what we get for free is at best useless, at worst confusing.

The reported position has to refer back to the uncompiled code - although this may give us a problem if the error is in the  compiled stuff (like `import {fn }`).

josephjclark:
## Short Description

This PR uses sourcemapping to report accurate positions on error messages

It also adds some general improvements to error messages, for clarity, brevity and accuracy.

Fixes #230

## Implementation Details

This is not a particularly focused bit of work, as it impacts lots of corners of code - including some gnarly bits.

The basic strategy is this:

* The compiler now returns compiled code AND a source map (produced by recast). The sourcemap is then grafted onto the ExecutionPlan for each step, and used to map positions when errors are thrown.
* Error handling more or less works like it used to, except that we catch adaptor errors directly in `expression.ts` now, rather than working them out later. This is so that we can map positions on top-level operations.
* The actual mapping logic is in `runtime/src/util/sourcemap-errors.ts`. The sourcemap library does most of this for us - so mostly this module just updates metadata on the error object and parses the callstack to map positions there.
* We now append stack traces to errors, but we only include frames from VM code or adaptor code, and we modify the reported paths to be relative to the adaptor, rather than the full system path. I need more data to make a judgement on how well this is working.
* The final error output has been cleaned up a bit form production - see #852 for details and isolated changes for that.

API changes from the compiler have a big knock-on affect, resulting in a large diff.

Note that for top-level operations, this got a bit difficult because the stack trace is entirely in runtime code, not vm code. See #851 for details. Basically there's no meaningful position we can report for these cases - so what we do is work out which operation is responsible for the error, and say something like "error throw by the fn() operation on line 8".

Changes:
* [x] map error positions and stack traces against the original job source
* [x] Generate a text string representing the source line and mark where the error occurred
* [x] Work out how and where to log this. Probably in the runtime's error reporter, which I don't really like. Maybe this is a good time to refactor this?
* [x] Make sure positions are logged appropriately for all the right error types.
* [x] Get stacktrace showing the callstack within the VM, but not the runtime. I've had a go at this but it doesn't seem to work as I expect. I might branch this out as a new issue #852
* [x] Figure out errors coming from top-level operations, which don't have a vm stack #851 
* [x] Graft sourcemaps to the Workflow
* [x] Fix the CLI to use new compiler ouputs
* [x] Fix the worker to use new compiler outputs
* [x] Look closely at worker outputs and make sure they're OK
* [x] Test from the CLI and worker and show outputs here
* [x]  Refine serialized error objects (less noise, maybe don't log json) #852
* [x]  Fix an issue in the logger where `logger.print()` wasn't reporting to the Worker because it didn't have a timestamp

## QA Notes

I will be asking for help testing the CLI and the worker. There are a lot of changes here

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


josephjclark:
TODO: I'm confusing myself  here about the  error reporter and the role it plays. I really think I'm gonna rip it out as part of this work.

What we really need is:
1) for the runtime to serialize errors in a nice way
2) CLI and the Worker to print those errors in a nice format for their platforms

That means CLI logs some nice pretty messages, and the worker probably just logs a JSON object (but not stringified)

Not sure. Needs more thought.

josephjclark:
It might be a distraction but I want to go back and look at the example in https://github.com/OpenFn/kit/issues/582

It still doesn't serialize well (in prod or this branch), and just shows the error as `[object Object]`

But when I look at the stack trace, there's no stack in the job code, so we can't extract any error information.

I think it's something to do with the axios callback - the error occurs in an async callstack, hence there's nothing useful in the stack to report. Is there anything we can everdo about this?

Btw to repro the error (and get a real error message out), do this:
```
import { CookieJar } from 'tough-cookie';

const cookiejar = new CookieJar();
cookiejar.getCookies({});
```

This goes away in later http versions, but I'm still worried about / interested in this async error case.

josephjclark:
Errors which occur within the job code directly are currently working well.

But I'm struggling for errors within adaptor code, and that's a real problem.

Basically when an error is thrown from inside an adaptor, the stack trace does not include the vm sandbox. So you get a stack like this:
```
Error: UNEXPECTED_RELATIVE_URL
    at assertUrl (/home/joe/repo/openfn/cli-repo/node_modules/@openfn/language-http_6.5.1/dist/index.cjs:133:15)
    at /home/joe/repo/openfn/cli-repo/node_modules/@openfn/language-http_6.5.1/dist/index.cjs:167:5
    at file:///home/joe/repo/openfn/kit/packages/runtime/dist/index.js:630:24
    at async file:///home/joe/repo/openfn/kit/packages/runtime/dist/index.js:597:22 
```

What I rely on for source mapping is a frame inside the vm, like this:
```
ReferenceError: x is not defined
    at vm:module(0):3:27
    at /home/joe/repo/openfn/cli-repo/node_modules/@openfn/language-common_2.1.1/dist/index.cjs:333:12
    at file:///home/joe/repo/openfn/kit/packages/runtime/dist/index.js:628:24
    at async file:///home/joe/repo/openfn/kit/packages/runtime/dist/index.js:595:22
```

Where that `vm:module` frame is the job code position

I don't really understand why the vm frames are omitted from the stack, but it'll be something to do with how the node vm module works I'm sure. I'll have to get my thinking cap on about this.

josephjclark:
Here is the solution:
* When the compiler runs, it generates a list of all the operation names, in order, with their starting line number.
* That information is appended to the source map - and index of operation number against name and line
* The runtime operation wrapper should catch any exceptions. If the first frame is not in VM code - if this error came from an operation - we wrap the error, or maybe just add information to it, so that the error contains a line number. Probably we just go and create our AdaptorError object right here, with an order number
* We also by the way customise the stack trace at this point to show the stack inside the adaptor. We probably exclude paths as well (maybe they're useful in the CLI?)
* When later reporting the error against the source map, if it has an operation number, we report the operation name and line from the extended source map. INSTEAD of the the sourcemapped position

This does make adaptor errors a bit different. They'll be attributed back to the operation generally, rather than a aprt of it. Which makes sense if you think about it.

 I am a bit worried about NESTED operations - I think if the nested operation throws, it's the root operation that will catch and report the error. This is an antipattern and I guess it's doubly so now. One benefit here I suppose is that the nested operation call might have vm frames in the sttcktrace. That would enable us to catch and identify the trace.

I will take a look at implementing this on Thursday and Friday this week. If I can't get it done in that time, I'll spin it out into a new issue and we  can at least ship sourcemapping on VM errors.

josephjclark:
Hey @doc-han - are you up for a bit of a review over the runtime changes next week?

I'll be putting this out to the team for QA in the CLI and Worker, I want to test this really thoroughly. But I'd welcome your feedback on the solution (if you can figure it out - it's quite a messy PR tbh!)

josephjclark:
```
R/T TypeError: state.data is not a function
    at vm:module(0):3:9
    @openfn/language-common_2.1.1/dist/index.cjs:333:12
R/T Error occurred at: New job
R/T 3:   state.data()
R/T            ^     
```

Maybe on "error occurred at" we should say `error occurred at ${step.name} (line N column N)`. That also makes the following source line a bit clearer

josephjclark:
Hey @doc-han I'm targeting a release of this on Monday next week. A review from you isn't essential but it is appreciated - if you'd like to review it, it needs to be in the next couple of days please :)

doc-han:
> Hey @doc-han I'm targeting a release of this on Monday next week. A review from you isn't essential but it is appreciated - if you'd like to review it, it needs to be in the next couple of days please :)

Alright. I'll commence with review today then.

packages/runtime/test/errors.test.ts
```diff
@@ -47,13 +48,18 @@ test('crash on runtime error with SyntaxError', async (t) => {
   t.is(error.message, 'SyntaxError: Invalid or unexpected token');
 });
 
-test('crash on runtime error with ReferenceError', async (t) => {
+test.only('crash on runtime error with ReferenceError', async (t) => {
```
doc-han: are we still running this test only?

doc-han: forget. going through by commits first.

josephjclark: Ahh don't go by commit, you'll be there for a million years :joy:  This PR has quite a long realtime history and a few pivots along the way. Also my commit history is rarely very neat

josephjclark: If it helps we can do a pairing session and I can talk you through some of the key areas, then you can review them in your own time

doc-han: So my review process is to go by commit to understand the entire history and then finally do a broad overview review. 

Yup, the commits don't seem neat.

packages/runtime/src/errors.ts
```diff
@@ -137,9 +182,34 @@ export class AdaptorError extends RTError {
...
   details: any;
-  constructor(error: any) {
+
+  line?: number;
+  operationName?: string;
+
+  constructor(error: any, line?: number, operationName?: string) {
     super();
-    this.details = error;
+    if (!isNaN(line!)) {
```
doc-han, line 191: Quite tricky. `isNaN` doesn't do great job of telling whether a value isn't a number. 

It thinks:
1. an empty string is a number
2. `null` is a number

Hence, using `!isNaN` to check for something being a number might be flawed when the value passed at runtime is either `null` or `''`. 

Great to add a `typeof` check to make it get only numbers. 
```suggestion
    if (typeof line === "number" && !isNaN(line!)) {
```

Hence, typeof == number will make sure that we're scoped to numbers(1,2,3,.... & NaN).

doc-han: In case a function like `parseInt` which is supposed to always return a number(NaN if not) is used before here, then line will always be in the number scope.

doc-han (approved):
Awesome!

I was thinking we could get a path to source code for a job running via a file.
eg. 
Instead of `at vm:module(0): 5:8` we get at `/jobs/get-data.js: 5:8` so that user can just click to file but then I started thinking of lightning.

doc-han:
The currently failing tests are right. We can't assume that a compiled code will always error at line `333:12`. 
Also, the adaptor version.

josephjclark:
> I was thinking we could get a path to source code for a job running via a file.

yeah, I'm pretty sure when we create the node vm that we can control the name of the vm, so instead of `vm:module` we can print `fetch-patients`. I would use the step NAME instead of the path (the CLI will default a name from the path) - I think we always have one.

I haven't found time to look into it properly. We can spin out a new issue to implement later - I'm sure it's not a big deal. We just to think about, in regard to THIS PR, making these custom vm frames easy to detect in stack traces (so maybe we prefix each name with `vm:`, ie, `vm:fetch-patients`

packages/runtime/src/errors.ts
josephjclark: Is `typeof number` robust? For some reason I didn't think it was :thinking:  Maybe I've invented that...

josephjclark: Oh look, lodash uses `typeof number`. OK cool, I accept!

josephjclark: Ok so I look at this code and I had no idea what was going on :shrug: 

The number thing isn't important. Basically when generating an adaptor error, we EITHER attribute it to a source offset in the VM, OR we attribute it to an operation (with a name and line number)

In https://github.com/OpenFn/kit/pull/848/commits/8882c4cd984b012dc135f88a45df14585de59662 I've cleaned up the AdaptorError so that the constructor reflects the correct behaviour. I've dropped a couple of unit tests too to explain and enforce the behaviour. Although the real intereesting logic behind this behaviour is elsewhere.

PR #856, opened 2025-01-14
Fix errors when using the monorepo

### LINKED ISSUE #855: Runtime: some error aren't logged at all!
Found a bit of a strange case in the new error handler (a shame we couldn't find this in QA)

I think if adaptor (or job?) code throws `new Error()`, we don't log it very nicely. Error details aren't logged, and there is no positional information.

For example:
```
[CLI] ♦ Versions:
         ▸ node.js                    18.17.1
         ▸ cli                        1.10.0
         ▸ @openfn/language-common    monorepo
[CLI] ✔ Loading adaptors from monorepo at /home/joe/repo/openfn/adaptors
[CLI] ⚠ Skipping auto-install as monorepo is being used
[R/T] ✘ job-1 aborted with error (130ms)

[R/T] ✘ TEST_ERROR: this is a test
[R/T] ✘ Check state.errors.job-1 for details
[CLI] ✔ State written to tmp/output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 199ms
```

The test error there should have a position, and should log details (key/value pairs)

You can repro by adding this to common:
```
export function tmp() {
  return state => {
    throwError('TEST_ERROR', { description: 'this is a test', jam: 'jar' });
  };
}
```

josephjclark:
## Short Description

When the vm throws an error, we try and determine if that error comes from an adaptor. This makes assumptions about the callstack and the path of the file. It works fine for production adaptors, but when loading from the monorepo these tests fail.

Fixes #855 

## QA Notes

I am reproducing like this:

* In the monorepo, add this to common:
```
```
* Create a simple job which just calls `tmp()`
* From `packages/cli`, run `pnpm openfn tmp/job.js -ma common` (where tmp/job.js points to the simple job)
* Against production, the error contains no details. Here, you get output like:
```
[CLI] ♦ Versions:
         ▸ node.js                    22.12.0
         ▸ cli                        1.10.0
         ▸ @openfn/language-common    monorepo
[CLI] ✔ Loading adaptors from monorepo at /home/joe/repo/openfn/adaptors
[CLI] ⚠ Skipping auto-install as monorepo is being used
[R/T] ✘ job-1 aborted with error (100ms)

[R/T] ✘ Error reported by "tmp()" operation line 1:
[R/T] ✘ Error: TEST_ERROR: this is a test
[R/T] ✘ Additional error details:
{
  type: 'Error',
  message: 'TEST_ERROR: this is a test',
  code: 'TEST_ERROR',
  description: 'this is a test',
  jam: 'jar'
}

[R/T] ✘ Check state.errors.job-1 for details
[CLI] ✔ State written to tmp/errors/output.json
[CLI] ⚠ Errors reported in 1 jobs
[CLI] ✔ Finished in 152ms
```

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/runtime/src/execute/expression.ts
+++ b/packages/runtime/src/execute/expression.ts
@@ -146,11 +146,16 @@ export const wrapOperation = (
 
         let firstFrame;
 
-        // find the first error from a file or the VM
-        // (this cuts out low level language errors and stuff)
+        // find the first frame from an adaptor or the VM
+        // (this cuts out low level language errors like TypeError)
         do {
           const next = frames.shift();
-          if (/(@openfn\/language-)|(vm:module)/.test(next)) {
+          if (
+            // detect an adaptor prod, adaptor monorepo, or vm frame
+            /(@openfn\/language-)|(adaptors\/packages\/.+\/dist)|(vm:module)/.test(
+              next
+            )
+          ) {
             firstFrame = next;
             break;
           }
--- a/packages/runtime/src/execute/expression.ts
+++ b/packages/runtime/src/execute/expression.ts
@@ -137,6 +137,7 @@ export const wrapOperation = (
     try {
       result = await fn(newState);
     } catch (e: any) {
+      console.log(e);
       if (e.stack) {
         const containsVMFrame = e.stack.match(/at vm:module\(0\)/);
 
@@ -152,9 +153,7 @@ export const wrapOperation = (
           const next = frames.shift();
           if (
             // detect an adaptor prod, adaptor monorepo, or vm frame
-            /(@openfn\/language-)|(adaptors\/packages\/.+\/dist)|(vm:module)/.test(
-              next
-            )
+            /(@openfn\/language-)|(packages\/.+\/dist)|(vm:module)/.test(next)
           ) {
             firstFrame = next;
             break;
--- a/.changeset/eleven-owls-decide.md
+++ b/.changeset/eleven-owls-decide.md
@@ -1,0 +1,7 @@
+---
+'@openfn/runtime': patch
+'@openfn/cli': patch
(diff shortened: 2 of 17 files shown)


packages/runtime/src/execute/expression.ts
```diff
@@ -146,11 +146,16 @@ export const wrapOperation = (
...
-        // find the first error from a file or the VM
-        // (this cuts out low level language errors and stuff)
+        // find the first frame from an adaptor or the VM
+        // (this cuts out low level language errors like TypeError)
         do {
           const next = frames.shift();
-          if (/(@openfn\/language-)|(vm:module)/.test(next)) {
+          if (
+            // detect an adaptor prod, adaptor monorepo, or vm frame
+            /(@openfn\/language-)|(adaptors\/packages\/.+\/dist)|(vm:module)/.test(
```
josephjclark, line 155: This is a bit hairy

We're looking in the callstack or a path like `adaptors/packages/common/dist`

Because the path to an adaptor in the monorepo isn't terribly distinctive.  path like`packages/*/dist` feels like it might erroneously match against many paths (then again maybe not? Most node_modules code would be like `packagename/dist`, with nothing in between.

Anyway, because of this, I wanted to put the repo name in the path.

But that doesn't work for anyone who forks the repo with a different name locally. In those cases, the path won't match and bad errors will be triggered, which leads to  misleading error reports when developing adaptors.

Perhaps I can do a regex like "not node_modules and packages/./dist"

josephjclark: Ah it's fine - I'm sure `packages/./dist` is all we need for this. There's only a small amount of non-vm error cases that we need to handle. Let's try it out.

PR #857, opened 2025-01-14
fix: replace adaptors in plan with resolved adaptors at install

### LINKED ISSUE #853: CLI: running `-a common@latest` throws an error in 1.9.1
Investigate, this should be fine!

doc-han:
## Short Description

When alias adaptor versions are used in cli, the runtime is unable to resolve them. resulting in the error below

Screenshot 2025-01-14 at 11 55 13 AM

This is because, During the installation step, `@latest` is resolved to the latest version but then during execution, we try linking `@latest` instead of the resolved version.

Fixes #853 

## Implementation Details
Replacing adaptors in a plan with already resolved adaptors before the plan gets passed to the compiler and the runtime.

<table><thead>
  <tr>
    <th>Workflow</th>
    <th>Logs</th>
  </tr></thead>
<tbody>
  <tr>
    <td>Screenshot 2025-01-14 at 11 55 13 AM</td>
    <td>Screenshot 2025-01-14 at 11 55 13 AM</td>
  </tr>
</tbody>
</table>

`@latest` at the point of writing is `v2.2.0`

## QA Notes

Run CLI by passing any adaptor with an alias version like `-a http@next`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/integration-tests/cli/test/autoinstall.test.ts
+++ b/integration-tests/cli/test/autoinstall.test.ts
@@ -49,6 +49,10 @@ test.serial(
     t.regex(stdout, /Installing packages.../);
     t.regex(stdout, /Will install @openfn\/language-common version/);
     t.regex(stdout, /Installed @openfn\/language-common@/);
+    t.regex(
+      stdout,
+      /Resolved adaptor @openfn\/language-common to version \d+(\.\d+)*/
+    );
   }
 );
 
@@ -68,6 +72,10 @@ test.serial(
       stdout,
       /Skipping @openfn\/language-common@(.+) as already installed/
     );
+    t.regex(
+      stdout,
+      /Resolved adaptor @openfn\/language-common to version \d+(\.\d+)*/
+    );
   }
 );
 
@@ -78,7 +86,6 @@ test.serial(
     const { stdout, err } = await run(t.title);
 
     // t.falsy(err); // TODO I think this is a broken adaptor build?
-
     t.regex(stdout, /Auto-installing language adaptors/);
     t.regex(stdout, /Looked up latest version of @openfn\/language-testing/);
     t.regex(stdout, /Installing packages.../);
@@ -91,6 +98,12 @@ test.serial(
     t.regex(
       stdout,
       new RegExp(`Installed @openfn\/language-testing@${TEST_LATEST}`)
+    );
+    t.regex(
+      stdout,
+      new RegExp(
+        `Resolved adaptor @openfn/language-testing to version ${TEST_LATEST}`
+      )
     );
   }
 );
--- a/integration-tests/cli/test/execute-workflow.test.ts
+++ b/integration-tests/cli/test/execute-workflow.test.ts
@@ -199,3 +199,25 @@ test.serial(
     });
   }
 );
+
+test.serial(
+  `openfn ${jobsPath}/different-adaptor-versions.json -l debug`,
+  async (t) => {
+    const { err, stdout } = await run(t.title);
+    t.falsy(err);
+
+    t.regex(
+      stdout,
+      /Resolved adaptor @openfn\/language-common to version 2.1.0/
+    );
+    t.regex(
+      stdout,
+      /Resolved adaptor @openfn\/language-common to version 2.0.3/
+    );
+
+    const out = getJSON();
+    t.deepEqual(out, {
+      z: 1,
+    });
+  }
+);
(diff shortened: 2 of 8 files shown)


doc-han:
@josephjclark 
When we get to the file below. We have two linker module informations. 
1. via `opts` which is coming from outside. 
2. we generate another one from the original workflow. 

Currently, we just pick 2. and that's what's causing the issue. 1. might contain already resolved modules. Hence we make 1. override 2. 
https://github.com/OpenFn/kit/blob/main/packages/runtime/src/execute/step.ts

packages/cli/src/execute/execute.ts
```diff
@@ -73,7 +76,8 @@ export function parseAdaptors(plan: ExecutionPlan) {
     // But there are a couple of cases mostly in test, when validation is skipped,
     // when the array may not be set
     // It's mostly redundant nbut harmless to optionally chain here
-    job.adaptors?.forEach((adaptor) => {
+    const jobAdaptors = (job.adaptors || []).concat(resolvedAdaptors);
```
josephjclark, line 79: If I have a workflow with two steps and different versions of adaptors - let's say `common@latest` and `common@1` - this will get us in trouble. Because versions of `common` will resolve to the latest.

A bit disappointing that I don't have a failing test for this :(

josephjclark:
Well done for finding a solution @doc-han. Some of this code is pretty gnarly. I'm looking over some of this stuff now and I'm a bit puzzled myself!

I'm not sure that this is the answer though, see my comment.

What if we did this instead, somewhere top of the execute handler:

- Run the autoinstall
- Take the list of resolved adaptors
- Traverse each step of the workflow, and for each adaptor with `@latest`, replace `@latest` with the version number.

We'll do that immediately after autoinstall.

Then the workflow that gets passed into the compiler and runtime will have fully resolved version numbers, and the complication basically goes away.

This won't affect the version readout at the start of execution, because that runs before we do any autoinstall processing. 
```
[CLI] ♦ Versions:
         ▸ node.js                    22.12.0
         ▸ cli                        1.10.0
         ▸ @openfn/language-common    latest
```
This is fine. I would probably prefer that we show the version (or even better, we show `@openfn/language-common    2.0.0 (latest)` - but that's a different story, and not something I want to address here

packages/cli/src/execute/execute.ts
doc-han: Yup, that's true. Just tested and the trouble mentioned holds.

doc-han:
> Traverse each step of the workflow, and for each adaptor with `@latest`, replace `@latest` with the version number.
Yeah, this should work!

packages/cli/src/execute/handler.ts
```diff
@@ -67,6 +69,14 @@ const executeHandler = async (options: ExecuteOptions, logger: Logger) => {
           { packages: autoInstallTargets, repoDir },
           logger
         );
+
+        // create a module resolution dictionary.
+        // this is to map aliases like @latest & @next to what they resolved into
+        if (autoInstallTargets.length === options.adaptors.length) {
```
doc-han, line 75: This is expected to always be equal. 
Because after installation. The runtime returns list of all resolved adaptors which is a one-to-one mapping to the list of installTargets passed to it.

packages/cli/src/execute/execute.ts
doc-han: Resolved!

integration-tests/cli/test/autoinstall.test.ts
```diff
@@ -49,6 +49,10 @@ test.serial(
     t.regex(stdout, /Installing packages.../);
     t.regex(stdout, /Will install @openfn\/language-common version/);
     t.regex(stdout, /Installed @openfn\/language-common@/);
+    t.regex(
+      stdout,
+      /Resolved adaptor @openfn\/language-common to version \d+(\.\d+)*/
+    );
```
doc-han, line 55: This Resolved test makes sure that linking was done successfully version is a number not string. 
This is added to an already existing test for `@latest` autoinstall

integration-tests/cli/test/execute-workflow.test.ts
```diff
@@ -199,3 +199,25 @@ test.serial(
     });
   }
 );
+
+test.serial(
```
doc-han, line 203: A test for a workflow that uses two different versions of the same Adaptor. 
We know it used both different versions if they're both resolved.

josephjclark (approved):
Fantastic @doc-han! PR is beautiful, tests are awesome, and it works great for me locally. Thank you! :pray:

PR #867, opened 2025-02-11
ws-worker: support job log level in lightning plan

### LINKED ISSUE #868: ws-worker: allow configuration of `jobLogLevel`
Users in lightning want the ability to turn on/off logging `console.log()` statements. This issue makes it possible for lightning to achieve this by passing the `job_log_level` option in lightning plan.

Related to: https://github.com/OpenFn/lightning/issues/2205

midigofrank:
## Short Description

This PR adds support for `job_log_level` in lightning plan. This is eventually passed down to the engine

Fixes #868 

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/unlucky-panthers-try.md
+++ b/.changeset/unlucky-panthers-try.md
@@ -1,0 +1,6 @@
+---
+'@openfn/engine-multi': minor
+'@openfn/ws-worker': minor
+---
+
+Allow configuration of job log level
--- a/.changeset/unlucky-panthers-try.md
+++ b/.changeset/unlucky-panthers-try.md
@@ -1,5 +1,6 @@
 ---
 '@openfn/engine-multi': minor
+'@openfn/lexicon': minor
 '@openfn/ws-worker': minor
 ---
 
--- a/.changeset/unlucky-panthers-try.md
+++ b/.changeset/unlucky-panthers-try.md
@@ -1,7 +1,0 @@
----
-'@openfn/engine-multi': minor
-'@openfn/lexicon': minor
-'@openfn/ws-worker': minor
----
-
-Allow configuration of job log level
--- a/integration-tests/execute/CHANGELOG.md
+++ b/integration-tests/execute/CHANGELOG.md
@@ -1,5 +1,12 @@
 # @openfn/integration-tests-execute
 
+## 1.0.16
+
+### Patch Changes
+
+- @openfn/compiler@1.0.1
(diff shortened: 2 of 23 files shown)


midigofrank:
Hey @josephjclark , need your help with possibly writing a test for this

josephjclark:
@midigofrank yeah given how the option is passed, it's a bit difficult to test this.

I would add an integration test for this. Maybe look at `integration-tests/worker/test/integration.test.ts`. Create a run, pass the options, and fail if if any log events get published.

midigofrank:
Got it. Let me try that

midigofrank:
@josephjclark I've added the changeset

stuartc (commented):
Why is it called `job_run_level` if it applies to all job? Would `run_log_level` be more specific? Happy to go with the majority here, but curious about the use of the word "job" in the context of a run.

midigofrank:
@stuartc it's called `job` because there are 3 loggers in the worker: `R/T` (runtime) , `JOB`, `ADA` (adaptor).  In this case we are targeting the job logger, so we should keep it as `job_log_level`. 

@josephjclark correct me if I'm wrong

josephjclark:
@midigofrank that's basically right - we do have a bunch of loggers, and internally to the worker we call the one we're trying to call the "job logger". Maybe if I was starting again today it would be called the "step" logger (but maybe not?)

The key is we want to differentiate "system/worker/GCP" log lines from "vm/job code/user" log lines - that's the key distinction here. 

This option will change the user's logs - their console.log lines, plus anything coming out of the adaptor.

`run_log_level` probably makes sense from lighting's point of view? That's like "what level of logs _inside_ the run do we want?". `runtime_log_level` might also work.

We can map whatever option lightning sends - job_log_level` or `step_log_level` or `superted` or whatever - into `jobLogLevel` internally in the worker. 

So the question is: what's the best name from Lighting's perspective for this?

midigofrank:
I'm still voting to have it remain as `jobLogLevel`.  `run_log_level` is too general. All those logs are for a `run`.

josephjclark:
That's ok by me!

midigofrank:
@josephjclark anything blocking this from going to main?

josephjclark:
@midigofrank sorry this took so long, entirely my fault. I think this is ready now and I'll merge when the build is complete

josephjclark:
What on earth have I done here... :rage:

PR #872, opened 2025-02-12
Handle DataCloneErrors

josephjclark:
## Short Description

Details to follow....

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/engine-multi/src/test/worker-functions.ts
+++ b/packages/engine-multi/src/test/worker-functions.ts
@@ -4,6 +4,8 @@ import { register, publish, threadId } from '../worker/thread/runtime';
 import { increment } from './counter.js';
 import { ExecutionPlan, Job } from '@openfn/lexicon';
 
+// import { Readable } from 'node:stream';
+
 const tasks = {
   test: async (result = 42) => {
     publish('test-message', {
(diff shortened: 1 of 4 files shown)


packages/engine-multi/src/test/worker-functions.ts
```diff
@@ -4,6 +4,8 @@ import { register, publish, threadId } from '../worker/thread/runtime';
 import { increment } from './counter.js';
 import { ExecutionPlan, Job } from '@openfn/lexicon';
 
+// import { Readable } from 'node:stream';
```
josephjclark, line 7: TODO remove this

packages/engine-multi/src/worker/thread/runtime.ts
```diff
@@ -38,7 +39,8 @@ export const publish = (
     type,
     threadId,
     processId,
-    ...payload,
+    // ensure the event message is serializable
+    ...JSON.parse(stringify(payload)),
```
josephjclark, line 43: So this is my solution for now.

It's pretty brute force.

Every event that gets sent through the thread will be serialized with `fast-safe-stringify`, then parsed back into JSON, and then published (Where presumably it's serialized again)

josephjclark:
The current approach is pretty brute force: we force serialisation of every event. This could get expensive for large objects.

I'd like to be more selective? 99% of our events are perfectly serializable. We just have to watch out for the exceptions, which are:
- console.log lines from adaptor or job code. Users could log anything here, and it probably makes sense to force stringify this stuff.
- Exceptions that are thrown and then processed (This is actually the use-case that triggers this business - errors getting thrown)
- Unserializable stuff written to state, which then gets emitted to eg step-complete. But the runtime should ensure that final state is serializable anyway right? So this doesn't seem helpful

PR #877, opened 2025-02-20
feat: ws-worker respond to wakeup call

### LINKED ISSUE #869: ws-worker: respond to a wake up call
When each worker starts up, it joins a websocket channel with lightning (see `src/channels/worker-queue.ts`)

It then polls that channel to "claim" any outstanding work items. Lighting will respond with 0 or more runs. If the worker receives no work, it'll back-off and poll a bit slower next time.

We want to add a feature where Lightning will send a "wake up" call to any workers with capacity. This allows us to short-cut the backoff and instantly claim any work, which lets is process runs faster.

So, the Worker needs to listen to some kind of event from lightning and, if it has capacity, instantly reset the backoff and trigger a fresh claim.

On the lightning side, this wake-up call needs to be sent similtaneously to everyone listening on the work queue. That could be a hundred workers. But it's Ok, because each worker will trigger a claim. The first one that happens to do so will get the work, and the others will be safely returned nothing.

We can engineer a response to this event on the worker side and test it with integration tests. Just set a super low backoff, trigger a wake up, and work should be instantly claimed. It's up to the Lightning team to work out how and when to send the wake up event.

Questions:

- What is the wake up event called? `on-queue-added`? `wake-up`? `trigger-claim`? `cancel-backoff`?

doc-han:
## Short Description

This PR allows a lightning instance to trigger a claim in all ws-workers connected to it.

Fixes #869 

## Implementation Details

**1. Lightning mock**
1. The `/run` function on lightning should now expose a query `?wakeup=true`. 
2. whenever a request to `/run?wakeup=true` happens. emit data to all connected clients on the `"worker:queue"`
3. eg. data
```json
{
  "topic": "worker:queue",
  "event": "work-available",
  "payload": {},
  "join_ref": "",
  "ref": "",
}
```

**2. Ws-Worker**
Whenever a worker gets a `work-available` event, it triggers a claim instantly!

## QA Notes

❗️QA? Be aware of backoff times. If backoff is in a very short time. then backoff might end up claiming instead of your manually initiated claim. ("in a very short time" is when the backoff time is exactly up in this case, less probability)

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/packages/lightning-mock/src/api-dev.ts
+++ b/packages/lightning-mock/src/api-dev.ts
@@ -149,7 +149,6 @@ const setupRestAPI = (
   const router = new Router();
 
   router.post('/run', (ctx) => {
-    const isWakeUp = ctx.query.wakeup === 'true';
     const run = ctx.request.body as LightningPlan;
 
     if (!run) {
@@ -177,7 +176,10 @@ const setupRestAPI = (
     app.enqueueRun(run);
 
     // triggering wakeup in all connected workers
-    if (isWakeUp)
+    if (new Boolean(ctx.query.wakeup)) {
+      logger.info(
+        'WAKE UP! Sending work-available event to all listening workers'
+      );
       app.messageSocketClients({
         topic: 'worker:queue',
         event: 'work-available',
@@ -185,6 +187,7 @@ const setupRestAPI = (
         join_ref: '',
         ref: '',
       });
+    }
     ctx.response.status = 200;
   });
 
--- a/packages/lightning-mock/src/api-dev.ts
+++ b/packages/lightning-mock/src/api-dev.ts
@@ -176,7 +176,7 @@ const setupRestAPI = (
     app.enqueueRun(run);
 
     // triggering wakeup in all connected workers
-    if (new Boolean(ctx.query.wakeup)) {
+    if ('wakeup' in ctx.query) {
       logger.info(
         'WAKE UP! Sending work-available event to all listening workers'
       );
--- a/packages/lightning-mock/src/api-dev.ts
+++ b/packages/lightning-mock/src/api-dev.ts
@@ -14,9 +14,11 @@ import type {
 import { ServerState } from './server';
 import { RUN_COMPLETE } from './events';
 import type { DevServer, LightningEvents } from './types';
+import { PhoenixEvent } from './socket-server';
 
 type Api = {
   startRun(runId: string): void;
+  messageClients(message: PhoenixEvent): void;
 };
 
 const setupDevAPI = (
@@ -39,6 +41,10 @@ const setupDevAPI = (
   };
 
   app.getDataclip = (id: string) => state.dataclips[id];
+
+  app.messageSocketClients = (message: PhoenixEvent) => {
+    api.messageClients(message);
+  };
 
   app.enqueueRun = (run: LightningPlan, workerId = 'rte') => {
     state.runs[run.id] = run;
@@ -143,6 +149,7 @@ const setupRestAPI = (
   const router = new Router();
 
   router.post('/run', (ctx) => {
+    const isWakeUp = ctx.query.wakeup === 'true';
     const run = ctx.request.body as LightningPlan;
 
     if (!run) {
@@ -169,6 +176,15 @@ const setupRestAPI = (
 
     app.enqueueRun(run);
 
+    // triggering wakeup in all connected workers
+    if (isWakeUp)
+      app.messageSocketClients({
+        topic: 'worker:queue',
+        event: 'work-available',
+        payload: {},
+        join_ref: '',
+        ref: '',
+      });
     ctx.response.status = 200;
   });
 
--- a/packages/lightning-mock/src/socket-server.ts
+++ b/packages/lightning-mock/src/socket-server.ts
@@ -80,6 +80,7 @@ type MockSocketServer = typeof WebSocketServer & {
     topic: Topic,
     events: Record<string, EventHandler>
   ) => { unsubscribe: () => void };
+  sendToClients: (message: PhoenixEvent) => void;
 };
 
 function createServer({
@@ -94,6 +95,8 @@ function createServer({
     // create a stub listener for pheonix to prevent errors
     phoenix: new Set([() => null]),
   };
+
+  const clients: Set<DevSocket> = new Set();
 
   const wsServer =
     server ||
@@ -202,6 +205,9 @@ function createServer({
 
         logger?.debug(`>> [${topic}] ${event} ${ref} :: ${stringify(payload)}`);
 
+        // tracking connected worker:queue workers
+        if (topic === 'worker:queue' && event === 'phx_join') clients.add(ws);
+
         if (event in events) {
           // handle system/phoenix events
           // @ts-ignore
@@ -227,9 +233,18 @@ function createServer({
         }
       }
     });
+
+    ws.on('close', () => clients.delete(ws));
   });
 
   const mockServer = wsServer as MockSocketServer;
+
+  mockServer.sendToClients = async (message) => {
+    clients.forEach((client) => {
+      // @ts-ignore
+      client.sendJSON(message);
+    });
+  };
 
   // debug API
   // TODO should this in fact be (topic, event, fn)?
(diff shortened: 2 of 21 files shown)


doc-han:
🏎️ tests on their way!

doc-han:
@josephjclark good to go!

doc-han:
Hence, with your scenario. yes. worker 1 will claim regardless of its current backoff 🎉

taylordowns2000:
@doc-han , on this:
> Caveat: Even tho claim checks it's current capacity before claiming. The logic in the snippet above seems to reject with an error. Hence, there might be a spike in this specific error.
Is the rejected promise in line 39 handled somewhere? Based on the wake up call, it seems to me that this isn't really an _error_ per se, but now a totally normal and expected part of the claim behaviour.

In other words, if all 10 workers are at capacity and lightning tells them to claim, I wouldn't want to see 10 errors pop up in Sentry (or even in the logs). I'd want to see all 10 workers say, "Nah bro... chill. I'm already working. Get off my back about it. I'll check for more work as soon as I'm done." (Not literally, but I think you get my point?)

josephjclark:
So not sure about the error thing (I'll have a look in a minute), but yes the design is:
- Lightning triggers a wake up call
- All workers _with capacity_ will issue a claim request
- First to get the claim will get the work
- The rest will happily receive no work
- Backoff is not affected for workers which do not receive work (this seems appropriate)

Uh, I need to think about when work _is_ claimed,  should the backoff be reset? Because probably yes? Otherwise after forcing a claim, we've short-cut the queue and all workers will sit on maximum backoff, even though there just was some work. This doesn't really matter if lightning issues a wake up every time something is added to the queue.

josephjclark:
@doc-han I'll get back to you - I'm just gonna do a deep dive and make sure I understand the current behaviour

taylordowns2000:
@josephjclark , on this:

> Uh, I need to think about when work is claimed, should the backoff be reset? Because probably yes? Otherwise after forcing a claim, we've short-cut the queue and all workers will sit on maximum backoff, even though there just was some work. This doesn't really matter if lightning issues a wake up every time something is added to the queue.

My understanding about how the backoff works is this:
1. if you check and don't see work, wait before checking again.
2. continue like this with exponential backoff
3. when you finish work a run, **_attempt to claim instantly_**
4. see step 1

doc-han:
> 3. when you finish working on a run, **_attempt to claim instantly_**
...
> i'm like 90% sure that's how you guys built it

Yup. I doubt 3. holds
It attempts a claim right after a run only when it was going to be overworked but then decided to ignore and stop the workloop. 

eg. It was at full capacity and still tried to do a claim. it'll ignore the additional work & stop the workloop. Hence right after its done with a run which will make it have some capacity, that's when it does a quick claim after run (because it restarts workloop which starts with an initial claim)

We can have a `restartWorkloop` which simple restarts the workloop when we want to claim manually.

josephjclark:
@taylordowns2000 I love how you know the worker better than I do :joy: I have no idea

Looking at the code I don't think that's right though. Basically we only immediately claim if the server is at capacity - otherwise we leave the backoff where it is.

I'm going to suggest we leave it as-is. If we think we need to add auto-claim on complete, it's a fairly easy thing to add later.

@doc-han That error thing doesn't matter and you should not add any special handling around it. Any error in the claim is caught by the backoff loop and the server silently backoff. There's no sentry reporting or anything.

(We  actually need to add some logic so that certain errors kill the worker or workloop or something - but right now we just backoff whenever claim throws, including for no claims returned)

taylordowns2000:
Are you totally sure? I really think that when the worker drops off their finished work, they ask if there's anything to be done next?

Or, at the very least, when work is _discovered_ the backoff is reset to zero. Otherwise the backoff would just increase forever.

doc-han:
> Are you totally sure? I really think that when the worker drops off their finished work, they ask if there's anything to be done next?
> 
> Or, at the very least, when work is _discovered_ the backoff is reset to zero. Otherwise the backoff would just increase forever.

I've not tested but from the code I've seen. when worker is done with a run. it waits for the next backoff time.

packages/ws-worker/src/server.ts
```diff
@@ -135,6 +135,12 @@ function connect(app: ServerApp, logger: Logger, options: ServerOptions = {}) {
     logger.debug(e);
   };
 
+  // handles messages for the worker:queue
+  const onMessage = (event: string) => {
+    if (event === WORK_AVAILABLE)
+      claim(app, logger, { maxWorkers: options.maxWorkflows });
```
josephjclark, line 141: Oh this is the error thing right?

So a couple of things here:
1) `claim` throws a lot. That's fine. It's designed to throw if it didn't get any work back, and this response informs the backoff.
2) But if `claim` throws here, we should catch and ignore it and do nothing.
3) Also note that the claim call here does not affect the backoff loop in anyway. It'll claim if the loop is closed. It won't reset or increase the backoff. 
4) When the claimed run is completed, the only impact on the backoff loop is: if the loop has stopped (presumably because we're at capacity), then restart it and make a claim at `minBackoff`. If we have capacity, and the loop is alive, it'll be unaffected by the completion of the run.

I think this is all perfectly okay.

josephjclark:
Confirmed through manual testing that the backoff loop is not affected by a wake up call.

@doc-han I'm gonna make a couple of tiny adjustments and then release :)

packages/ws-worker/src/server.ts
doc-han: error?

packages/lightning-mock/src/api-dev.ts
```diff
@@ -177,14 +176,18 @@ const setupRestAPI = (
     app.enqueueRun(run);
 
     // triggering wakeup in all connected workers
-    if (isWakeUp)
+    if (new Boolean(ctx.query.wakeup)) {
```
doc-han, line 179: This will always be true no matter the string that's passed to wakeup. 
henc, `/run?wakeup=true`, `/run?wakeup=false`, `/run?wakeup=anything` will always make this evaluate true.

doc-han: Oh. wrong

doc-han: actually true. I was overthinking after typing.

josephjclark: Yep - changed it so I can do `/run?wakeup` and save myself 5 characters of typing time

josephjclark: Actually I should do `if ('wakeup' in ctx.query)` because that's really what I want

packages/lightning-mock/src/socket-server.ts
```diff
@@ -227,10 +233,19 @@ function createServer({
         }
       }
     });
+
+    ws.on('close', () => clients.delete(ws));
   });
 
   const mockServer = wsServer as MockSocketServer;
 
+  mockServer.sendToClients = async (message) => {
```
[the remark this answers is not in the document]
josephjclark: Ignore this please :)

packages/lightning-mock/src/api-dev.ts
doc-han: Yeah exactly. else with `/run?wakeup` the Boolean constructor will make it false

packages/ws-worker/test/wake-up.test.ts
```diff
@@ -0,0 +1,108 @@
...
+  return new Promise<void>(async (done) => {
+    // timeout to be sure no claim happened
+    setTimeout(() => {
+      t.pass();
+      done();
+    }, test_timeout - 65);
+
+    const run = getRun();
+    lng.onSocketEvent(e.CLAIM, run.id, () => {
+      t.fail('expected claim received');
```
doc-han, line 74: can you change this wording to `unexpected claim received`

packages/lightning-mock/src/api-dev.ts
josephjclark: Oh I see what you mean :sweat_smile:

packages/ws-worker/test/wake-up.test.ts
doc-han: 🙏🏽

josephjclark: I'll drop a fix on main

PR #880, opened 2025-02-25
Fix worker log level

josephjclark:
This PR fixes an issue with the new `jobLogLevel` option sent from Lightning, which it turns out isn't working ¯\\_(ツ)_/¯

Two problems:

1) The `jobLogLevel` option wasn't getting properly fed through to the engine's options
2) The integration test was accidentally passing because it checked the logger's stdout to see  if the job log was present. But the engine actually doesn't send job logs to stdout, it just forwards them straight to the worker. Confusing!

https://github.com/OpenFn/kit/blob/10ad46888bec8a4333de6481940f54907f8ceac1/packages/engine-multi/src/api/lifecycle.ts#L127-L150

So to fix the test, we instead listen to the log events that Lightning receives. Which is a better test anyway because taht's actually the thing we care about.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/beige-kings-speak.md
+++ b/.changeset/beige-kings-speak.md
@@ -1,0 +1,6 @@
+---
+'@openfn/engine-multi': patch
+'@openfn/ws-worker': patch
+---
+
+Fix an issue where Lightning log level options don't get fed to the engine properly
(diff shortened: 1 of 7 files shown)


midigofrank (changes requested):
I have tested this branch with lightning but still the `JOB` logs are being sent 
EDIT:
Note that I have started the worker using `pnpm start`

josephjclark:
Uh ok yeah it's not working here either :blush:  Let me check..

josephjclark:
@midigofrank It won't build the engine, no

midigofrank:
Aha!! It works!!
 I just ran `pnpm build` and restarted the worker

PR #881, opened 2025-02-25
Release/next

josephjclark:
.

### DIFF
--- a/.changeset/beige-kings-speak.md
+++ b/.changeset/beige-kings-speak.md
@@ -1,0 +1,6 @@
+---
+'@openfn/engine-multi': patch
+'@openfn/ws-worker': patch
+---
+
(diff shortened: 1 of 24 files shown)


josephjclark:
Just want to note that [this publish](https://github.com/OpenFn/kit/actions/runs/13521247215) of worker 1.11.1 contains a **critical** bug

The latest tag fixes it. The bad build was not released to staging or main or anything. So I think we're OK

We just need to run QA on the staging 1.11.1 release that all job logging works as expected.

This commit, made AFTER the first build of 1.11.1, fixed  the problem: https://github.com/OpenFn/kit/pull/881/commits/08210efbabb6b6a90c9e13fabfd9fa0eeccc8b1b

PR #890, opened 2025-03-04
Engine: Limit payload sizes of events

### LINKED ISSUE #888: OOM blow up in the worker
We've  caught a case of the a worker blowing up for OOM

```
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[resource.labels.containerName: global-worker] [R/T] ❯ Final memory usage: [step 482mb] [system 739mb]
[SRV] ℹ 1c87ffb1-e05f-4489-8ee3-9e9e3045108f :: run:log :: OK
[resource.labels.containerName: global-worker] {}
[resource.labels.containerName: global-worker] <--- Last few GCs --->
[resource.labels.containerName: global-worker] {}
[resource.labels.containerName: global-worker] [1:0x7a736db93000] 108822028 ms: Scavenge (interleaved) 1177.3 (1206.6) -> 1176.5 (1214.6) MB, pooled: 0 MB, 89.14 / 0.00 ms (average mu = 0.675, current mu = 0.547) task;
[resource.labels.containerName: global-worker] [1:0x7a736db93000] 108824468 ms: Mark-Compact (reduce) 1203.4 (1234.6) -> 1203.1 (1226.1) MB, pooled: 0 MB, 552.17 / 0.00 ms (+ 1390.0 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 2440 ms) (average mu = 0.544
[resource.labels.containerName: global-worker] {}
[resource.labels.containerName: global-worker] <--- JS stacktrace --->
[resource.labels.containerName: global-worker] {}
[resource.labels.containerName: global-worker] FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
[resource.labels.containerName: global-worker] ----- Native stack trace -----
```

[Cloud link](https://console.cloud.google.com/logs/query;cursorTimestamp=2025-03-03T06:17:53.510584158Z;endTime=2025-03-03T07:28:55.988Z;pinnedLogId=2025-03-03T06:17:43.864463018Z%2Fm33t3g7ldzwa156n;query=labels.%22k8s-pod%2Fapp%22%3D%22global-worker%22%0Atimestamp%3D%222025-03-03T06:17:53.510584158Z%22%0AinsertId%3D%22qiodkwyrvb947cyc%22;startTime=2025-03-03T06:15:55.988Z?project=platform-test-267207)

josephjclark:
## Short Description

This PR fixes an issue where large payloads can cause an OOM exception while being passed from the child process to te main worker, causing the main worker thread to blow up.

Fixes #888

## Implementation Details

This may be a tenporary fix, I'm not sure. It's so hard to debug this area of the engine it's tricky to make progress.

Recall that the architecture is like:
- The worker sits in a process of its own
- It creates a pool of child processes
- Each run then spins out a worker thread to execute inside

Basically what we're doing is:
- Every event that's about to leave the worker thread gets tested
- We stringify it and guess the size
- If we think it's too big, we trim/redact the content before sending it out

So large payloads never leave the worker thread, and so don't get processed downstream. This is definitely how I should have done it in the first place. As you'll see in the TODOs, maybe we should be doing something very different (maybe even using a local socket for data streaming? Not use IPC at all?)

Also  TODO: I'd like a better way to process these IPC events. There must be a better  way to handle large JSON objects. We for sure don't need to keep serializing/deserialising JSON between all the processes - I'd love to add a more efficient encoding in the worker thread, and decode it in the main process. Maybe everything gets dumped into an array buffer, which needs minimal parsing. We might still need to make the event name available - but actually, between the threads, I don't really think so. We should be able to encode it all and do all our analysis in the main thread. I've raised a new issue for this.

## QA

Here's a job to test this in the app:

```
fn((state) => {
  const limit = state.limit || 20;
  state.data = [];
  for (let i = 0; i < limit; i++) {
    state.data.push(new Array(1024 * 1024).fill('a').join(''));
  }
  // log it if you like - it SHOULD be redacted
  //console.log(state
  return state;
})
```
This should create a state object which blows the payload limit and does not get sent back to lightning.

Now, in production, returning too many large objects  from the job can trigger an OOM exception. This is a little hard to reproduce because you need to create something big enough to blow up the worker, but small enough to not be OOM killed by the engine.

Locally, starting the worker like this will trigger an OOM on main, but be stable here:

```
NODE_OPTIONS='--max-old-space-size=100' pnpm start
```

That gives the worker 100mb of memory. On main a 20mb payload is enough to kill it, but it's fine here because a 20mb payload won't be processed in the main thread at all.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


packages/engine-multi/src/util/ensure-payload-size.ts
```diff
@@ -0,0 +1,50 @@
+export const REDACTED_STATE = {
```
josephjclark, line 1: This part 1: this file has been moved from the worker to the engine, and lightly modified.

Note that it selectively validates payload objects so that it can generate an appropriate "fix"

josephjclark:
I've opened out https://github.com/OpenFn/kit/issues/891 to handle more efficient serialisation and comms between processes. I don't want that to block release though.

doc-han:
> Why? The large object should not be loaded into the main worker environment at all. What is requiring all that memory?

Exactly what I was thinking.

josephjclark:
This morning's revelation: remember that the engine is sitting in the main process too. It's not the worker blowing up, its the _engine_

But! The same principle applies. The inner worker thread should not be sending large payloads out to the engine. I can see it here that the engine emits the final workflow-complete event with the redacted state object.

Am I looking at something unrelated? Is 120mb just not enough for basic running of the worker and engine? I don't think so. If I bump the worker memory AND payload size up, we get the same blowup

josephjclark:
Got it: the engine is publishing the full, non redacted state on the internal "resolve_task" event. That's what's causing the blowup now. If I fix/redact that, I should be stable.

PR #903, opened 2025-03-11
Sentry Integration

### LINKED ISSUE #603: Worker: Sentry Integration
We should add Sentry integration to the worker so that we get better alerting of major errors.

Things like channel comms, token auth fails, unexpected errors, autoinstall fails

josephjclark:
## Short Description

This PR adds basic Sentry integration for:

* General server errors (these should be incredibly rare?)
* Errors on Worker-Lightning events (this is the key bit)

Fixes #603

## Implementation Details

I've had a _horrible_ time with this.

Apparently Sentry is fine if you want to  catch exceptions at the top level of your koa web app. But its documentation is dreadful if you want to do anything outside of its standard lanes.

I'm basically worried that when we trap errors, they don't include any useful diagnostic information.

After hours of trial and error I've managed to get to this for an error on `run:start`

image

image

image

See the [OG entry report](https://openfn.sentry.io/issues/6386387564/?project=4508936084848640&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=0)

One specific problem I have is that you can't just pass an Error into the capture function and expect a decent report. Oh no. They also don't give any documentation (that I can find anyway) on what a nodejs error should look like to get good reporting.

## TODO
* Sourcemapping (or disable stack traces)
* Can we do anything about OOM? Likely a new issue
*  Testing (how can we mock sentry and ensure we call it it in the right place? I just wanna test the hooks)
* I am concerned that the scope stuff isn't right. If the worker is running two runs concurrently, is it going to report the correct state? I presume that each koa route will have its own scope (no docs say this anywhere), but I don't know if that works for web sockets. Writing this out I think  I just need to force create a new scope for each execute block

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/ws-worker/package.json
+++ b/packages/ws-worker/package.json
@@ -25,6 +25,7 @@
     "@openfn/lexicon": "workspace:^",
     "@openfn/logger": "workspace:*",
     "@openfn/runtime": "workspace:*",
+    "@sentry/node": "^9.5.0",
     "@types/koa-logger": "^3.1.2",
     "@types/ws": "^8.5.6",
     "fast-safe-stringify": "^2.1.1",
--- a/packages/ws-worker/src/api/execute.ts
+++ b/packages/ws-worker/src/api/execute.ts
@@ -6,7 +6,6 @@ import type {
   RuntimeEngine,
   Resolvers,
   WorkerLogPayload,
-  JobStartPayload,
 } from '@openfn/engine-multi';
 
 import {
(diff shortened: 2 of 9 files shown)


josephjclark:
TODO: see if we can add anything useful around credential load errors

josephjclark:
At the moment, we're not sending sentry sourcemaps, so the stack traces will all report wrong positions.

We can set up a plugin which I believe will upload sourcemaps to sentry's servers. We would need to do this at build time in CI.

Thinking about it, I don't actually have a problem with seeing the minified source in errors. Just show me the compiled JS< it's fine, It's not even minified.  So why are we seeing completely wrong positions? I think it's trying to be smart.

I think this is something to do with our typescript loader and how it reports source positions. In production, we should be running the js directly and so get more accurate error positions

So I'm gonna skip sourcemapping for now. I'm really not keen on it and I don't think its necessary.

josephjclark:
Just got to work out these test fails (they look unrelated but presumably they're not) and then we're good to go :+1:

josephjclark:
Hey @doc-han, do you think you can give this a review sometime tomorrow? Don't look at my history :joy:  Just take a look at where and how sentry is integrated and if you have any thoughts, I'd love to hear them

PR #906, opened 2025-03-14
Logger: ignore empty lines in JSON Mode

### LINKED ISSUE #899: Worker: don't send empty log lines
Don't send empty log lines to Lightning. Do send `console.log(undefined)`, especially in job code, because it might be important.

From discussion at: https://github.com/OpenFn/lightning/pull/2959

* In the job logger - the thing that logs user's console.log statements - I could (probably) detect an undefined value and pass an 'undefined' string instead. Now Lightning will receive message: ['undefined']. I don't know if that will confuse anything. I suppose I could also do message: [undefined] if an argument was passed, which is probably more useful
* The logger has console.break() function, which I think is mostly what the CLI uses for this. I could have a rule like: when logging json output, ignore any console.break statements - because we assume that the JSON will be consumed by something richer that a terminal downstream.
The worker could refuse to send any logs with message: [] on the assumption that it's a waste of everyone's time
If I do that, then I think I can promise:

* message: [undefined] is useful and even important to log
 *You will not be sent meaningless empty lines which are just stdout padding

josephjclark:
## Short Description

In the logger, when running in json mode (ie, emit JSON objects, not strings), ignore anything with an empty message.

Fixes #899

## Implementation Details

This PR is a little misleading because _really_ what it does is prevents the worker sending empty log lines to Lightning.

But the implementation can be handled very neatly by the low level logging component. It'll simply not emit any empty log lines.

This makes sense: json logging is designed for tools to consume logs. And those tools don't need formatting log lines. Usually in the runtime we print empty lines to make the CLI pretty.

Note that `console.log(undefined)` will still log.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/eight-parents-add.md
+++ b/.changeset/eight-parents-add.md
@@ -1,0 +1,5 @@
+---
+'@openfn/logger': patch
+---
+
+In JSON mode, don't log empty messages
--- a/.changeset/old-swans-clean.md
+++ b/.changeset/old-swans-clean.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Ignore empty log lines (don't send them to lightning)
--- a/integration-tests/worker/test/integration.test.ts
+++ b/integration-tests/worker/test/integration.test.ts
@@ -561,6 +561,31 @@ test.serial("Don't send job logs to stdout", (t) => {
   });
 });
 
+// This is a test against job logs - but it should work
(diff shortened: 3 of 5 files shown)


josephjclark:
@stuartc this should do you for empty log lines. Do you want to run any QA against this branch or would you rather we deploy to staging to see?

PR #909, opened 2025-03-23
CLI: collections - Upload data from google sheets

### LINKED ISSUE #902: collections: allow a collection to be set/mapped from a spreadsheet
I want to:
- define a bunch of mappings in a spreadsheet (XLS or gsheets)
- Pull those mappings into the CLI
- Upload them into a collection

This feels like it's something the CLI is very capable of doing.

If the sheet just has two columns, we can just pull them as key/value pairs. Otherwise we need to take a range (or I suppose two columns). We may need to know whether the first two rows are a header and can be ignored (although it's probably harmless to include a header in the map - it's just one redundant key)

The mapping should be a single JSON object uploaded as a key to a single value

So the CLI command is like:
```
openfn collections set my-collection my-key --items pathOrUrlToSheet
```

Maybe we need extra CLI args for "parse items as spreadsheet" or "choose this range" or "ignore headers".

SatyamMattoo:
## Short Description

Allow users to upload data to collections from public google sheets. 

Fixes #902 

## Implementation Details

- Added new keys to collections command 
- Created a function to get and parse data from a public google sheet
- Implementation to handle sheet URL and IDs

## QA Notes

Will add tests once this is complete.

## AI Usage

- [x] Strategy / design
- [x] Optimisation / refactoring
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/cli/src/collections/command.ts
+++ b/packages/cli/src/collections/command.ts
@@ -33,6 +33,13 @@ export type RemoveOptions = CollectionsOptions &
 export type SetOptions = CollectionsOptions & {
   items?: string;
   value?: string;
+  sheet?: string;
+  range?: string;
+  skipHeaders?: boolean;
+  worksheetName?: string;
+  nested?: boolean;
+  valueColumn?: number | string;
+  keyColumn?: number | string;
 };
 
 const desc = `Read and write from the OpenFn Collections API`;
@@ -237,6 +244,61 @@ const items = {
   },
 };
 
+// In your options definitions (collections/command.ts)
+const googleSheetOptions = [
+  {
+    name: 'sheet',
+    yargs: {
+      description: 'Google Sheet URL or ID',
+    },
+  },
+  {
+    name: 'range',
+    yargs: {
+      description:
+        'Range of cells to read from the Google Sheet -  default: A0:B0. First column is the key.',
+      default: 'A1:Z',
+    },
+  },
+  {
+    name: 'skip-headers',
+    yargs: {
+      description: 'Skip the first row of the Google Sheet (headers)',
+      type: 'boolean',
+      default: false,
+    },
+  },
+  {
+    name: 'worksheet-name',
+    yargs: {
+      description: 'Name of the worksheet to use (default: "Sheet1")',
+      default: 'Sheet1',
+    },
+  },
+  {
+    name: 'key-column',
+    yargs: {
+      description: 'Column to use as key (either index or header name)',
+      default: 0,
+    },
+  },
+  {
+    name: 'value-column',
+    yargs: {
+      description: 'Column to use as value (either index or header name)',
+      default: 1,
+    },
+  },
+  // We can have this later if we want
+  // {
+  //   name: 'api-key',
+  //   yargs: {
+  //     description: 'Google API key  for OAuth authentication',
+  //     default: '',
+  //   },
+  // },
+];
+
 const setOptions = [
   collectionName,
   override(key as any, {
@@ -246,7 +308,7 @@ const setOptions = [
   lightningUrl,
   value,
   items,
-
+  ...googleSheetOptions,
   override(o.log, {
     default: 'info',
   }),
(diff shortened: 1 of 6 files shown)


SatyamMattoo:
@josephjclark Hi, this is an initial setup to outline how things might look. Currently, we can retrieve and parse data from the sheet. There are still some changes and validations I need to implement. I’ve created the PR so you can track the progress as well.

Regarding the [@openfn/language-googlesheets](https://www.npmjs.com/package/@openfn/language-googlesheets) package, I encountered some unresolved errors, so I implemented a similar function within the utils as a workaround for now. I will try to use that if there is a similar implementation available.

josephjclark:
Hi @SatyamMattoo - great work!

I'm actually on leave this week so I'll be a bit less active. But I'll take a little look now.

Re the googlesheets adaptor, the idea is to use it as a reference rather than to use it directly. But so long as you have what you need, no problem

packages/cli/src/collections/command.ts
```diff
@@ -237,6 +244,61 @@ const items = {
...
+    },
+  },
+  {
+    name: 'value-column',
+    yargs: {
+      description: 'Column to use as value (either index or header name)',
+      default: 1,
+    },
+  },
+  // We can have this later if we want
```
josephjclark, line 292: Yep - I fear this will be quite hard. I'll ask around and see if anyone has any insight.

packages/cli/src/util/google-sheet-parser.ts
```diff
@@ -0,0 +1,153 @@
...
+ *   - worksheetName: Optional worksheet name.
+ *
+ * @returns A promise that resolves to the mapping object.
+ */
+export async function parseGoogleSheet(
+  sheetId: string,
+  options: Partial<SetOptions>
+): Promise<Record<string, any>> {
+  // Fetch sheet data using our custom function.
+  const sheetData = await getPublicSheetValues(sheetId, options.range || 'A1:Z', options.worksheetName || 'Sheet1');
```
josephjclark, line 87: I'd love to be able to just use the first sheet, if there's only one in the document, regardless of its name. Maybe this is the sort of thing the sheets client can help us with.

packages/cli/src/util/google-sheet-parser.ts
```diff
@@ -0,0 +1,153 @@
...
+ *   - range: The range to fetch (default: 'A1:Z'). (Ignored: we fetch the entire sheet.)
+ *   - nested: If true, creates nested objects from each row (using all columns except the key column).
+ *             If false, uses a single value from a specified value column.
+ *   - keyColumn: The column to use as key (either index or header name).
+ *   - valueColumn: (Required if nested is false) The column to use as value (either index or header name).
+ *   - worksheetName: Optional worksheet name.
+ *
+ * @returns A promise that resolves to the mapping object.
+ */
+export async function parseGoogleSheet(
```
josephjclark, line 82: I'd love to see some unit tests around this - even at this early state - because this is the meat of the implementation.

If I was writing this feature, I'd rename the function to `parseValues()`, pass the values directly as an array (ie, do the sheet loading stuff separately and divorce the "download data from sheets" and "process sheet data steps" entirely). Then I'd write a bunch of unit tests against parseValues with different options and datasets, to really make sure (to myself, my reviewer, and to future developers) that the parsing functionality was working properly.

josephjclark:
Hi @SatyamMattoo  - I think the next step here is to get the sheets client working instead of using rest. Do you need help with that? What was the error you were getting?

SatyamMattoo:
Hi @josephjclark Sorry for the delayed response. I was caught up with the work, could not give time to this. I will update you this weekend regarding this (will try to make things working along with the sheets client). I was trying to use the adaptor itself, that might be causing the error. I'll update the function with the OAuth thing, will be easier to debug and will hopefully work as well.

josephjclark:
@SatyamMattoo yep, don't use the adaptor directly. Copy code out of its useful. Documentation on the sheets client is dreadful, but usage in the adaptor should have everything you need really

PR #928, opened 2025-04-17
worker: by default, exclude references from statePropsToRemove

### LINKED ISSUE #487: Runtime: remove state.references
Many adaptors write old state.data objects into `state.references`

openfn v1 does not remove this from the final state, but encourages users to do so manually because it's basically just bloat.

https://docs.openfn.org/documentation/build/triggers#a-quick-fix-for-final-state-bloat

Refernces is useful inside a job but probably not outside it (and references from job a presumably should not be passed to job b).

The new runtime does not do any management of references.  Should it? Jjust like it removes state.configuration.

josephjclark:
## Short Description

Simple PR to update the default worker to remove the `references` key from state at the end of each step. This closes #487

I've also updated the environments in deployment

Update: also don't exclude `reponse`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/eight-tips-admire.md
+++ b/.changeset/eight-tips-admire.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Update default statePropsToRemove value to exclude references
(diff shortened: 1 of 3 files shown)


josephjclark:
hmm, as Taylor has flagged, we need to be careful about this because 233 production workflows use `references`. How will they be affected?

Issue #621, opened 2024-03-05
CLI: support execute from workflow.yaml


josephjclark:
There are currently two workflow formats (which is a bit confusing in itself!). `workflow.json` is natively supported by the CLI and is a JSON representation of a workflow. There's also `workflow.yaml`, used by `deploy` and Lightning, which is a yaml representation of the same thing.

We should really add an interpretation layer to the top of the CLI which can read in a workflow.yaml and convert it to a "json" internal representation.

This would enable the CLI to execute directly from a project folder, which really is the first thing you expect when you do `openfn pull <myproject>`.

utnim2:
Working on this @josephjclark !
can i get assigned with it ?

josephjclark:
Hey @utnim2, thanks for speaking up! Are you coming from the Code for Govtech DMP programme?

utnim2:
> Hey @utnim2, thanks for speaking up! Are you coming from the Code for Govtech DMP programme?

Yes @josephjclark

utnim2:
i want to solve the issue in order to understand the codebase more not as a part of DMP
@josephjclark

josephjclark:
Hi @utnim2. Ok, that sounds like a good idea! But to be absolutely clear - this wouldn't contribute to or influence your DMP project.

This issue is not very well specced. I'll add some more details to you in a few hours and tag you here.

It shouldn't be a big project. If the input file is a yaml file, convert it to a JSON format. I'm not actually sure how complex the conversion will be, but I'll link you to the type definitions/specifications you'll need.

josephjclark:
Hi @utnim2, I've added more details to the issue and assigned you. May the force be with you :vulcan_salute:

josephjclark:
I've unassigned @utnim2  from this issue, so it is now open to contributors.

Akshanshkaushal:
Hi @josephjclark To better understand the codebase, I want to solve this issue. Could you please assign me this?

josephjclark:
Hi @Akshanshkaushal  - thanks for the interest! 

It would be a good idea for you to reply here with a rough solution, just enough to suggest you understand the problem and how best to solve it. In particular I'm interested in what manual and automation tests you'll use and how you'll prove your final solution works.

Akshanshkaushal:
All right, @josephjclark. It will be carried out in the manners listed below:

1>User Input: The user specifies the path to the project.yaml file and, optionally, the ID of the specific workflow they want to execute.

[...]

PR #929, opened 2025-04-23
Projects & New sync phase 1

josephjclark:
This PR adds a new `@openfn/project` package and interfaces it to the CLI:
* Use `openfn pull --beta` to use new projects and expand to the file system
* Use `openfn deploy --beta` to use new projects and deploy on lightning
* `openfn execute` now works from the new project.yaml file

## Projects Package

The basic idea is that you can read in an openfn project (or a repo of projects) from any source - the file system, a project.yaml file - into a standard js representation.

That has a few benefits:
1) It's a common base for us to sync changes between projects. No need to sync 4 yaml & json files together - just compare the two projects models, diff them (as JSON), and push state from on to the other
2) This opens up the ability to serialize projects to disk with a nicer, cleaner, leaner format

I'm really hoping to get all the project mappings done over a couple of days, then we can get into the more interesting syncing and versioning issues.

## Docs

I've added sketchy documentation to a wiki page: https://github.com/OpenFn/kit/wiki/Pull-Deploy-Beta

This is linked from `openfn deploy --help`

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI


josephjclark:
Great progress on this today: I can now pull a state file from the provisioner, convert it to a project model (which includes workflows compatible with the runtime, so it should be executable), and convert it back to the provisioner state json without losing anything.

We're moving quickly towards this:
image

A project model can be parsed from, or serialized to, any of these data structures. The project is the key.

I'm also sketching out subsets of properties - like here is app metadata (uuids, updated_at), here CLI metadata (last_synced_at), here are app settings (credentials, options). When syncing any two projects, we can choose to include or exclude these buckets of props. So "sync without credentials" becomes a pretty easy thing to do.

Next step is getting the  filesystem output spiked.

josephjclark:
I've set up the filesystem serializer. So now when you `openfn pull`, by default the project will be expanded to disk. The CLI command is still very crude but the serialiser seems to work nicely.

josephjclark:
Fiddly afternoon trying to work out the names of things mostly. Improved loading from the file system a bit.

Should plug in yaml and deploy support tomorrow :crossed_fingers:

josephjclark:
We can now serialie project artifacts to json or yaml

image

Internally we handle everything as a JSON data structure, then at the start of an import or the end of an export, we convert it to/from yaml. So the yaml layer is now super lightweight.

We should be able to handle JSON diffing without caring about key order. I may at some stage see if I can optimise key order for readability - ie always ensure id and name are the first keys. But that's nicity.

Next step is better CLI integration. Pull is sort of done (very roughly!) but I want to get deploy and execute working too.

I expect to put out a test release soon - but my priority is to validate provisioner compatibility, diffing and versioning as soon as possible. Those are the risky - and valuable - bits.

josephjclark:
Great milestone today: managed to pull a workflow, make a (trivial) edit, and deploy it back to the app.

Tomorrow I'll tidy up pull, deploy and execute and see if I can get them ready for a beta release (I expect loads of issues but early feedback is always helpful)

josephjclark:
We can now execute from a local project - like if you do `openfn my-workflow` it'll find the workflow called `my-workflow` and run it.

We can also execute a workflow.yaml directly.

I've decided not to support execute from the older deploy artifacts - project.yaml and state.json - at least for now. Maybe we'll provide a migration tool later.

I've left quite a lot of mess behind me. I'll tidy up a bit this afternoon but then I need to work out whether to:
1. Release as-is
2. Do some testing and fixing around tracking local file name changes - there are likely to be some fs <-> project sync issues that need resolving
3. Start looking at versioning or credentials

I also know I want a really strong sync test suite - probably in integration tests. I'm not sure when to start that - I think I still need to prove out a couple of key ideas

josephjclark:
Tidying and fixing and documenting today. I might actually be able to release this soon

josephjclark:
We can now sort of execute from a project file, see #621

The caveat is that it only works (so far) on new project formats, not the old ones (that's in #935 and I don't think we should bother)

josephjclark:
Bit of QA with openfn:

* `openfnx` will throw when trying to output if the dir doesn't exist. That might be because the output path is wrong? yes I think so
* When pulllng, the project name is not written to `openfn.yaml`. It should be at the top

PR #932, opened 2025-04-30
Engine: encourage better garbage collection

### LINKED ISSUE #826: engine: memory leak
Worker memory seems to steadily increase over time:

Image

When idle, the worker's memory should really drop down to base levels.

josephjclark:
## Short Description

This PR fixes an issue in the engine where state references are not garbage collected, and so memory leaks over time.

This is a probably cause of at least some lost runs.

Fixes #826

## Implementation Details

After a heck of a lot of investigation I've worked out that the suspect memory leak flagged by @taylordowns2000 is real. I've also worked out that it comes from the engine, and if you cut out child processes and the runtime and just statically return a large object from each run, the leak still occurs

What seems to be happening is:
- for every run we build a context object, which tracks a bunch of data for and about that run
- This is designed to be a temporary object, dropped as soon as the run completes
- When a run completes, we take the final state object returned by the workflow and write it to this context
- And for whatever reason, this object does not get unreferenced and so does not get garbage collected

The fix - the dumbest, easiest fix I can find - is to force that object to be unreferenced after the workflow has finished.

I've added a new `test:mem` script to the engine. This creates an engine with quite a low memory limit, and then infinitely runs a job with a large-ish payload.

Against `main` the test fails (heap error) on my machine after about 20 runs.

At the time of writing, on this branch, I've just crossed ~2000~ ~3000~  ~4000~ 5000 runs (my poor CPU!)

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/integration-tests/worker/CHANGELOG.md
+++ b/integration-tests/worker/CHANGELOG.md
@@ -1,5 +1,14 @@
 # @openfn/integration-tests-worker
 
+## 1.0.85
+
+### Patch Changes
+
+- Updated dependencies
+  - @openfn/engine-multi@1.6.5
+  - @openfn/ws-worker@1.13.4
+  - @openfn/lightning-mock@2.1.7
+
 ## 1.0.84
 
 ### Patch Changes
--- a/integration-tests/worker/package.json
+++ b/integration-tests/worker/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@openfn/integration-tests-worker",
   "private": true,
-  "version": "1.0.84",
+  "version": "1.0.85",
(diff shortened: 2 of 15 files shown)


josephjclark:
There are two alternatives here, both better:
1) do we even need to write the final state to the run context? Surely all we need to do is emit it?
2) `WeakMaps` don't have such strong references and are more likely to be garbage collected. This goes a bit beyond my computer science skill - but maybe if we used a weakmap for the run context, it'll be better cleaned up after a run has finished? See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap

PR #947, opened 2025-06-26
Credential timeouts

josephjclark:
Investigating what happens when credential load times out. It looked for a while like it might trigger a lost but after plugging in these tests, I don't think so.

There is no diff and no release here yet - just a few extra usage tests

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/ws-worker/test/reasons.test.ts
+++ b/packages/ws-worker/test/reasons.test.ts
@@ -239,7 +239,10 @@ test('exception: failed to load credential', async (t) => {
 
   t.is(reason.reason, 'exception');
   t.is(reason.error_type, 'CredentialLoadError');
-  t.is(reason.error_message, 'Failed to load credential zzz: err');
+  t.is(
+    reason.error_message,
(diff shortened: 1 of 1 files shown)


packages/ws-worker/test/reasons.test.ts
```diff
@@ -236,7 +239,27 @@ test('exception: failed to load credential', async (t) => {
 
   t.is(reason.reason, 'exception');
   t.is(reason.error_type, 'CredentialLoadError');
-  t.is(reason.error_message, 'Failed to load credential zzz: err');
+  t.is(
+    reason.error_message,
+    'Failed to load credential zzz: [fetch:credential] Error: err'
```
josephjclark, line 244: I don't understand this diff... nothing has changed here, so why is this error message different (and worse?)

josephjclark: It think it's just changed how the event bubbles out of the mock. I'm OK with it. There's no diff in prod here.

PR #950, opened 2025-06-27
log lightning error objects

### LINKED ISSUE #948: worker: errors coming from lightning events are objects, not strings
When Lightning rejects an event, it sends a payload like this:

```
{
  errors: {
    [job-id]: ["field cannot be blank"]
   }
}
```
I think this has changed (probably a long time ago) because the error payloads used to be simple strings `not_found`

We need to update error handlers in the worker to be able to handle these error payloads. I think it's as simple as stringifying them.

This is particularly important on credential errors.

An immediate fix is to just dump the payload, but a better one would be to unwrap the message into a nice string and log it.

Later, we might want to feed these errors back into the fail reason returned by `run:complete`

midigofrank:
## Short Description

Log whatever error lightning sends back, including objects.

Fixes #948 

## Implementation Details

Just `JSON.stringify`

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/.changeset/purple-forks-drum.md
+++ b/.changeset/purple-forks-drum.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Improve logging of errors returned by lightning
--- a/integration-tests/worker/test/exit-reasons.test.ts
+++ b/integration-tests/worker/test/exit-reasons.test.ts
@@ -113,6 +113,10 @@ test('exception: bad credential (not found)', async (t) => {
     ],
   };
 
+  const expectedLightningErrorResponse = {
+    errors: { id: ['Credential not found!'] },
+  };
+
   const result = await run(attempt);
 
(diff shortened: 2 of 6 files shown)


josephjclark (changes requested):
Would love to see some tests against some common errors!

- a string
- an actual lightning error object
- maybe an array

I was really hoping as part of this work to update the lightning mock so that the errors look more realistic. We   could also afford to do some nicer formatting here if we're expecting certain structures back from lightning - because I think the raw errors are going to be a bit hard to read? But we can merge this as a quick fix and coming back to prettify things later

josephjclark (approved):
Thank you @midigofrank, I'll try and get this into a release tomorrow

PR #984, opened 2025-09-03
Implement merge algorithm for workflows

doc-han:
## Short Description

Mapping nodes in two workflows

Fixes #

## Implementation Details

This is a simple write up on how we merge two workflows with an attempt to keep reference to nodes in the target workflow. Basically, we try to figure out which nodes still exists in both workflows.

At the end, we output both the node mappings and the edge mappings like below.
- **nodes**: source node id → target node uuid
- **edges**: source edge key → target edge uuid

## How we find a match for each node

We have several steps for finding suitable matches for nodes by picking up the obvious ones first before the less obvious ones.

Note: We keep a pool of unmapped nodes on both sides, the source and target.

1. **Direct ID matching**
    - If a source node has the same ID as a target node, we map it immediately.
    - This removes both from the unmapped pools and reduces ambiguity for the remaining nodes.

2. **Root matching**
    - We find the roots of both workflows and immediately map them. these don't need to be contested for.

3. **Structural matching (parents, then children)**
    - Parent matching: if a source node's parents are already mapped, we look for target candidates that have those same (mapped) parents.
    - Children matching: if a source node's children are already mapped, we prefer target candidates whose children line up with those mapped children.
    - Shortlist: The parent/child matching produce a shortlist. If that shortlist contains exactly one candidate, we accept it immediately and map that node.

4. **Expression/content check**
    - If structural mathcing didn't uniquely identify a candidate, we compare expressions.
    - When a candidate has the same expression, that is a strong signal. If it produces a single clear candidate, we accept it.

**Approximating a Structural pick**
If the shortlist after parent/child matching has more than one candidates and expression matching doesn't settle the tie, we pick the best candidate in the shortlist but not immediately. The approximated candidate is picked on our last loop. (reason explained below)

## Why we do several passes (the loop)

We make multiple passes. After each pass, more nodes become mapped, which makes their neighbours less ambiguous on the next pass. We keep doing this until the size of the unmapped array doesn't decrease, meaning no more unambiguous nodes can be mapped.

**On each pass:**
- Consider only nodes that are still unmapped.
- Apply the same order of checks: parents → children → expression.
- As soon as a node has a single clear match, map it and remove it from later passes.

**On the final pass only:**
- We do one final pass after the unmapped array can't be reduced in size. meaning all the unambiguous nodes have been mapped.
- If expression still doesn't disambiguate, but structure produced a best candidate, we accept the best structural match.

## How edges are mapped

Once node IDs are mapped, we translate each source edge parent -> child to the equivalent edge in the target:
- Build the source edge key as `<parent-id>-<child-id>`.
- Replace parent/child IDs with their mapped target IDs. that is `<mapped-parent-id>-<mapped-child-id>`
- Look up the corresponding edge in the target and store its UUID.

With this, we're able to do a better edge mapping making use of the mapping knowledge we have during the node mapping. Hence, edge mapping can only be done after node mapping.

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/packages/project/test/merge/map-uuids.test.ts
+++ b/packages/project/test/merge/map-uuids.test.ts
@@ -409,3 +409,35 @@ test('move: children move to a sibling', (t) => {
     'o-e': target.getUUID('b-e'),
   });
 });
+
+test('expression-based mapping: nodes only distinguishable by expression', (t) => {
+  const source = generateWorkflow(['trigger-x', 'trigger-y']);
+  const target = generateWorkflow(['trigger-a', 'trigger-b']);
+
+  source.set('x', {expression: 'foo'});
+  source.set('y', {expression: 'bar'});
+  target.set('a', {expression: 'foo'});
+  target.set('b', {expression: 'bar'});
+  const result = mapUUIDs(source, target);
+
+  t.deepEqual(result.nodes, {
+    trigger: target.getUUID('trigger'),
+    x: target.getUUID('a'),
+    y: target.getUUID('b'),
+  });
+
+  // switch the expressions and expect mappings to switch
+  // from x -> a, y -> b to x -> b, y -> a
+  source.set('x', {expression: 'bar'});
+  source.set('y', {expression: 'foo'});
+  const sresult = mapUUIDs(source, target);
+
+  t.deepEqual(sresult.nodes, {
+    trigger: target.getUUID('trigger'),
+    x: target.getUUID('b'),
+    y: target.getUUID('a'),
+  });
+});
+
+// we need to do some testing around multiple roots. 
+test.skip('multiple roots:', ()=> {});


packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +39,178 @@ export interface MappingResults {
...
-  // const mapping: Record<string, MappingRule> = {};
+  const idMap = new Map<string, string>();
+  // NODE MAPPING
+  for (const target_step of target.steps) nodeMapping[target_step.id] = null;
+  for (const source_step of source.steps) {
+    if (!source_step.id) continue; // yh. we'll always have it.
+    const sampleSteps: Workflow['steps'] = target.steps.filter(
+      (step) => !idMap.has(step.id)
+    );
+    let result: Workflow['steps'] = [];
```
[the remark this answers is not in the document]
josephjclark:
```
const pool = target.steps.slice()

// now inside your main loop, use `pool` instead of `target.steps` and `simpleSteps`
```

josephjclark:
Agreed on meeting: let's do one loop for the id mapping (the solid nodes), then one loop with various filters for everything else.

The algorithm in the issue  needs updating to reflect this, because Lightning needs to implement exactly the same logic

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +40,234 @@ export interface MappingResults {
...
 
-  // sets all nodes as not existing in target
-  const targetIds: Record<string, string> = {};
-  for (const tstep of target.steps) {
-    nodeMapping[tstep.id] = null;
-    targetIds[tstep.id] = tstep.openfn?.uuid || tstep.id;
+  // Map by id
+  let {
+    mapping: nodeMapping,
+    pool,
```
josephjclark, line 48: the way the pool is handled here - splitting into nodes and edges - makes me think my algorithm isn't working. This is important!

It's totally fine if the algorithm doesn't work, but I need to know so that we can fix it and we can share it with the other team

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +40,234 @@ export interface MappingResults {
...
-    for (const [toNode, toValue] of Object.entries(next)) {
-      const edgeId = tstep.id + '-' + toNode;
-      edgeMapping[edgeId] = null;
-      targetIds[edgeId] =
-        typeof toValue === 'object' ? toValue.openfn?.uuid || edgeId : edgeId;
+  for (const source_step of pool.source) {
+    if (!source_step.id) continue; // yh. we'll always have it.
+
+    // these are the candidates for the search
+    let candidates = pool.target.filter((step) => !idMap.has(step.id));
```
josephjclark, line 56: I think think nodes should already have been removed from the pool, doing away with this step

packages/project/test/merge/map-uuids.test.ts
```diff
@@ -71,3 +71,251 @@ test('one connecting node missing', (t) => {
...
+
+// NEW TESTS
+
+// narrowing by parent
+// id changed but parent stayed the same
+test('parent stayed same but id changed', (t) => {
+  const source = generateWorkflow(['a-b']);
+  const target = generateWorkflow(['a-x']);
+
+  const result = mapUUIDs(source.workflow, target.workflow);
```
josephjclark, line 239: Wait, shouldn't we be passing the `Workflow` class directly into mapUUIDs? `Workflow` should have utility functions that can help us

packages/project/test/merge/map-uuids.test.ts
```diff
@@ -71,3 +71,251 @@ test('one connecting node missing', (t) => {
...
+  t.deepEqual(result.nodes, {
```
josephjclark, line 240: Love the structure of this :ok_hand: :kiss: 

Obviously you can simplify:
```
t.deepEqual(result.nodes, {
    a: target.getUUID('a'),
    b: target.getUUID('x'),
    x: null,
});
```

packages/project/test/merge/map-uuids.test.ts
```diff
@@ -71,3 +71,251 @@ test('one connecting node missing', (t) => {
...
+    ['a']: target.getUUID('a'),
+    ['b']: target.getUUID('x'),
+    ['x']: null,
```
josephjclark, line 243: These nulls don't look right - why are they in the mapping?

Null means "remove this node" - but I'm not yet convinced that that's necessary (Lightning's API can work out what to remove, if needed), but also node `x` hasn't actually been _removed_ exactly, it's just been renamed

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +40,234 @@ export interface MappingResults {
...
+  source: Workflow['steps'],
+  target: Workflow['steps']
+): MapStepResult {
+  const targets: Record<string, Workflow['steps'][number]> = {};
+  const mapping: Record<string, string> = {};
+  const idMap = new Map<string, string>();
+
+  for (const target_step of target) {
+    targets[target_step.id] = target_step;
+    mapping[target_step.id] = null;
```
josephjclark, line 177: I think I'm wrong about this bit - Lightning might have to worry about removal, but looking at it today I'm sure that we _don't_

packages/project/test/merge/map-uuids.test.ts
doc-han: that's correct. resolved

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +40,234 @@ export interface MappingResults {
...
+    // If none matched, mark as new
+    nodeMapping[source_step.id] = true;
   }
 
-  for (const sstep of source.steps) {
-    const ex = nodeMapping[sstep.id];
-    // matched nodes are mapped
-    if (ex === null) nodeMapping[sstep.id] = targetIds[sstep.id];
-    else if (ex === undefined) nodeMapping[sstep.id] = true; // true to create id for new nodes
+  // EDGE MAPPING
```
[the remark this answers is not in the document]
doc-han: The `getEdges` function returns `Record<string, string[]>` representing `<parent, children[]>`.
So for a single parent `trigger` with children `a` and `b`. The output will be `{"trigger": ["a", "b"]}`. 

This is because I didn't want to use an array for faster lookups and also we have the children being an array because we can't have multiple keys of the same value in an object. `{"trigger": "a", "trigger": "b"}` wouldn't work.

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +39,178 @@ export interface MappingResults {
...
 
 export default (source: Workflow, target: Workflow): MappingResults => {
+  // create nodes, edges
   const nodeMapping: Record<string, MappingRule> = {};
   const edgeMapping: Record<string, MappingRule> = {};
-  // const mapping: Record<string, MappingRule> = {};
+  const idMap = new Map<string, string>();
+  // NODE MAPPING
+  for (const target_step of target.steps) nodeMapping[target_step.id] = null;
+  for (const source_step of source.steps) {
```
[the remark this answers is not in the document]
doc-han: with rules 1. and 2. enforced, we'll never end up in an infinite loop

packages/project/src/merge/map-uuids.ts
```diff
@@ -8,114 +8,327 @@
...
 };
 
-// if newchildren has some additional nodes, that's fine. it's the same children!
-// when a node is missing, nope, not the same children
-function sameChildren(mainChildren: string[], newChildren: string[]) {
-  if (newChildren.length < mainChildren) return false;
-  for (const child of mainChildren) {
-    if (!newChildren.includes(child)) return false;
+// util for getting uuid from and edge
+function getEdgeUuid(
```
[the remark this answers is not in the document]
doc-han: Removed in favour of the indexing inside the workflow implementation?

doc-han:
> TODO: should we sort the nodes top-down breadth first?
> 
> Does it matter at this point?

I doubt it matter at this point since we do iterate over the nodes enough to clear all the unambiguous ones.

packages/project/src/merge/map-uuids.ts
```diff
@@ -39,82 +39,178 @@ export interface MappingResults {
...
-    const next =
-      typeof sstep.next === 'string'
-        ? { [sstep.next]: true }
-        : sstep.next || {};
-    for (const toNode of Object.keys(next)) {
-      const edgeId = sstep.id + '-' + toNode;
-      const ex = edgeMapping[edgeId];
-      if (ex === null) edgeMapping[edgeId] = targetIds[edgeId];
-      else if (ex === undefined) edgeMapping[edgeId] = true;
+  for (const [parent, children] of Object.entries(sourceEdges)) {
```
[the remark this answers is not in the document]
doc-han: I don't think it can be made simpler. an edge is `<parent, children>` then we iterate for each <parent, child>

PR #999, opened 2025-09-17
feat: Project level merging

### LINKED ISSUE #973: sandbox: merge two projects
Farhan is currently working on this. Outstanding tasks:
- Merge simple "solid nodes"
- Make merge algorithm more easily shared
- Crate tests we can re-use in lightning (at the yaml level - merge two yaml files and get a third yaml back)

Maybe for now we should console.warn every time we create a new uuid, just to communicate that ids may not track

Acceptance criteria:
- We can basically merge two projects (so long as node names or ids don't change)
- We can share the logic of the merge with the Lightning team
- We can share a bunch of tests with the Lightning team

doc-han:
## Short Description

Currently, supports workflow mappings provided by the user. 

Fixes #973 

## Implementation Details

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`


packages/project/test/util/base-merge.test.ts
```diff
@@ -0,0 +1,149 @@
+import baseMerge from '../../src/util/base-merge';
+import test from 'ava';
+
+test('full merge', (t) => {
```
josephjclark, line 4: This is an excellent suite of tests, thank you!

packages/project/test/util/base-merge.test.ts
```diff
@@ -0,0 +1,149 @@
...
+  const result = baseMerge(target, source, ['key', 'props', 'colours']);
+
+  t.deepEqual(result, {
+    key: 'two',
+    title: 'target',
+    colours: ['green', 'yellow'],
+  });
+});
+
+// TODO: should a merge create a union of two objects?
```
josephjclark, line 125: No, I don't think so. Keep it to a shallow merge

packages/project/test/util/base-merge.test.ts
```diff
@@ -0,0 +1,149 @@
...
+test('special: dealing with objects', (t) => {
```
josephjclark, line 126: Again, test name needs to explain the behaviour: objects are shallow merged

packages/project/src/merge/merge.ts
```diff
@@ -2,9 +2,11 @@ import { Workflow } from '@openfn/lexicon';
 import { Project } from '../Project';
```
josephjclark, line 2: This file should probably be called `merge-project` then?

packages/project/src/merge/merge.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
  * target, but the UUIDs of the source
  */
 // TOOD what if a workflow is removed from the target?
```
josephjclark, line 22: This TODO needs resolving (or just removing if it's already done)

packages/project/src/merge/merge.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
...
-  for (const workflow of target.workflows) {
-    const sourceWorkflow = source.getWorkflow(workflow.id);
-    if (sourceWorkflow) {
-      const mappings = mapUuids(sourceWorkflow, workflow);
-      finalWorkflows.push(mergeWorkflows(sourceWorkflow, workflow, mappings));
-    } else finalWorkflows.push(workflow);
+  const usedTargetIds = new Set<string>();
+
+  for (const sourceWorkflow of source.workflows) {
+    const mappedTargetId = options.workflowMappings?.[sourceWorkflow.id];
```
josephjclark, line 32: You can simplify all this to:
```

const mappedTargetId = options.workflowMappings?.[sourceWorkflow.id] ?? sourceWorkflow.id;
const targetWorkflow = target.getWorkflow(mappedTargetId)
```

OR even

```
const targetWorkflow = target.getWorkflow(options.workflowMappings?.[sourceWorkflow.id] ?? sourceWorkflow.id)
```

packages/project/src/merge/merge.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
  * target, but the UUIDs of the source
  */
 // TOOD what if a workflow is removed from the target?
-export function merge(source: Project, target: Project, options) {
+export function merge(
+  source: Project,
+  target: Project,
+  options: Options = { workflowMappings: {} }
```
josephjclark, line 26: As well as mappings, we need an option for "only transfer these workflows from the source"

packages/project/src/merge/merge.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
...
+      finalWorkflows.push(sourceWorkflow);
+    }
   }
 
-  const mergedProject = new Project(
-    { ...source, ...target, workflows: finalWorkflows },
-    // TODO probably just preserve target repo?
-    { ...source.repo, ...target.repo }
+  // workflows from target that didn't get merged
+  for (const targetWorkflow of target.workflows) {
```
josephjclark, line 49: Yeah need to be careful here because we need allow workflows to be removed

I would say:

- When merging everything, then delete all target workflows and only keep the source workflows
- When merging partially, allow a null value to be an explicit delete - but otherwise ignore the other workflows

packages/project/src/merge/merge.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
...
+
+  // TODO: clarify repo preservation strategy
+  // TODO: how other properties of a project are being merged.
+
+  // with project level props merging, target goes into source because we want to preserve the target props.
+  return new Project(
+    baseMerge(
+      source,
+      target,
+      ['description', 'history', 'name', 'meta', 'options', 'repo', 'openfn'],
```
josephjclark, line 63: IIRC repo shouldn't get merged, because that has stuff like "what is the remote app name". And we don't want to override that bit

doc-han: quick question. when merging at the project level. don't we just merge the workflows and leave the other properties as is? `history`, `description`...

josephjclark: 3 days to reply to a quick question :joy: 

Yes, I think `workflows` is the only thing to merge when handling projects.

If users want to change the original project name (or another sandbox name), they have to edit it directly. You can't _merge_ that change, it doesn't make sense.

Most (if not all) options need to be updated directly and also aren't merged. 

Oh, we do need to track `collections`. If you add a collection to a sandbox then that will need to be merged across (shallow merge).

This possibly opens up an issue (for me). Right now a Project doesn't have a version. So if you compare two projects and the only difference is that one has added a collection, it'll count as no difference. That means either we need a project version hash, or we need special diffing logic for Projects

doc-han: okay. so if I'm getting it right.
1. when merging on the project level we merge just the workflows
2. with collections, if sandbox has collection, then we pick that. 

with 2. what if both sandbox and main have collection? we keep sandbox priority?

packages/project/src/merge/merge-project.ts
```diff
@@ -18,28 +20,48 @@ type Options = {
...
+      finalWorkflows.push(sourceWorkflow);
+    }
   }
 
-  const mergedProject = new Project(
-    { ...source, ...target, workflows: finalWorkflows },
-    // TODO probably just preserve target repo?
-    { ...source.repo, ...target.repo }
+  // workflows from target that didn't get merged
+  for (const targetWorkflow of target.workflows) {
```
doc-han: So the options are 3.

1. keep all unmapped workflows (both main and sandbox)
2. keep unmapped workflows in target (main)
3. keep unmapped workflow in source (sandbox)

josephjclark (approved):
Approved subject to #1006, which has a few suggestions

Issue #997, opened 2025-09-15
worker: memory leak


josephjclark:
Issue to track the second memory leak the worker

We solved one some time ago, and set up some mitigation in GKE to kill workers frequently and stop memory rising.

But we are still seeing Lost runs as a result of the leak.

Tracking findings and solutions in this issue.

josephjclark:
Here's some long running output, posting one run every 30 seconds
```
[SRV] ❯ workflow run-1 complete in 0.779s (114.15mb heap) releasing worker thread
[SRV] ❯ workflow run-2 complete in 2.936s (74.33mb heap) releasing worker thread
[SRV] ❯ workflow run-4 complete in 10.241s (91.18mb heap) releasing worker thread
[SRV] ❯ workflow run-3 complete in 40.287s (91.25mb heap) releasing worker thread
[SRV] ❯ workflow run-5 complete in 48.26s (91.81mb heap) releasing worker thread
[SRV] ❯ workflow run-6 complete in 20.201s (72.61mb heap) releasing worker thread
[SRV] ❯ workflow run-7 complete in 19.593s (72.62mb heap) releasing worker thread
[SRV] ❯ workflow run-8 complete in 23.375s (72.64mb heap) releasing worker thread
[SRV] ❯ workflow run-9 complete in 84.488s (111.56mb heap) releasing worker thread
[SRV] ❯ workflow run-10 complete in 57.992s (82.82mb heap) releasing worker thread
[SRV] ❯ workflow run-11 complete in 71.783s (141.79mb heap) releasing worker thread
[SRV] ❯ workflow run-12 complete in 78.102s (141.92mb heap) releasing worker thread
[SRV] ❯ workflow run-13 complete in 87.199s (102.08mb heap) releasing worker thread
[SRV] ❯ workflow run-15 complete in 116.523s (111.74mb heap) releasing worker thread
[SRV] ❯ workflow run-14 complete in 149.33s (111.79mb heap) releasing worker thread
[SRV] ❯ workflow run-16 complete in 180.181s (121.95mb heap) releasing worker thread
[SRV] ❯ workflow run-18 complete in 142.575s (161.02mb heap) releasing worker thread
[SRV] ❯ workflow run-23 complete in 217.14s (141.37mb heap) releasing worker thread
```

I see this pattern consistently under different setups. Some observations:
* The first run is always super quick
* Memory goes up and down but does stay quite high. It's worse with more frequent runs
* Why does the execution time show down so much?? Is this related?

[...]

PR #1003, opened 2025-09-23
engine: move compilation down into the worker thread

### LINKED ISSUE #1002: worker: run the compiler in an isolated thread
Compilation currently happens in the main worker thread. There are a couple of problems with this:
1. Compilation can take a lot of memory and CPU, and so increase the memory footprint and decrease performance of the main worker thread
2. If the compiler chews up too much memory simply because it's parsing a massive source file, it could throw a heap error and blow up the main worker

If we were to isolate compilation to a different thread, we'd be more stable. This has sort of been the plan forever but there's never been any urgency on it.

There are two ways we can do this:

1. compile inside the run thread. This means that each run has to load the typescript engine, which is a bit of a bummer. But means any compilation issues are totally isolated. An OOM in the compiler means an OOM for the job, which is just cleaner. It also means we can compile steps just-in-time, rather then pre-compiling as we do. Does this have implications for loading from the adaptor registery for `preloadAdaptorExports`?
2. compile in a dedicated worker thread. This means we still pre-compile, and we have to set some kind of compile memory limit. If a run takes say more than 100mb to compile, we have to crash it with an OOM state. This means we only have to load the TS environment once, and could even do some in-memory caching (athough that'll only increase load on memory so maybe that's not actually a good thing)

Both feel like a lot of work. Both would cut out any lost runs arising from heap errors in the compiler. Neither would address the memory leak

josephjclark:
UPDATE: I've reduced the compiler overhead and got the word from Taylor that users should NOT be affected by these changes due to their existing limits. So it should be fine.

This PR moves compilation out of the main worker thread and into the execution threads.

This may work around #997 and closes #1002

I generally like the architecture but maybe it's not the clearest win ever

### Pros

- Main thread is isolated from compilation CPU and memory usage
- You'll get OOM errors if you try and compile a step that's too long (but how would you fix this?)
- A compilation heap error won't take out the whole worker, so fewer lost runs (I hope!)
- We ensure no TS caching between steps
 
### Cons

- The typescript engine is now loaded into each child thread, increasing its memory overhead and likely decreasing overall performance. This is improved (but not entirely fixed) by  #1012
- We ensure no TS caching between steps

### Future work

- We should let the runtime compile expressions just-in-time. This need an explicit pre-execute callback (or something) in the runtime. It would mean that code that's never executed is never compiled, so you can have steps with broken code without blowing up the runtime entirely. They'll be safely ignored. This could be argued to be a good or bad thing.
- I'd actually love to drop typescript from the compiler - we only need it to  calculate exports, and there's surely  better way to handle this (maybe even reflection?). That would reduce a huge overhead from the compiler's runtime

### Lightning

As a consequence of this work, compilation logs will start appearing in Lightning:
image

Users may also start finding that large jobs throw memory errors while trying to  compile. If so I'm not sure what to do about this it's sort of fair enough and guarantees that the run won't be lost

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/ws-worker/CHANGELOG.md
+++ b/packages/ws-worker/CHANGELOG.md
@@ -1,6 +1,10 @@
 # ws-worker
 
-## 1.14.6
+## 1.15.0
+
+Move expression compilation down into the run thread, isolating its memory from the main worker thread. This should work around a memory leak in long-running workers and improve the robustness of the worker generally.
+
+Note that workflows with large expressions and low run memory limits may be OOM killed as a result of these changes,
 
 ### Patch Changes
 
--- a/packages/ws-worker/package.json
+++ b/packages/ws-worker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openfn/ws-worker",
-  "version": "1.14.6",
+  "version": "1.15.0",
   "description": "A Websocket Worker to connect Lightning to a Runtime Engine",
   "main": "dist/index.js",
   "type": "module",


josephjclark:
I think I'm gonna add a fix for https://github.com/OpenFn/kit/issues/74, which will mitigate the increased risk of crashes

PR #1009, opened 2025-09-25
feat: openfn projects & checkout command

### LINKED ISSUE #1001: sandbox: cli project
Add a new command to the CLI: `openfn project`

- List the currently checked out project
- Say whether there's a diff from the yaml file
- maybe List the workflows by name
- Maybe if --verbose is passed, show the version hash of each workflow
- List all the domains associated with the project
- For each domain, list the project and its sandboxes by name
- List all the project files, by name and domain, that are available
- Maybe show the last sync date

Some of  these bits can be added later - for now we just need the basic project command wired up.

The project logic is closely coupled to the `.openfn` file, which says "this folder is an openfn Workspace". The `.openfn` file won't contain much yet, but will come to include:
- options for the repo, like where workflows should live, or the path to the project files
- the name of the original project in Lightning - the basic name for the project
- the name of the currently checked out project (ie, sandbox)
- Any metadata that a) is needed in the CLI, b) isn't contained in the workflow files themselves, and c) is held in the project files

### `project`

We need to add to the `project` subpackage a new `Workspace` class.

With the constructor `new Workspace('path/to/some/folder')`, a `Workspace` needs to:
* Look for a `.openfn` file in the root of the folder, or else error
* Set paths for where workflows live and where project files live
* Load other options from the `.openfn` file
* For each project file found in the directory substructure, create a new `Project` instance

A `Workspace` will have a bunch of utility functions, to be developed as we go. For example to get the list of known projects. The workspace does't _do_ very much, it's just a central lookup for information

There should be a static `init` function which creates a new `.openfn` file with some defaults

### `cli`

Once the project stuff is working, we can link the CLI to it

The CLI will need a bunch of subcommands: `project init`, `project merge`, `project deploy`, `project pull` etc. Some of these will eb shorthanded (ie, `openfn merge`) and some are already implemented ie, `openfn deploy`)

But the basic command should describe the Workspace, by listing all the known projects and sandboxes.

We can extend the `project` command later so right now, it really just needs to display a minimal set of information. Just enough to validate that the `Workspace` class is working.

The `Workspace` class does all the heavy lifting here.

### What is a project file?

Today, if you run `openfn pull` on a project, it'll download three artefacts: 
* a config.json file, which I think has metadata about the url of the app and stuff
* A project.yaml file, which is an abstract definition of the project and its workflows. Like you get from the workflow source in the app. This is human readable.
* A state.json file, which is all the project  data _plus_ app metadata - basically UUIDs. It's not really meant to be human readble and is strutured a bit differently to the project.yaml. This is returned by the provisioner API.

The project file is _basically_ the state file, but it also includes information about the owning app - like the url of the app that the project came from. A project file can be json or yaml. In CLI deploy v2, there is no config file or project.yaml - just the project/state file

Every time you run `cli pull` in v2, the CLI wll download the state from the provisioner, convert it into a project file, and write it to a folder (probably `/projects`) in the style `name@domain.{yaml|json}`. The name of the project is either the sandbox name (if there's a parent id) or something like `main` if its the primary project.

To know the connected apps for a given repo/folder/workspace, you need to lookup where the project files are stored in `.openfn`, and read the compatible files in that directory.

doc-han:
## Short Description

Introduces new openfn cli commands
1. `openfn projects [project-path]` - lists the openfn projects available in a given path. path is optional.
2. `openfn checkout <project-id>` - checks out a new project as the active one. you can provide either name or id of the project. 

Fixes #1001 maybe

## Implementation Details

## QA Notes

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/packages/cli/src/checkout/handler.ts
+++ b/packages/cli/src/checkout/handler.ts
@@ -3,9 +3,13 @@ import type { Logger } from '../util/logger';
 import getFsProjects from '../projects/get-projects';
 import path from 'path';
 import fs from 'fs';
+import Project, { yamlToJson } from '@openfn/project';
+import { rimraf } from 'rimraf';
 
 const checkoutHandler = async (options: CheckoutOptions, logger: Logger) => {
   const commandPath = path.resolve(process.cwd(), options.projectPath ?? '.');
+  // local name (alias)
+  // remote name (actual project name)
   const idOrName = options.projectId;
   // look for .projects folder
   const projectsDir = path.join(commandPath, '.projects');
@@ -13,13 +17,13 @@ const checkoutHandler = async (options: CheckoutOptions, logger: Logger) => {
     logger.error('.projects folder not found');
     return;
   }
-  const projects = await getFsProjects(projectsDir, logger);
-  if (!projects.length) {
+  const availableProjects = await getFsProjects(projectsDir, logger);
+  if (!availableProjects.length) {
     logger.error('No openfn projects available');
     return;
   }
 
-  const checkoutProject = projects.find(
+  const checkoutProject = availableProjects.find(
     (p) => p.id === idOrName || p.name === idOrName
   );
   if (!checkoutProject) {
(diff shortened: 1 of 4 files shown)


packages/cli/src/projects/handler.ts
```diff
@@ -0,0 +1,28 @@
...
+    logger.error('.projects folder not found');
+    return;
+  }
+  const projects = await getFsProjects(projectsDir, logger);
+  if (!projects.length) {
+    logger.error('No openfn projects available');
+    return;
+  }
+
+  process.stdout.write(
```
josephjclark, line 21: Which `process.stout` and why not our own `logger` class?

PR #1017, opened 2025-09-29
Improved compiler logs

josephjclark:
Explicit log line saying that compilation completed:

image

Error when compilation failed:

image

Sadly the fail error comes a bit out of error because by the time we catch it to report it, a lot of the error processing has happened. There's not a lot I'm really able to do about that, and I'm not willing to spend a lot of time on it this week :(

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/.changeset/late-beers-hope.md
+++ b/.changeset/late-beers-hope.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Improve compiler logging output


josephjclark:
>  "3 character plus space" rule

I have never heard of this rule! This is why the log output looks wierd?

What do you know about it and where can I learn more?

Certainly something is up with how that error has been parsed by the logger there

josephjclark:
The OOM error is just a line break in the message above it. So the source it `R/T`. I'm a bit bamboozled on why that's not working, I'll poke around a bit.

Namespaces are three chars but just by convention IIRC. It looks neater that way

josephjclark:
TODO: log run-start before compilation

PR #1033, opened 2025-10-02
Fix validate tests

josephjclark:
I don't know why this is a problem now, but that validator test uses a `mock` function which loads pnpm modules in to the mock. It's slow and I _think_ it's  the source of our grief.

I think this pattern was copied out from the CLI, but it's not needed here


josephjclark:
Hmm. Failed but for a different reason :thinking:

josephjclark:
Unit tests have passed twice in a row. It'll do. Let's see.

PR #1048, opened 2025-10-06
report capacity

taylordowns2000:
When joining the worker_channel, this makes the worker report their configured run capacity.

It's used by https://github.com/OpenFn/lightning/pull/3725 so that Lightning can check to see how much capacity exists (connected workers * individual worker capacity) when it wants.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/.changeset/sharp-pumas-shop.md
+++ b/.changeset/sharp-pumas-shop.md
@@ -1,5 +1,0 @@
----
-'@openfn/ws-worker': minor
----
-
-New feature: Report on configured capacity when connecting to the worker channel
--- a/packages/ws-worker/CHANGELOG.md
+++ b/packages/ws-worker/CHANGELOG.md
@@ -1,5 +1,11 @@
 # ws-worker
 
+## 1.16.0
+
+### Minor Changes
+
+- 64176c6: Report on configured capacity when connecting to the worker channel
+
 ## 1.15.4
 
 ### Patch Changes
--- a/packages/ws-worker/package.json
+++ b/packages/ws-worker/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@openfn/ws-worker",
-  "version": "1.15.4",
+  "version": "1.16.0",
   "description": "A Websocket Worker to connect Lightning to a Runtime Engine",
   "main": "dist/index.js",
   "type": "module",


packages/ws-worker/src/channels/worker-queue.ts
```diff
@@ -72,7 +74,10 @@ const connectToWorkerQueue = (
       didOpen = true;
       shouldReportConnectionError = true;
 
-      const channel = socket.channel('worker:queue') as Channel;
+      // Build join payload with capacity if provided
+      const joinPayload = capacity !== undefined ? { capacity } : {};
```
josephjclark, line 78: When would a worker NOT have capacity?

Ah, I see this is the result of bad typing in the server options, which I ought to raise an issue for. But capacity is defaulted and not really optional.

For now, I would simply do `const joinPayload = { capacity }`

packages/ws-worker/src/mock/sockets.ts
```diff
@@ -121,10 +121,15 @@ export class MockSocket {
       });
   }
 
-  channel(topic: string) {
+  channel(topic: string, params?: any) {
     if (!this.allChannels[topic]) {
       this.allChannels[topic] = mockChannel();
     }
+    // Store params on the channel for testing purposes
```
josephjclark, line 128: It doesn't look like this is used, so I was going to say drop it

But actually! The tests you've generated jump through hoops to mock a  mock function :thinking: And it's wierd.

I'd much prefer this pattern - the mock channel simply saves its params, and the test can assert the value has been passed through

packages/ws-worker/test/channels/worker-queue.test.ts
```diff
@@ -106,5 +106,61 @@ test('should fail to connect with an invalid auth token', async (t) => {
...
+      };
+
+      return socket;
+    }
+
+    connectToWorkerQueue('www', 'a', 'secret', logger, {
+      capacity: 10,
+      SocketConstructor: createSocket as any,
+    }).on('connect', () => {
+      t.pass('connected with capacity');
```
josephjclark, line 132: This assertion is inappropriate and should be removed

packages/ws-worker/test/channels/worker-queue.test.ts
```diff
@@ -106,5 +106,61 @@ test('should fail to connect with an invalid auth token', async (t) => {
...
+        return channel;
+      };
+
+      return socket;
+    }
+
+    connectToWorkerQueue('www', 'a', 'secret', logger, {
+      SocketConstructor: createSocket as any,
+    }).on('connect', () => {
+      t.pass('connected without capacity');
```
josephjclark, line 159: This assertion also needs removing. It may cause the test to pass wrongly

packages/ws-worker/test/util/cli.test.ts
```diff
@@ -108,3 +108,19 @@ test('cli should handle array options correctly for env variables', (t) => {
 
   t.deepEqual(args.statePropsToRemove, ['prop1', 'prop2', 'prop3']);
 });
+
+test('cli should parse WORKER_CAPACITY from env', (t) => {
```
josephjclark, line 112: I guess I don't have a problem with these two test, but is there any reason they were added? What is being tested, the names of the env vars?

packages/ws-worker/test/channels/worker-queue.test.ts
```diff
@@ -106,5 +106,61 @@ test('should fail to connect with an invalid auth token', async (t) => {
   });
 });
 
+test('should pass capacity in join payload when provided', (t) => {
+  return new Promise((done) => {
+    function createSocket(endpoint: string, _options: any) {
+      const socket = new MockSocket(endpoint, {}, async () => {});
+
+      // Override channel method to capture join params
```
josephjclark, line 114: As mentioned above, I don't think this is a very productive pattern and I'd quite like to simplify

josephjclark (changes requested):
Cool - I've left a few comments, mostly stylistic but those `t.pass()` assertions will make your tests pass even if you comment out the actual feature code.

taylordowns2000:
@josephjclark , simplified significantly and (i think) addressed all your requests

josephjclark (approved):

PR #1051, opened 2025-10-07
Tmp: fooling around with compiler diagnostics

josephjclark:
.


packages/compiler/src/transforms/add-imports.ts
```diff
@@ -107,6 +107,10 @@ export type IdentifierList = Record<string, true>;
 export function findAllDanglingIdentifiers(ast: ASTNode) {
   const result: IdentifierList = {};
   visit(ast, {
+    visitObjectExpression: () => {
```
josephjclark, line 110: Do the same thing for lazy state. Probably a good idea to share the transformer function?

PR #1084, opened 2025-10-21
Fixing pull --beta and refactoring along the way

### LINKED ISSUE #1083: pull --beta doesn't respect openfn.yaml
The beta pull doesn't take advantage of any of the nice new workspace stuff. So when you pull, it won't look in openfn.yaml for config - it'll just use defaults

In practice this means I can't easily pull state in json format

josephjclark:
Had to get an old state.json format for Elias and it turned into a nightmare

Basically `pull` was written before a lot of our nice `Workspace` APIs and stuff, and was rushed out even then

And what's happening is the pull isn't using config properly from the openfn.yaml file

This PR basically plugs in the Workspace class to make life easier.

But I hit loads of little problems on the way - the code here is super messy and we really need to tidy up a bit

Fixes #1083

## Headline changes

* Rename `project.repo` to `project.config`
* Generally renamed the repo/config stuff do "Workspace Config", or  variations of
* Centralised option parsing into a single file, which sets defaults
* Introduced a new project.yaml format  (`{ project, workspace }`)
* Standardised and simplified how we load openfn.yaml

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/project/src/Workspace.ts
+++ b/packages/project/src/Workspace.ts
@@ -26,9 +26,11 @@ export class Workspace {
     let context;
     try {
       const { type, content } = findWorkspaceFile(workspacePath);
+      console.log(content);
       context = loadWorkspaceFile(content, type);
       this.isValid = true;
     } catch (e) {
+      console.log(e);
       // invalid workspace
       return;
     }
--- a/packages/project/src/util/config.ts
+++ b/packages/project/src/util/config.ts
@@ -108,7 +108,7 @@ export const loadWorkspaceFile = (
   let project, workspace;
   let json = contents;
   if (format === 'yaml') {
-    json = yamlToJson(contents);
+    json = yamlToJson(contents) ?? {};
   } else if (typeof contents === 'string') {
     json = JSON.parse(contents);
   }
@@ -146,10 +146,13 @@ export const loadWorkspaceFile = (
 };
 
 export const findWorkspaceFile = (dir: string = '.') => {
+  console.log({ dir });
   let content, type;
   try {
     type = 'yaml';
+    console.log(path.resolve(path.join(dir, 'openfn.yaml')));
     content = readFileSync(path.resolve(path.join(dir, 'openfn.yaml')), 'utf8');
+    console.log({ content });
   } catch (e) {
     // Not found - try and parse as JSON
     try {
@@ -159,7 +162,7 @@ export const findWorkspaceFile = (dir: string = '.') => {
         content = JSON.parse(file);
       }
     } catch (e) {
-      // console.log(e);
+      console.log(e);
       // TODO better error handling
       throw e;
     }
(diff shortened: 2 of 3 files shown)


packages/cli/src/pull/beta.ts
```diff
@@ -37,47 +37,52 @@ export type PullOptionsBeta = Required<
...
 
+  // TODO `path` or `output` ?
+  // I don't think I want to model this as output. deploy is really
+  // designed to run from the working folder
+  // could be projectPath or repoPath too
+  const outputRoot = path.resolve(options.path || '.');
+
+  // TODO is outputRoot the right dir for this?
+  const workspace = new Workspace(outputRoot);
+  const config = workspace.getConfig();
```
josephjclark: Not a fan of the getter - I just want to use workspace.config here. Why make it private?

packages/cli/src/pull/beta.ts
```diff
@@ -37,47 +37,52 @@ export type PullOptionsBeta = Required<
...
-
   const projectFileName = project.getIdentifier();
 
   await fs.mkdir(`${outputRoot}/.projects`, { recursive: true });
   let stateOutputPath = `${outputRoot}/.projects/${projectFileName}`;
 
   const workflowsRoot = path.resolve(
     outputRoot,
-    project.repo?.workflowRoot ?? 'workflows'
+    project.repo.dirs?.workflows ?? 'workflows'
```
josephjclark: Shouldn't need any of this optional chaining bullshit: I want good robust config with everything defaulted

josephjclark:
Oh and `openfn.yaml` needs to look like this:
```
project:
workspace:
```

That gives you metadata about the checked out project, and config for the workspace generally

We should support legacy and new file formats for a while, but always serialize to the new

josephjclark:
`openfn.yaml` now looks like this:
```
project:
  uuid: d3367267-ef25-41cf-91b2-43d18f831f3c
  endpoint: https://app.staging.openfn.org
  env: project
  inserted_at: 2025-10-21T17:10:57Z
  updated_at: 2025-10-21T17:10:57Z
workspace:
  dirs:
    projects: .projects
    workflows: workflows
  formats:
    openfn: yaml
    project: json
    workflow: json
```

Old yaml files, pre 0.6, are still 100% compatible. Very few tests have updated to the new structure.

:man_cook: :kiss:

josephjclark:
@doc-han When you get a chance I'd like you to have a close look over this please: partly for your opinions (this is the right thing to do right??) but also to keep you in-sync with the changes. There's  more work I want to do and you might be "it" for the next ticket

josephjclark:
@doc-han agreed - let's try and fix those as we go. We should be striving to improve the typing and general semantics on every PR from here going forward - I welcome your contributions in cleaning this all up!

Issue #926, opened 2025-04-14
worker: use streaming to stringify and calculate size


josephjclark:
There are a few places in the worker where we stringify message payloads and state objects.

It's hard to be sure, but I have a suspicion that stringifying is what's causing heap trace errors. I'm not really sure what else it might be at this stage (downloading a huge payload into the worker thread through a request should trigger a graceful OOMkill because the stream is parsed asynchronously).

Instead of using JSON.stringfy (especially to estimate the size of a payload), we should try using streaming techniques. These should have a lower working memory overhead.

Now this might not be straightforward. We use `fast-safe-stringify` to handle circular structures and unsafe properties in objects. Does it mean we have to drop that? Can we make that compatible with streaming? Can we do this case-by-case?

Maybe related to #905

josephjclark:
Having spent some time investigating this with Claude - I basically don't think we  can get any better than JSON.stringify for this.

See https://github.com/OpenFn/kit/pull/1155 for notes and ideas, but I'm ready to write this off as "not worth it"

PR #1155, opened 2025-12-04
Engine: try to make calculating payload sizes more efficient

josephjclark:
Addresses #926

The engine will refuse to send an object if it exceeds a certain size.

The only way we can work out the size in prod right now is to stringify it and measure the string.

This is obviously problematic. It could lead us to fully parse a 100mb string. And for log statements, which might be high volume, we're just adding a lot of needless processing. This might even be causing lost runs.

With the help of Claude, this PR adds a solution to stream an object to JSON, measure its size, and exit streaming when the limit is exceeded.

The two benefits of this approach are:
1) We can abort as soon as we exceed the limit. By default right now we'd only parse say 10mb of that 100mb object
2) It's now async and non-blocking, yeilding the CPU. A large string won't block the engine from doing other stuff.

I've written a fancy benchmark script to try and assess what this does to performance. See the comments at the end of this PR for the latest results (tl;dr - speed is the same, which is good, memory usage may be reduced by 10-20% for larger payloads)
## AI Usage

- [x] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/engine-multi/src/util/ensure-payload-size.ts
+++ b/packages/engine-multi/src/util/ensure-payload-size.ts
@@ -66,7 +66,6 @@ export const calculateSizeStream = async (
 };
 
 export default async (payload: any, limit_mb: number = 10) => {
-  return payload;
   const newPayload = { ...payload };
 
   for (const key of KEYS_TO_VERIFY) {
--- a/packages/engine-multi/test/integration.test.ts
+++ b/packages/engine-multi/test/integration.test.ts
@@ -518,7 +518,7 @@ test.serial('redact final state if it exceeds the payload limit', (t) => {
 
     const expression = `
 export default [(state) => {
-  state.data = new Array(1024 * 1024).fill('a')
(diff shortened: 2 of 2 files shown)


josephjclark:
I get very similar diffs if I don't run either calculation and just return. Sort of makes me thinks a <5% diff is just noise

josephjclark:
With a 16mb JSON object the difference gets a little bit more pronounced:
```

Time Performance:
  Min: -6.90%
  Max: -13.43%
  Avg: -6.30%

Memory Usage:
  Min: -8.80%
  Max: -21.60%
  Avg: -10.10%

(Negative values mean traverse is better)
```

For traverse, this is consuming about 315mb of memory (and is quite tight around that). Stringify averages around 250mb but spikes up to 400

My feeling is that this is better, but it's not a paradigm shift in terms of performance. The non-blocking is probably nice, but remember that each run is in its own thread anyway - so it doesn't affect other runs or the main worker if it does block

Issue #790, opened 2024-09-26
Worker: batch log events


josephjclark:
The worker currently sends log lines individually. Because there's a lot of logging, it can result in tens of logs per run per second.

During times of high load, that means a lot of strain coming from log events.

We should batch up subject to an argment driven limit - say every second. The worker batches log events up and only sends once per second. Probably on a throttle rather than debounce.

PR #1161, opened 2025-12-08
New event processor

josephjclark:
New attempt at batching log events for #790

First step is this: refactor the  core event look so that  instead of just listening to events and doing stuff, we form an orderly queue of events.

This PR touches deep internals but shouldn't materially affect anything else.

The new processor isn't tested standalone, but I think there's so much testing in this repo we can get away with it.

This only took an hour to write. Next step from here is to use the queue to intelligently batch events. I'll splice it with the other  branch PR because I want to keep tests.
## AI Usage

- [ ] Strategy / design
- [ ] Other
- [x] I have not used AI

### DIFF
--- a/packages/ws-worker/src/api/process-events.ts
+++ b/packages/ws-worker/src/api/process-events.ts
@@ -60,6 +60,7 @@ export function eventProcessor(
     }
 
     if (name in callbacks) {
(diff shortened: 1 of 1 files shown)


josephjclark:
Superceded by #1162

PR #1162, opened 2025-12-09
Worker: support optional batching of log events

### LINKED ISSUE #790: Worker: batch log events

josephjclark:
This PR builds on top of #1161.

It adds batch processing to the new event processor. It  basically works like this:

- Events can be arbitrarily queued. Just pass an option to the processor
- When an event is queued, we do the following:
  - Set a batch flag to listen for all incoming events
  - Save any batched events to an array
  - Look in the queue to see if there are any other events already queued. If so, add them straight into the batch
  - After 10ms, send the array of batched messages to the listener. A batch listener must be designed to handle an array of events, not just one event
  - If a different event comes through while batching, cancel the timeout, instantly send the batch, and start processing the new event

Fixes #790

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/ws-worker/src/api/process-events.ts
+++ b/packages/ws-worker/src/api/process-events.ts
@@ -48,8 +48,13 @@ const allEngineEvents = [
   WORKFLOW_ERROR,
 ];
 
-// argument says which events can be batched
-// batched events will call the callback with an array
+/**
+ * Queues engine events for sequential processing to Lightning with optional batching.
+ *
+ * Queuing ensures events are sent in order while allowing batching to reduce network calls.
+ * Batching helps with high-volume logs by sending fewer requests with larger payloads,
+ * reducing websocket latency. Events batch by count or time interval, whichever comes first.
+ */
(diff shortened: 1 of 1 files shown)


josephjclark:
TODO need to add more integration level tests around batching. I've made the log handler robust to batch and non batch, which basically means that most of the tests are in non batch mode.

I need to either add a few explicit batch tests, or make batching default in all tests. For the sake of an easier life I think I'll just add a few new tests

josephjclark:
When enabling batch logs by default on the worker, we should make that a breaking change (because not all lightning versions will be compatible)

Alternatively, negotiate support with lightning when joining the run channel, so that we ensure compatiblity

PR #1167, opened 2025-12-12
Add Project aliases with fuzzy matching

josephjclark:
Adds Aliases to Projects.

The `alias` replaces `env` (which was chosen before sandbox environments and is now causing chaos and confusion). `alias` is saved in the Project class, but does not get serialized to a state file. It's just in the file name. This is nice because if a user wants to change alias, all they have to do is rename the file and it's instantly recognised.

You can also use a domain with an alias.

Basically the idea is that anywhere you can reference a project, you can now use its id, UUID or alias.

There are a bunch of other refactors here:
- Removed `getIdentifier` in favour of `get qname`. A qname, or qualified name, is simply a `alias@domain` identifier.
- Because the big options list in `options.ts` has gotten too big and broken as a the CLI has diversified, I've started a new `project/options` file which just holds options used by the project subcommand. This was a little painful but it's definitely a positive refactor, and I'll start slowly replicating that refactor across other commands
- Error handling in fetch has been improved

So now we can do:
- Fetch or pull a project by UUID, and specify an alias
- fetch or pull an existing project by alias
- Checkout by alias
- Merge by alias

I've not done the deploy side of things as there's already a large deploy PR in progress. I'll rebase onto this and fix deploy over there.

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/cli/src/projects/fetch.ts
+++ b/packages/cli/src/projects/fetch.ts
@@ -87,7 +87,7 @@ export const handler = async (options: FetchOptions, logger: Logger) => {
 
   ensureTargetCompatible(options, remoteProject, localTargetProject);
 
-  // TOOD should we use the local target projet for output?
+  // TODO should we use the local target project for output?
 
   // Work out where and how to serialize the project
   const outputRoot = resolvePath(outputPath || workspacePath);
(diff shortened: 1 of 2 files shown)


josephjclark:
I think I'm going to change my mind here and make alias the bit before the `@` in the file name, and otherwise not write it to the project file at all.

The benefit of this is that changing aliases is easier - you just rename the file and it'll be instantly recognised.

otherwise there's confusion between the `alias` and the file name, and it's harder to change

josephjclark:
Ah, also need aliasing on merge

josephjclark:
Almost happy with this but I've got myself a bit confused by something.

```
openfn fetch 21345 --alias main
openfn fetch main --alias main
```

What is the meaning of alias?

If you're asking for a UUID, the alias is the file name to fetch to locally. Right now we try and find a local project file based on the UUID - not the alias. And that means when we write to the local alias file, we haven't properly checked for conflicts.

If the identifier is an alias, then `--alias` doesn't make sense to pass. You've passed two local identifiers.

I suppose you could be saying "fetch the project with local alias A, and save it to local alias B!.

So the identifier is what you use to look up the remote uuid. The `--alias` flag, like the `--output` flag, is what you want to name it locally.

I will need to refactor a bit around this important distinction, and add acouple of considered tests.

PR #1199, opened 2025-12-30
Collections integration test

josephjclark:
Attempt to create an integration test for collections

I've updated the lightning mock to use the mock collections API in the adaptor (needs pnpm link to work)

The integration tests uses a real lightning mock server, and so far as I can tell its hitting the endpoint (kinda wierd to debug). But for some reason it always returns a 404? Or an error deep in adaptor code.

I also notice that the collections adaptor verison isn't listed in the versions list. Neither is common for that matter.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -7,10 +7,6 @@ settings:
 importers:
 
   .:
-    dependencies:
-      '@openfn/language-collections':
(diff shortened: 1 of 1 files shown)


josephjclark:
I think it's just something wrong in the mocked response structure. Can't see what. Time for a break.

josephjclark:
Ah got it - it was me all along. A collection value MUST be a string. I was adding objects using the debug API.

Maybe a good idea for the mock API to throw in these case eh

PR #1206, opened 2026-01-06
Include start  from pulled projects

### LINKED ISSUE #1204: Projects v2: ensure a start node is set
When using new Project fetch to pull a project, the workflow.yaml a) does not include a start step and b) orders steps by id.

The problem with this, of course, is that when you try and run the workflow, it won't start from the trigger node - it'll start from the first step!

Either we need to pin the trigger to be the first step - which is a bit wierd. Or we need to set the `start` option in the workflow.yaml to be the trigger.

This start option needs to get ignored when pushing to the app. That needs a bit of finessing.  I'd prefer that all options sync, or none of them - not a different rule per option.

josephjclark:
* [x] Fix how the CLI currently reads workflow.yaml files to support the `options` key
* [ ] Fix how project currently serialises yaml files (use workflow, options)
* [ ] Write a `start` to all workflows when pulling from state
* [ ] Ensure workflow  start is properly  serialized to v2

When done, this will change the structure of workflow files on disk, and enable `openfn execute` to work on pulled workflows.

Fixes #1204 

## AI Usage

- [x Code generation (copilot but not intellisense)
- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/lexicon/core.d.ts
+++ b/packages/lexicon/core.d.ts
@@ -161,8 +161,6 @@ export type Workflow = {
 
   // holds history information of a workflow
   history?: string[];
-
-  start?: string;
 };
 
 export type StepId = string;
--- a/packages/project/src/Workflow.ts
+++ b/packages/project/src/Workflow.ts
@@ -16,9 +16,9 @@ class Workflow {
   start?: string;
   id: string;
   openfn?: l.WorkflowMeta;
-  options: any; // TODO
+  options: l.WorkflowOptions;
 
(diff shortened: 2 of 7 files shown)


josephjclark:
Ok, so this whole implementation is wrong.

- We need to drop the `{ workflow, options }` structure
- We should default the trigger to the first step
- We should support a first class `start` key on a workflow structure
- The project file and workfow.yaml should both have an options key, which is probably unused

PR #1211, opened 2026-01-08
Release: CLI 1.23.0

josephjclark:
- tweak to log output
- updated start handling
- execute from workflow.yaml direct

### DIFF
--- a/packages/cli/CHANGELOG.md
+++ b/packages/cli/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 ### Minor Changes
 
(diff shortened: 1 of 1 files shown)


packages/cli/CHANGELOG.md
```diff
@@ -1,5 +1,33 @@
...
+
+  ```
+  openfn ./workflows/process-patients/process-patients.yaml
+  ```
+
+  When running through a Workspace, credential maps and collections endpoints are automatically applied for you.
+
+### Patch Changes
+
+- b262d10: Support workflow.jaml/json files without a top workflow key
```
josephjclark: The point of this is: you can run workflows direct now, with the correct start node

PR #1214, opened 2026-01-08
Include start node when pulling workflows, plus options refactor

### LINKED ISSUE #1204: Projects v2: ensure a start node is set

### LINKED ISSUE #1213: Rethink Workflow options
this is a kinda big project, needs to be done with time and patience

Current a workflow.json serializes like this:
```
{ workflow, options }
```
Where 99% of the time options is just the start node

It's kind of an annoying data structure, and annoying to hand write. So workflow.yaml shuld just contain a workflow and nothing else.

The seralized workflow structure, in the v2 project yaml and and the serialized workflow.yaml, should have an options key.

Considerations here:
- For back  compat, we need to support the `{ workflow, options }` format and convert it at ingress time in the CLI
- Is `start` an option, or core to a workflow? When I pull from lightning, the trigger node is always the start, that's not an option really. It's how the workflow is designed. That's different to a CLII start option, where the workflow starts at trigger but I want to skip that and just run from step x. That's probably not even an option we want to serialize?
- So I think `start` is not an option

The other way of looking at this is:
- Yes, drop the `{ workflow, options }` wrapper
- when converting a lightning project, make the trigger the first step, so that it is naturally executed first
- Support an optional `start` key on a workflow to set a custom start node. Not likely to  be useful but cheap to support
- Don't serialize workflow options at all? Just ignore them. Reintroduce them only when needed.

josephjclark:
Third attempt to set a `start` key on workflows when loading from app state.

Fixes #1213 I think
Fixes #1204

This turns out to be quite a big deal because basically the handling of start and options just isn't so good.

So changes in this PR:

- The workflow.json format of `{ workflow, options }` is now considered old hat and legacy
- Now a workflow.yaml or workflow.json just contains the workflow. No option stuff.
- A workflow supports a `start` key as  first class citizen
- When pulling from app state, we set this start key to the trigger (hard coded)
- Project.yaml and workflow.yaml now have a consistent structure (including options, which I think is never used)

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/packages/cli/src/util/load-plan.ts
+++ b/packages/cli/src/util/load-plan.ts
@@ -37,6 +37,12 @@ const loadPlan = async (
     const content = await fs.readFile(path.resolve(options.path), 'utf-8');
     options.baseDir = dirname(options.path);
     workflowObj = yamlToJson(content);
+    const { options: o, ...rest } = workflowObj;
+    // restructure the workflow so that options are not on the workflow object,
+    // but part of hte execution plan options instead
+    if (!workflowObj.workflow && workflowObj.options) {
+      workflowObj = { workflow: rest, options: o };
+    }
   }
   // Run a workflow from a project, with a path and workflow name
   else if (options.path && options.workflow) {
--- a/packages/cli/test/util/load-plan.test.ts
+++ b/packages/cli/test/util/load-plan.test.ts
@@ -444,7 +444,7 @@ test.serial('xplan: append collections', async (t) => {
 });
 
 test.serial(
-  'xplan: load a workflow.yaml (without top workflow key)',
+  'xplan: load a workflow.yaml without top workflow key',
   async (t) => {
     mock({
       'test/wf.yaml': `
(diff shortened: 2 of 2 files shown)


josephjclark:
Wow I  thought this was going to take ages  - but I think this is it!

I want to run some manual QA tomorrow with openfnx:
- Run a `{ workflow, options }` from a workflow.json
- Run a `{ workflow, options }` from a workflow.yaml
- Run a `(workflow)` from a workflow.json
- Run a `(workflow)` from a workflow.yaml
- Respect options.start when it's provided
- Fetch from the app and run the project: it should start from the correct node

I'll also check what integration tests are like for this stuff

PR #1226, opened 2026-01-16
More deploy stuff & cli 1.25.0

josephjclark:
Likely to be a bitty PR with numerous small changes:

Everythign in  #1230

- Allow `--format state` on fetch to download a v1 state file. Not well tested. Useful for debugging only. Should write to .projects if `-o` isn't set, but that's not tested either. Proabably I should undocument this flag

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/.changeset/bumpy-things-report.md
+++ b/.changeset/bumpy-things-report.md
@@ -1,5 +1,0 @@
----
-'@openfn/cli': patch
----
-
-Fix an issue where start is not correctly loaded from workflow.yaml
--- a/.changeset/forty-steaks-study.md
+++ b/.changeset/forty-steaks-study.md
@@ -1,5 +1,0 @@
----
-'@openfn/project': patch
----
-
-Workspace: add getters for project and workflow paths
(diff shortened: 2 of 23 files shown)


josephjclark:
Think I've stumbled into a sizeable problem here. When loading the deploy project from the fs, there's no history. So  you're always try to merge a zero-history workflow into one with a history. Which results in an error, because we can't guarantee no overwrite.

But when deplying, don't we merge the fs into the local target? So why doesn't the local-merged Workflow have the history from the local file? Maybe this s just a tiny bug in workflow merging

Is there also a problem here:
1. User fetches a workflow
2. User matches a change locally. New history.
3. User  fetches from app. App has a new edit. So the local history (which excludes the current change) has an extra item.
4. Now the history locally is kinda wrong right? there's a mismatch between what's on the disk and where we forked from the remote

What we may need to do is: on fetch, if the local version is different to the new version coming down, we generate a new cli version in local history. Part of the problem there is that the version hashes aren't actually compatible.

Hmm. Getting a bit tricky in here...

PR #1238, opened 2026-01-23
Update version hash

### LINKED ISSUE #1078: CLI and App version hashes are not the same
Right now the CLI and app use a different structure to represent a workflow, and as a result they generate a different hash for the same state.

This can be a problem: if the CLI and App both make the same change to a workflow, it'll generate a different hash and so register as a conflict.

This would be resolved by doing a full diff, but it kind of undermines the the versioning strategy.

The resolution to this is to use the same language for keys and structures in both sides. And the only real lingua franca we have is the Provisioner's state.

So this is an issue for kit and lightning to BOTH generate version hashes from the state representation.

This would have to support v1 and v2 etc state versions.

josephjclark:
A rewrite of the version hash algorithm to better align with  Lightning

I am now convinced that lightning and the CLI  MUST generate identical hashes, otherwise we can't really work out what's changed when deploying.

The basic strategy here is, rather than use the Project structure to generate the hash, convert to app state and use that. This means we're much more closer to the lightning structure, and it's a lot easier to compare the algos.

Fixes #1078

## AI Usage

- [ ] Strategy / design
- [ ] Other
- [ ] I have not used AI

### DIFF
--- a/packages/cli/test/projects/fetch.test.ts
+++ b/packages/cli/test/projects/fetch.test.ts
@@ -458,7 +458,7 @@ test.serial(
             lock_version: 1,
           },
           id: 'my-workflow',
-          history: ['cli:02582f3bb088'],
+          history: ['cli:ba19e179317f'],
         },
       ],
     };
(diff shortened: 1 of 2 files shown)


josephjclark:
Ok, that was super hard to debug, but I've finally got CLI and Lightning hashes to look identical.

All tests should be passing on this branch, so I've gotta investigate those fails before I can merge down

PR #1294, opened 2026-03-10
Fast lanes: dedicated worker capacity for named queues

### LINKED ISSUE #1288: Worker: Add `--queues`  option and slot group configuration parsing
### Summary

Add a new `--queues` CLI option to the worker that lets operators divide
worker capacity into slot groups, each with its own queue preference chain.
This issue covers configuration parsing and validation only — the runtime
behavior (per-group workloops) is a separate issue.

### Background

Currently the worker has a single `--capacity` flag (default 5) that controls
the total number of concurrent runs. All runs are claimed from a single global
queue with no differentiation.

The fast lanes feature requires workers to dedicate specific slots to specific
queues. For example: 1 slot pinned to sync webhook work, 4 slots for general
work.

### Configuration syntax

```
--queues "<preference>:<count> <preference>:<count> ..."
```

| Delimiter | Meaning |
|-----------|---------|
| `,` | Queue preference separator (left = highest preference) |
| `*` | Wildcard: accept runs from any queue |
| `:N` | Slot count for this preference group |
| ` ` (space) | Slot group separator |

**Examples:**
```bash
# 1 slot pinned to fast_lane, 4 slots preferring manual then anything
--queues "fast_lane:1 manual,*:4"

# 5 generic slots (pure FIFO across all queues)
--queues "*:5"

# 2 fast lane (strict), 3 with manual preference
--queues "fast_lane:2 manual,*:3"
```

**Environment variable:** `WORKER_QUEUES`

### Slot group data structure

Each parsed group should produce something like:
```typescript
interface SlotGroup {
  queues: string[];     // e.g., ["fast_lane"] or ["manual", "*"]
  maxSlots: number;     // e.g., 1
}
```

### Backward compatibility

- `--queues` and `--capacity` are **mutually exclusive** — specifying both is
  an error
- If neither is provided, default to `--capacity 5` behavior
- When `--queues` is not provided, derive from `--capacity`:
  `manual,*:<capacity>` (e.g., capacity 5 → `"manual,*:5"`)
- This preserves current behavior where manual/immediate runs get preference

### Validation rules

- Each slot group must have count >= 1
- Queue names must be non-empty strings (alphanumeric + underscore)
- `*` can only appear as the last element in a preference chain
- Total slot count across all groups becomes the effective capacity
- `--queues` and `--capacity` are mutually exclusive

### Protocol type update

Update `ClaimPayload` in the lexicon types to include the optional `queues`
field:
```typescript
interface ClaimPayload {
  demand?: number;
  worker_name?: string;
  queues?: string[];  // NEW: queue preference chain
}
```

### Testing

- [ ] Parse `"fast_lane:1 manual,*:4"` → two slot groups with correct queues and counts
- [ ] Parse `"*:5"` → one slot group with `["*"]` and 5 slots
- [ ] Parse `"fast_lane,manual,*:1 *:4"` → correct multi-preference chain
- [ ] Reject `"*,fast_lane:1"` (wildcard not at end)
- [ ] Reject `":0"` or `":abc"` (invalid counts)
- [ ] Reject conflicting `--queues` and `--capacity` flags
- [ ] Default derivation: capacity 5 with no `--queues` → `[{ queues: ["manual", "*"], maxSlots: 5 }]`

### LINKED ISSUE #1289: Implement per-group workloops and queue-aware claiming
### Summary

Replace the worker's single workloop with independent per-group workloops.
Each slot group gets its own workloop that claims with its queue preferences,
tracks its own capacity, and backs off independently.

### Background

Currently the worker has one workloop that calls `claim()` with
`{ demand: 1, worker_name }`. After this change, each slot group runs its own
workloop that sends `{ demand: 1, worker_name, queues: [...] }`.

A worker configured with `--queues "fast_lane:1 manual,*:4"` will have:
- One workloop claiming `["fast_lane"]` with 1 slot
- One workloop claiming `["manual", "*"]` with 4 slots

### Per-group capacity tracking

Each slot group independently checks its own capacity:
```
if (group.activeRuns + group.pendingClaims >= group.maxSlots) → stop workloop
```

The global `app.workflows` map continues to track all active runs. The
per-group tracking determines which group's workloop to resume when a run
completes.

When a run is claimed by a slot group, it is associated with that group for
its lifetime. On completion, the group's `activeRuns` decrements and its
workloop resumes if it was at capacity.

### Claim protocol

Each workloop sends its slot group's queue preference in the claim message:
```typescript
channel.push("claim", {
  demand: 1,
  worker_name: "pod-1",
  queues: ["fast_lane"]           // pinned slot
})

channel.push("claim", {
  demand: 1,
  worker_name: "pod-1",
  queues: ["manual", "*"]         // preference slot
})
```

Lightning interprets this as:
- No `*` → filter mode: only return runs from the named queues
- With `*` → preference mode: all runs, ordered by preference position

If Lightning hasn't been updated yet, it will ignore the `queues` field and
return runs as before (Phoenix channels ignore unknown payload keys).

### work-available handling

When Lightning pushes `work-available`, attempt to claim for **each slot
group that has free capacity**. This means a single `work-available` event
may trigger multiple claim requests (one per group with free slots).

This is acceptable because:
- Most deployments will have 2-3 slot groups at most
- Claim requests are lightweight channel pushes
- Lightning handles concurrent claims via `FOR UPDATE SKIP LOCKED`

### Join payload

Update the channel join to include queue configuration:
```typescript
channel.join("worker:queue", {
  capacity: 5,
  queues: { "fast_lane": 1, "manual,*": 4 }  // informational
})
```

The `queues` map is informational — Lightning doesn't use it for routing in
this release. It enables future presence-based queue tracking. `capacity`
remains the total for backward compatibility.

### Backoff considerations

Fast-lane slots with strict pinning (`["fast_lane"]`, no wildcard) will
frequently get empty responses when there's no sync work. The existing backoff
(1s min → 10s max) handles this fine because:
- `work-available` pushes bypass backoff for immediate claiming
- A 1-10s backoff on an idle fast-lane slot has negligible cost

### Things to watch out for

- The refactor of server.ts / workloop management is the most complex part —
  the single workloop assumption is likely baked into several places
- Make sure `resumeWorkloop` only resumes the correct group's workloop, not all
- The run-to-group association needs to survive the full run lifecycle
- Test the edge case where a fast_lane group is idle (getting empty claims)
  while the default group is fully busy — they should not interfere with each
  other

### Testing

- [ ] Each slot group runs its own independent workloop
- [ ] A slot group at capacity stops its workloop
- [ ] Run completion resumes the correct group's workloop (not all groups)
- [ ] Claim messages include the correct `queues` array per slot group
- [ ] `work-available` triggers claims on all groups with free slots
- [ ] Fast_lane group backs off independently when getting empty responses
- [ ] Default group claims normally while fast_lane is idle
- [ ] Join payload includes `queues` map
- [ ] Backward compat: single `manual,*:5` group behaves like old capacity-5

stuartc:
## Short Description

Adds the ability to subscribe worker capacity to named queues via `--workloops`, guaranteeing dedicated capacity for latency-sensitive workloads (e.g., sync webhooks). This PR combines the two Kit-side fast lanes issues.

```
  Main Process (ws-worker)
 ├── Workloop 1 (manual>*:2)  ─┐
 ├── Workloop 2 (fast_lane:1) ─┼── all run in the main process as async loops
 ├── Workloop 3 (*:5)         ─┘
 │
 └── Engine (single instance, shared by all lanes)
     └── Child Process Pool (capacity = sum of all lanes/slots)
         ├── Child 1 (forked) → Worker Thread (per task)
         ├── Child 2 (forked) → Worker Thread (per task)
         ├── ...on demand, reused after each task
         └── Child N
```

Fixes #1288, fixes #1289

## Feature Design

At present, the worker has a single workloop (or claim loop), which repeatedly calls lightning on a backoff and asks for more work. In this architecture, Lightning is responsible for prioritising which work gets scheduled next.

The problem with this architecture is that Lightning cannot reserve capacity: it'll just hand off the most important work _right now_ to be done. But if something super high priority comes in, like a synchronous workflow request, and if all the workers happen to be busy, the work has to wait.

To resolve this, Lightning has three work queues: a fast lane for sync requests, a manual queue for user triggered runs, and a default queue for everything else.

Workers mirror this architecture by having several independent workloops. Each workloop makes claim requests to lightning and specifies a _prioritised_ list of channels to pull from. Each workloop has  its own independent backoff and  capacity.

Now a worker can pull work from the manual and default queues in the app, but reserve capacity for fast lane work, ensuring that those priority workflows get pulled ASAP.

Note that the prioritised queues allow for work stealing. The `*` queue will pull anything - so if there's nothing in the manual runs queue, work will be pulled from the default queue. It'll even pull from the fast lane if work is queued up there.

The precise configuration of queues and workloops is purely config driven.

Benefits of this approach:
- Workers can reserve capacity for priority work - like synchronous requests
- Workers can still request from queues in a prioritised way, taking from the manual runs queue before the scheduled work queue.

## Implementation Details

This PR is composed of two feature branches:

### #1291 — `--workloops` CLI option and slot group configuration parsing (kit#1288)

- Adds `--workloops` CLI option (and `WORKER_WORKLOOPS` env var), mutually exclusive with `--capacity`
- Parses slot group definitions like `--workloops "fast_lane:1 manual>*:4"` into typed `WorkloopConfig[]` objects
- Validates queue names, slot counts, wildcard placement, and duplicate queue assignments
- Adds `queues` field to `ClaimPayload` in `@openfn/lexicon`

### #1293 — Per-group workloops and queue-aware claiming (kit#1289)

- Each slot group gets its own independent workloop, replacing the single shared loop
- Each group tracks its own active runs and capacity independently
- Claims are scoped to the group's queue list
- The Lightning join payload now includes a `queues` map so the server knows the slot distribution
- Default behavior (no `--workloops`) is preserved with a single `manual>*` group

### Additional fixes included

- **Fix CLI args via pnpm**: pnpm v7+ passes `--` through to `process.argv`, causing yargs to ignore all flags. Stripped the leading `--` before parsing.
- **Fix `Promise.any` type error**: Added `ES2021` to ws-worker's tsconfig `lib` for `Promise.any` support.

## QA Notes

- Test with `--workloops "fast_lane:2 *:3"` — should show two separate workloops in logs
- Test without `--workloops` — should behave identically to before (single `manual>*` group)
- Verify `--capacity` and `--workloops` cannot be used together (should throw)
- Lightning-side queue support (issues #4500, #4501) is required for end-to-end queue routing; without it, Lightning will ignore the queue preferences and return runs from any queue

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/per-group-workloops.md
+++ b/.changeset/per-group-workloops.md
@@ -2,4 +2,8 @@
 '@openfn/ws-worker': minor
 ---
 
-Implement per-group workloops and queue-aware claiming. Each slot group now gets its own independent workloop, tracks its own active runs and capacity, and sends queue-scoped claims to Lightning. The join payload now includes a queues map so Lightning knows the slot distribution. Default behavior (no `--workloops`) is preserved with a single `manual>*` group.
+Fastlane support: multiple concurrent workloops, each with its own isolated capacity and backoff.
+
+Claims also include a `queues` key, which specifies a prioritised list of Lightning work queues to claim from.
+
+Configure workloops with the `--workloops` CLI option or env. By default the worker users `manual>*5`, which provides parity behaviour to prior production.
--- a/.changeset/queues-cli-option.md
+++ b/.changeset/queues-cli-option.md
@@ -1,6 +1,0 @@
----
-'@openfn/ws-worker': minor
-'@openfn/lexicon': patch
----
-
-Add `--workloops` CLI option for slot group configuration, enabling dedicated worker capacity for named queues (e.g., `--workloops "fast_lane:1 manual>*:4"`). Mutually exclusive with `--capacity`. Also adds `queues` field to `ClaimPayload` type.


stuartc:
> So in this diagram
> 
> ```
>   Main Process (ws-worker)
>  ├── Workloop 1 (manual,*:2)  ─┐
>  ├── Workloop 2 (fast_lane:1) ─┼── all run in the main process as async loops
>  ├── Workloop 3 (*:5)         ─┘
>  │
>  └── Engine (single instance, shared by all lanes)
>      └── Child Process Pool (capacity = sum of all lanes/slots)
>          ├── Child 1 (forked) → Worker Thread (per task)
>          ├── Child 2 (forked) → Worker Thread (per task)
>          ├── ...on demand, reused after each task
>          └── Child N
> ```
> 
> the effective capacity is 8? The worker's memory must be greater than `8 * runMemory`
> 
> I think at the moment we run with a capacity 2, just to ward against lost runs (less of a problem now than when we configured this). This means we run well under capacity, but at least its safe.

Yeah thats correct, the diagram was more to illustrate the Workloop to Engine connection.

> It's unrelated to this story but I'd really like to implement dynamic memory. Tell the worker its maximum memory (ie 2gb) and it'll track how much it has reserved, and maybe include a memory limit in claims ("hi lightning, can I have 1 workflow with no more than 256mb limit? Thanks bro"). Of course that story is a lot more complex in this lanes architecture because presumably you need to reserve memory per lane. Like don't block the fast lane by over-claiming the default lane. I don't know how we'd manage that.

I could argue we are closer to being able to do this, at least from a query perspective. Sure the config format is going to be finicky. I drew "inspiration" from how fstab mounts and docker does contain network and volume mounts (when using nfs) for the config. Which should immediately make any shudder, but yeah it's a heck of a lock better than jamming 20-40 chars of json inside a shell argument or env.

Anyway, if the workloop is able to (atomically) make enquiries about how much room we have we now have the means to tell Lightning something about what we want. So closer...?

> Was there a discussion around creating multiple lanes in a worker, vs creating a worker with a single dedicated lane ? Like some config that says "give me 1 worker capacity 5 for fast lane, and 2 workers capacity 2 for everything else"

Yes and no, the requirement was specifically for limited resource on-prem deployments, after checking how two of our deployments have been done. There was a massive trade off at least in one of them, that in order to have a dedicated worker, they would effectively have 1 worker for manual and normal runs, and 1 worker for syncronous jobs. Cutting the max capacity 85% of the in half. I landed on having the worker being able to have multiple lanes.

PR #1297, opened 2026-03-12
Fastlanes simpler workloops

josephjclark:
Simplifying the `workloops` implementation with the help of claude code.

The base branch is  a bit too complex - too many moving parts - which makes it harder to think about.

Changes made:

* Flattens the Workloop, WorkloopConfig and WorkloopHandle down into a single Workloop class. This manages its own state and removes the need for complex config and lookups. Way nicer!
* Some test changes (the workloop tests really need to be serial, I think they're only passing through luck)
* Tweaked CLI docs
* Removed app.openClaims, which is now just spreading confusion yet is quite redundant

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/packages/ws-worker/src/api/destroy.ts
+++ b/packages/ws-worker/src/api/destroy.ts
@@ -1,4 +1,4 @@
-import { ServerApp, pendingClaims } from '../server';
+import { ServerApp } from '../server';
 import { INTERNAL_CLAIM_COMPLETE, INTERNAL_RUN_COMPLETE } from '../events';
 
 import type { Logger } from '@openfn/logger';


packages/ws-worker/src/api/claim.ts
```diff
@@ -52,12 +52,12 @@ export const resetClaimIdGen = () => {
 
 const claim = (
   app: ServerApp,
-  logger: Logger = mockLogger,
   workloop: Workloop,
-  options?: ClaimOptions
+  logger: Logger = mockLogger,
```
josephjclark: I've moved this signature around so that the required stuff comes first. I know it's just an internal API but it was bothering me a bit 😬

josephjclark:
that last commit has made integration tests go beserk. Bit concerned about this

PR #1352, opened 2026-04-02
feat: new deploy diff

### LINKED ISSUE #1182: new deploy diff
Thinking about what the new deploy diff might look like

In v1, the diff just ran against the json representation of the state. Trivial things like arrays out of order would trigger a false diff.

I want the v2 diff to be a bit higher level, and data structure agnostic

- First we list all workflows that changed
- Then we list steps that have been added, removed or renamed (just name them, don't show an actual diff)
- Then for any step where properties have changed, render as key-value pairs what the differences are
- Maybe treat expression as binary for now. Or just show like a partial diff on the expression string, up to n characters

It might also be useful to get a machine-readable version of the diff. This is probably a JSON object with paths to diffs. Maybe at each path you have keys like `{ $new: true, $deleted: true, $diff }`, where diff is the new value (for expression strings it's probably the whole string?)

doc-han:
## Short Description

Add a rich diff and JSON diff readout to `openfn project deploy`

Fixes #1182 

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Issue: changing the id of a step results in this error:
```
[CLI] ✘ TypeError: Cannot read properties of undefined (reading 'name')
    at file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:279:57
    at Array.map (<anonymous>)
    at generateHash (file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:276:46)
    at Workflow.getVersionHash (file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:462:12)
    at findLocallyChangedWorkflows (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2277:35)
    at syncProjects (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2412:41)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async handler (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2507:24)
    at async parse (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:4126:12)
```

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
   }
 }
 
-export const reportDiff = (
+const printStepChanges = (
```
josephjclark: Can we move this out of the main deploy.ts and into its own file?

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
...
+      const localLines = localExpr.split('\n').length;
+      const remoteLines = remoteExpr.split('\n').length;
+      const added = Math.max(0, localLines - remoteLines);
+      const removed = Math.max(0, remoteLines - localLines);
+      const summary =
+        added === 0 && removed === 0 ? '<changed>' : `<+${added}/-${removed}>`;
+      lines.push(`      - body: ${summary}`);
+    }
+
+    if (lines.length) {
```
josephjclark: Just thinking aloud

Instead of one function which generates ANd prints the diff at the same time, what if we have a generateDiff and a printDiff?  Where the generator creates a data structure that makes printing super easy, but is also unit testable.

josephjclark: Actually, just poking around, I've reminded myself that the `project` package has its own `project-diff` function, which basically does exactly that: returns a simple data structure to represent the diff (I think the the CLI then returns it)

Maybe the best approach to this then is to add a `workflow-diff` file (or maybe just a `diff` file with two functions) which can handle all the logic of generating a diff. And does complicated stuff like tracking id changes (renames, basically)

josephjclark:
Issue: removing the `configuration` key doesn't seem to generate a diff (in the rich or the json one). But it should result in a change in the state.json right? I wonder if that's a wider bug

josephjclark:
Issue: changes to edges don't seem to manifest in the rich diff  (I do see them in JSON)

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
...
+      logger.always(c.yellow(`    ${step.id || step.name}:`));
+      for (const line of lines) {
+        logger.always(c.yellow(line));
+      }
+    }
+  }
+
+  for (const step of remoteSteps) {
+    if (!localById[step.id]) {
+      logger.always(c.yellow(`    ${step.id || step.name}: removed`));
```
josephjclark: I think a remove step should come out in red. I also think an added step should come out in green

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
...
+
+    const localExpr = step.expression ?? step.body ?? '';
+    const remoteExpr = remote.expression ?? remote.body ?? '';
+    if (localExpr !== remoteExpr) {
+      const localLines = localExpr.split('\n').length;
+      const remoteLines = remoteExpr.split('\n').length;
+      const added = Math.max(0, localLines - remoteLines);
+      const removed = Math.max(0, remoteLines - localLines);
+      const summary =
+        added === 0 && removed === 0 ? '<changed>' : `<+${added}/-${removed}>`;
```
josephjclark: I'd like to colour that summary - new stuff in green and removed stuff in red.

But actually, aren't added and removed mutually exclusive? In which case you just want to say like `+1 lines` or `-1 lines`

packages/cli/src/projects/deploy.ts
```diff
@@ -335,7 +342,21 @@ export async function handler(options: DeployOptions, logger: Logger) {
   logger.debug(JSON.stringify(state, null, 2));
   logger.debug();
 
-  // TODO: I want to report diff HERE, after the merged state and stuff has been built
+  if (remoteProject) {
+    if (options.jsonDiff) {
```
josephjclark, line 346: Before we print the diff, can we add a line saying something like:

"This will make the following changes to the remote project"

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
...
+    const lines: string[] = [];
+
+    if (step.name !== remote.name) {
+      lines.push(`      - name: "${remote.name}" -> "${step.name}"`);
+    }
+
+    const localAdaptor = step.adaptor ?? step.adaptors?.[0];
+    const remoteAdaptor = remote.adaptor ?? remote.adaptors?.[0];
+    if (localAdaptor !== remoteAdaptor) {
+      lines.push(`      - adaptor: ${remoteAdaptor} -> ${localAdaptor}`);
```
josephjclark: This stuff is nice - but I'd like a better way to control which keys generate this from/to diff

packages/cli/src/projects/deploy.ts
```diff
@@ -404,7 +425,65 @@ export async function handler(options: DeployOptions, logger: Logger) {
...
+      continue;
+    }
+
+    const lines: string[] = [];
+
+    if (step.name !== remote.name) {
+      lines.push(`      - name: "${remote.name}" -> "${step.name}"`);
+    }
+
+    const localAdaptor = step.adaptor ?? step.adaptors?.[0];
```
josephjclark: Doesn't the serialized format ensure a consistent typing on `step.adaptor`?

josephjclark (changes requested):
Looks really cool!

I've tested around a bit and found some issues. But before we go too much further I'm not sure the architecture is right, and I'd like to dig into that a bit with you later

doc-han:
Suggestion: When a user hits divergence because the project on remote changed but the locally pulled project has zero changes since last pulled. `openfn project pull` should automatically do a `--force` pull. The user shouldn't have to explicitly pass `--force`
cc: @josephjclark

doc-han:
> Issue: changing the id of a step results in this error:
> 
> ```
> [CLI] ✘ TypeError: Cannot read properties of undefined (reading 'name')
>     at file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:279:57
>     at Array.map (<anonymous>)
>     at generateHash (file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:276:46)
>     at Workflow.getVersionHash (file:///home/joe/repo/openfn/kit/packages/project/dist/index.js:462:12)
>     at findLocallyChangedWorkflows (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2277:35)
>     at syncProjects (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2412:41)
>     at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
>     at async handler (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:2507:24)
>     at async parse (file:///home/joe/repo/openfn/kit/packages/cli/dist/process/runner.js:4126:12)
> ```

This has been resolved. It happens when you change the id of a job but then didn't do same for all edges referencing it. The error has been updated to print out a bit more meaningful message.

doc-han:
I've done several manual testing on this and it works fine. 
This morning I added a change which shows what changes for an edge.

josephjclark:
This is looking absolutely lovely

image

Will be testing closely and might jump in a with a fix. Aiming to get this released today.

josephjclark:
Struggling with the review because I'm hitting type issues. I've raised out https://github.com/OpenFn/kit/issues/1381

I think I'll continue  to test and validate this and then still aim to get this PR released today, even if it has type issues

packages/project/src/util/workflow-diff.ts
```diff
@@ -0,0 +1,151 @@
...
+  { key: 'adaptor' },
+];
+
+export const generateStepDiff = (
+  localWf: Workflow | undefined,
+  remoteWf: Workflow | undefined
+): StepChange[] => {
+  if (!localWf || !remoteWf) return [];
+
+  const localSteps = localWf.steps as any[];
```
josephjclark, line 39: Struggling here with:
a) the typing as `any` - this tells me that types are broken and I need to go off and fix that
b) we're not using provisioner state to do our diffing here- that was a core tenant of the issue

Now I'm going to relax on the provisioner state thing. The problem there is that in provisioner state we call it a `body`. but in your workflow.yam it's called `expression`. We have to relate the diff back in local terms.

It just opens up a gap between what we report on for the diff, and what diff factually materialises when we post to the server. Makes me nervous.

The JSON diff resolves this by doing a pure diff on the two JSON structures... but it'll report the diff with keys that may not be familiar. And of course the JSON diff might not necessarily be the same as the database diff (like you can add a key to the JSON that the provisioner will ignore, so it's not _really_ a diff)

packages/project/src/util/workflow-diff.ts
```diff
@@ -0,0 +1,151 @@
...
+  name: string;
+  type: StepChangeType;
+  changes?: {
+    name?: { from: string; to: string };
+    adaptor?: { from: string; to: string };
+    body?: string;
+  };
+};
+
+const TRACKED_FIELDS: Array<{ key: 'name' | 'adaptor' }> = [
```
josephjclark: This makes me a little nervous. Why not just diff on all keys? If we add a new key tomorrow then we'll have to patch the CLI to report a diff on it.

I get that we have special handling for some keys, like `expression`, but I sort of thing we should have a fallback of just reporting anything.

josephjclark:
Ok here's what I've done:
- I've added support for diffing configuration (not that this seems terribly useful)
- I've made the diff reporter in the CLI a bit more generalised
- I've generated a bunch of unit tests against `diff-workflow` with Claude
- I've raised an issue to reorganise typings properly

I think we're good to launch

josephjclark (approved):

PR #1362, opened 2026-04-09
fix(project): include trigger reply fields in version hash

officialasishkumar:
## Description

This keeps the project-side workflow version hash aligned with Lightning by including `webhook_reply` and `cron_cursor_job_id` in the trigger field list used by `generateHash()`.

Without this, the CLI can compute a different hash from Lightning for the same workflow after those trigger fields change.

## Changes

- include `webhook_reply` and `cron_cursor_job_id` in `packages/project/src/util/version.ts`
- add regression tests covering both trigger fields in `packages/project/test/util/version-workflow.test.ts`
- add a changeset so the fix is released through the monorepo flow

Companion to OpenFn/lightning#4599.

## AI Usage

Used only for research.

### DIFF
--- a/.changeset/quiet-seals-battle.md
+++ b/.changeset/quiet-seals-battle.md
@@ -1,0 +1,6 @@
+---
+'@openfn/project': patch
+'@openfn/cli': patch
+---
+
+Include `webhook_reply` and `cron_cursor_job_id` in workflow version hashes.
--- a/packages/project/test/util/version-workflow.test.ts
+++ b/packages/project/test/util/version-workflow.test.ts
@@ -384,6 +384,40 @@ test('hash a trigger', (t) => {
   t.not(generateHash(webhook), generateHash(cronExpression));
 });
 
+test('hash changes when webhook_reply changes', (t) => {
+  const wf1 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=webhook,webhook_reply=x)-x
+    `
+  );
+  const wf2 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=webhook,webhook_reply=y)-x
+    `
+  );
+
+  t.not(generateHash(wf1), generateHash(wf2));
+});
+
+test('hash changes when cron_cursor_job_id changes', (t) => {
+  const wf1 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=cron,cron_expression="1",cron_cursor_job_id=x)-x
+    `
+  );
+  const wf2 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=cron,cron_expression="1",cron_cursor_job_id=y)-x
+    `
+  );
+
+  t.not(generateHash(wf1), generateHash(wf2));
+});
+
 test('hash changes across an edge', (t) => {
   const basicEdge = generateWorkflow(
     `
--- a/packages/project/src/util/version.ts
+++ b/packages/project/src/util/version.ts
@@ -54,7 +54,13 @@ export const generateHash = (
     'body',
   ].sort();
 
-  const triggerKeys = ['type', 'cron_expression', 'enabled'].sort();
+  const triggerKeys = [
+    'type',
+    'cron_expression',
+    'enabled',
+    'webhook_reply',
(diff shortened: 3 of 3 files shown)


.changeset/quiet-seals-battle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/project': patch
+'@openfn/cli': patch
+---
+
+Include `webhook_reply` and `cron_cursor_job_id` in workflow version hashes so CLI deploy stays aligned with Lightning.
```
josephjclark: You need to remove the "stay aligned with Lightnig bit"  - this PR is just adding support for extra keys in the version hash, that's it.

packages/project/test/util/version-workflow.test.ts
```diff
@@ -384,6 +384,51 @@ test('hash a trigger', (t) => {
...
+    `
+  );
+  const wf2 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=webhook)-x(expression=x)
+    `
+  );
+
+  (wf1.steps[0] as any).openfn.webhook_reply = 'before_start';
```
josephjclark: Instead of this awkward assignment, you can do this in the workflow generator - that's exactly what's it for!
```
 generateWorkflow(
    ` @id wf1
    t(type=webhook,webhook_reply=before_start)-x
    `
```
Please also make sure you're generating the absolute minimal keys on the workflow so that it doesn't distract the test code. We shouldn't need an expression on `x` for example

packages/project/test/util/version-workflow.test.ts
```diff
@@ -384,6 +384,51 @@ test('hash a trigger', (t) => {
...
+
+  t.not(generateHash(wf1), generateHash(wf2));
+});
+
+test('hash changes when cron_cursor_job_id changes', (t) => {
+  const wf1 = generateWorkflow(
+    `@name wf-1
+    @id workflow-id
+    t(type=cron,cron_expression="1")-a
+    a-b
```
josephjclark: You shouldn't need the `a-b` nodes here at all right? Better to clean that up and keep the tests simple

packages/project/test/util/version-workflow.test.ts
```diff
@@ -384,6 +384,51 @@ test('hash a trigger', (t) => {
...
+    `
+  );
+
+  const [trigger1, jobA1] = wf1.steps as any[];
+  const [trigger2, , jobB2] = wf2.steps as any[];
+
+  trigger1.openfn.cron_cursor_job_id = jobA1.openfn.uuid;
```
josephjclark: Again you can do this straight in workflow generation. It doesn't need to be logically consistent either- you can just set the  value to `x` or `y`, rather than digging out the actual uuids like this

josephjclark (changes requested):
Thank you @officialasishkumar! This is cool. Just a couple of tidy ups before I can merge it

josephjclark:
Oh and @officialasishkumar we like contributors to disclose AI usage in some way so that everyone knows where we stand. We have a checkbox for this in the PR template. Can you update the PR description to at least say "I have used AI to generate this PR"?

I've updated the PR description to remove the weird  validation stuff. It's enough in this case that tests are passing in CI

officialasishkumar:
Hi @josephjclark , yes, i have used AI in research and understanding the codebase

.changeset/quiet-seals-battle.md
officialasishkumar: Removed that wording in f021d985.

packages/project/test/util/version-workflow.test.ts
officialasishkumar: Moved this into the workflow generator and trimmed the test setup in f021d985.

packages/project/test/util/version-workflow.test.ts
officialasishkumar: Removed the extra nodes and kept the test minimal in f021d985.

packages/project/test/util/version-workflow.test.ts
officialasishkumar: Changed this to literal `x` and `y` values in the generated workflow in f021d985.

josephjclark (approved):
Thanks @officialasishkumar 🙏

PR #1366, opened 2026-04-13
Worker: memory management improvements

### LINKED ISSUE #897: Worker: Runs get lost if the dataclip is >150mb
On Staging, with a 200mb dataclip, I'm seeing this pattern of behaviour

- The step complete with pretty high memory usage: `[R/T] ❯ Final memory usage: [step 515mb] [system 603mb]`
- The step complete message is sent to lightning
- The run complete message seems to fail to send
- Maybe 1 second later, the server seems to restart. No exception or message in the log.

Since run:complete fails to send, the run will be Lost

Here's an [example on staging](https://app.staging.openfn.org/projects/b387f5a4-a3bb-4bb8-8093-f99436d09d92/history?filters[workorder_id]=f0fc406c-f2da-412d-8817-8388d70c6402)

Here's logs on [GCP](https://console.cloud.google.com/logs/query;cursorTimestamp=2025-03-06T14:15:26.029886548Z;endTime=2025-03-06T14:16:24.861Z;pinnedLogId=2025-03-06T14:15:26.012605292Z%2Fh3rvytmd43g40k8s;query=labels.%22k8s-pod%2Fapp%22%3D%22staging-worker%22%0Atimestamp%3D%222025-03-06T14:15:26.012605292Z%22%0AinsertId%3D%22h3rvytmd43g40k8s%22;startTime=2025-03-06T13:16:24.861Z?project=platform-test-267207)

josephjclark:
Responding to a spate of OOMkills and lost runs: some experiments with memory management

Fixes #897

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have not used AI

### DIFF
--- a/.changeset/humble-ravens-send.md
+++ b/.changeset/humble-ravens-send.md
@@ -1,0 +1,5 @@
+---
+'@openfn/logger': patch
+---
+
+Fix an issue where logger.proxy can throw for nil values
--- a/.changeset/odd-sloths-make.md
+++ b/.changeset/odd-sloths-make.md
@@ -1,0 +1,6 @@
+---
+'@openfn/engine-multi': patch
+'@openfn/ws-worker': patch
+---
+
+Use streaming algorithm to calculate payload size
--- a/packages/engine-multi/src/engine.ts
+++ b/packages/engine-multi/src/engine.ts
@@ -40,7 +40,6 @@ const createWorkflowEvents = (
   workflowId: UUID
 ) => {
   // proxy all events to the main emitter
-  // uh actually there may be no point in this
   function proxy(event: string) {
     context.on(event, (evt) => {
       // ensure the run id is on the event
--- a/packages/engine-multi/src/test/worker-functions.ts
+++ b/packages/engine-multi/src/test/worker-functions.ts
@@ -29,13 +29,13 @@ const tasks = {
   // Most tests should use the mock-worker instead
   run: async (plan: ExecutionPlan, _input: any, _adaptorPaths: any) => {
     const workflowId = plan.id;
-    publish('worker:workflow-start', {
+    await publish('worker:workflow-start', {
       workflowId,
     });
     try {
       const [job] = plan.workflow.steps as Job[];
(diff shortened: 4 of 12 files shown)


josephjclark:
I don't understand why this is failing in tests :thinking: 

Running around in circles. Even in code where the json streaming function isn't called it seems to fail.

I've also bumped into a possible issue with circular structures. It makes me nervous - shouldn't the JSON.stringify function throw if it hits a circular structure? That's something that could be hurting us right now in production

EDIT: Well we do seem to handle circular structures - the runtime must be dealing with that before  events are emitted
```
{
  "x": {
    "x": "[Circular]"
  },
  "data": {}
}
```

josephjclark:
managed to fix one of the failing tests but still got fails. Will have to  look into this later. Claude unable to help.

josephjclark:
So nervous of this PR. It's a tiny change really. Works just great locally. But my tests go beserk - random tests seem to fail to return basic events, for no good reason that I can see. Very hard to debug this stuff because it's cross-process. Not sure.

josephjclark:
If I stare at this any longer I'll go actually mad.

But I might have figured it out.

It's just the asynchronicity in the logging.

If you make the `stringify` algo async, then it'll exhibit exactly the same behaviour and events will stop being emitted. I thought it was a secret exception in the async iterator.

But no - what I think is happening is that the  workflow is finishing while events are still being stringified to be emitted by the child process. But the child process gets killed as soon as the run is over.

We need some tracker somewhere which makes sure that all pending events are properly set.

This behaviour does make me wonder if we have an async hole in prod which is the underlying cause of some rare exits. Could the worker thread be getting closed before messages have been sent?

EDIT: not 100% sold on this theory (just allowing the worker more time beefore closing doesn't help) but I'm sure it's something around the async message processing for sure.  This may also cause events to exit the worker out of sequence, so maybe we need a queue here too, just like in the main engine (same queue?)

josephjclark:
Blocked by #1394

PR #1373, opened 2026-04-14
Bump slack and axios

josephjclark:
Security alerts on axios

Axios comes out of slack api used in our slack integrations

This should resolve it. Maybe need a test before merging. Or at least check the slack changes


josephjclark:
no breaking changes so this should be OK. And if not, all we lose is slack messages

Issue #1072, opened 2025-10-16
Worker: sometimes sends events without a step id


josephjclark:
Strange sequence of events in worker logs:
```
[SRV] ✘ cf698f95-4b16-4a75-9ac4-98e1729f7017 :: step:complete :: ERR: [step:complete] {"step_id":["This field can't be blank."]}
[SRV] ℹ cf698f95-4b16-4a75-9ac4-98e1729f7017 :: run:log :: OK
[SRV] ✘ cf698f95-4b16-4a75-9ac4-98e1729f7017 :: step:start :: ERR: [step:start] {"job_id":["does not exist"]}
[SRV] ✘ cf698f95-4b16-4a75-9ac4-98e1729f7017 :: run:log :: ERR: [run:log] {"step_id":["must be associated with the run"]}
```

How is step_id blank?

This should lead to lost.  I think this run might recover, but interestingly there is a lost run (ea330df2-6b88-4cb3-84b9-4619e5d3c319) at almost the same time as this. Coincidence?

[GCP logs for the run](https://console.cloud.google.com/logs/query;query=resource.type%3D%22k8s_container%22%0Aresource.labels.location%3D%22europe-west6-a%22%0Aresource.labels.pod_name%3D%22global-worker-6876dd568f-44958%22%0Aresource.labels.cluster_name%3D%22swiss-standard-1%22%0Aresource.labels.container_name%3D%22global-worker%22%0Aresource.labels.project_id%3D%22platform-test-267207%22%0Aresource.labels.namespace_name%3D%22prod%22%0A%22cf698f95-4b16-4a75-9ac4-98e1729f7017%22;pinnedLogId=2025-10-09T18:01:49.502329184Z%2F0s47ns2zgw2jjy7h;cursorTimestamp=2025-10-09T18:01:46.216371811Z;aroundTime=2025-10-09T18:30:28.989Z;duration=PT1H?project=platform-test-267207)

josephjclark:
A quick search suggests this is coming up quite a bit - I'm scanning the logs but something in the order of 20 in the last 30 days

josephjclark:
I think I'm going to add some dedicated debugging around this, see what I can work out

josephjclark:
So the step complete event looks like this:
```
{
  "mem": { "job": 110118312, "system": 281231360 },
  "duration": 3631,
  "thread_id": 367,
  "timestamp": "1776442235870123",
[...]

PR #1389, opened 2026-04-23
Worker: Fix batch events being dropped

josephjclark:
## Short Description

Relates to #1072 but it's actually a different fix

## Implementation Details

This is a very claude driven solution. The whole algorithm is bit too claudey. I'm tempted to rethink the whole thing but maybe that's even riskier?

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/stale-lights-occur.md
+++ b/.changeset/stale-lights-occur.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Fix an issue in batch events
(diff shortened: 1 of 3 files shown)


josephjclark:
So this isn't the issue we're seeing in production because
a) this only affects batch logs
b) batch logs is off in production

So this is worth pushing through but its a fix for later, not a fix for now

josephjclark:
Try it out I guess!

PR #1393, opened 2026-04-29
deploy: support webhook_response in triggers

midigofrank:
## Short Description

Support `webhook_response` in webhook triggers

Fixes #1351 

Depends on https://github.com/OpenFn/lightning/pull/4666

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/frank-canyons-report.md
+++ b/.changeset/frank-canyons-report.md
@@ -1,0 +1,5 @@
+---
+'@openfn/deploy': minor
+---
+
+Add support for webhook_response_config
--- a/.changeset/salty-areas-juggle.md
+++ b/.changeset/salty-areas-juggle.md
@@ -1,0 +1,5 @@
+---
+'@openfn/lexicon': minor
+---
+
+Support `cron_cursor_job_id`, `webhook_reply` and `webhook_response_config` in Provisioner types
--- a/.changeset/shy-needles-attack.md
+++ b/.changeset/shy-needles-attack.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': minor
+---
+
+Support webhook responses in sync & deploy
--- a/.changeset/wide-seals-tease.md
+++ b/.changeset/wide-seals-tease.md
@@ -1,0 +1,5 @@
+---
+'@openfn/project': patch
+---
+
+Support more trigger keys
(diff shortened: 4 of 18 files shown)


josephjclark:
Hey @midigofrank - this looks fine for v1, will need a `pnpm changeset` to track the changes.

We need to see some tests for this for v2, and think about versioning. We could spin this out into a new issue.

Check out `integrations-tests/cli/test/deploy.v2.test.ts`. @doc-han can help you navigate. Use `pnpm exec ava test/deploy.v2.test.ts` to just run that file.

Note that the integration tests mostly use the provisioner state format - define a project in provisioner state, then pull it down, then assert on the local structure.

Any changes on a trigger that affect versioning need to be written to the trigger node (and the version algorithm will need updating on both ends, come to think of it). If it doesn't affect versioning, the new keys should spread onto the `.openfn` key on the trigger.

You can probably just add the new keys to one of the existing structures, but it might make sense to create a dedicated test. I'm happy either way.

The other way of looking at this would be to add unit tests to `packages/project`, ensuring the keys are properly serialized in `to-app-state.ts` and `from-app-state.ts`.

midigofrank:
Thanks @josephjclark . @doc-han could you please lend a hand in here? claude is taking in circles

josephjclark:
Let's skip the versioning stuff as it's not implemented in lightning either (but we need an issue to track versioning on both sides!)

josephjclark:
ack @midigofrank I forgot this wasn't done - I've muddled up the two PRs a bit

 Is this still API correct for V1? If so I'll add changes to v2 and merge this down.

josephjclark:
Turns out this is causing version conflicts right now, so I'm going to take it over and merge it down

midigofrank:
Hey @josephjclark , I actually wanted to rename the field to `webhook_response_config` and `WebhookResponseConfig` to avoid type conflict with https://github.com/OpenFn/kit/pull/1404

josephjclark:
@midigofrank ok cool - I'll make those changes here

I've pushed some commits which improve testing and track some keys that aren't being included in the version hash right now. I'm not actually feeling too well so this may be a fever dream :sweat_smile: 

I expect the last commit to fail. I'll get it fixed in the morning

midigofrank:
@josephjclark I can update the version hash on the lightning side as well, I'm updating the branch.
EDIT: and I have another incoming PR on the cli for supporting channels

josephjclark (approved):

midigofrank:
Thanks @josephjclark . Let me merge in lightning's side

josephjclark:
@midigofrank it's not out yet - I'll release this and the other work together soon

PR #1397, opened 2026-05-07
feat(ws-worker): support multi-root @local adaptor resolution

jeremi:
## Summary

Lets `@openfn/ws-worker` resolve `@local` adaptors against a colon-separated `OPENFN_ADAPTORS_REPO`, so a private adaptor monorepo can be loaded alongside the canonical OpenFn adaptors monorepo without forking.

- Single-path values keep working unchanged (a one-element root list).
- Order is precedence: when more than one root contains `packages/<shortName>/package.json`, the earlier root wins.
- Whitespace and empty segments in the colon list are trimmed.
- When no root contains the adaptor, the worker falls back to the first root's resolved candidate path. The directory does not exist, so the runtime still surfaces a recognisable "missing local adaptor" path, and single-path callers see the same behaviour as before this PR (the path was returned without an existence check).

## Companion to Lightning PR

This is the executor-side half of OpenFn/lightning#4714. Without this patch, colon-separated `OPENFN_ADAPTORS_REPO` only widens what Lightning's `AdaptorRegistry` exposes (picker UI, metadata, version validation); the worker still tries to load the literal joined path and dies with `ENOENT`. The two PRs ship together so multi-root @local execution works end-to-end.

The first-wins semantics here mirror Lightning's `AdaptorRegistry`: a private adaptor repo listed before the canonical OpenFn monorepo will override individual adaptors locally, and the registry view and the worker's execution path stay consistent.

## Test plan

- [x] `pnpm --filter @openfn/ws-worker test` — 30/30 `convert-lightning-plan` tests green, including 5 new cases:
  - resolves `@local` against a single existing root
  - walks colon-separated roots in order, first match wins
  - ignores roots that do not contain the adaptor
  - trims whitespace and drops empty segments
  - falls back to the first root when no root has the adaptor
- [x] `pnpm --filter @openfn/ws-worker test:types` clean
- [x] `pnpm prettier --check` clean for the touched files
- [x] Live verification inside a Lightning dev container with the companion Lightning PR applied: a workflow with `@openfn/language-publicschema@local` was resolved by the patched worker to `/private-adaptors/packages/publicschema/dist/index.js` (the first root in the colon list), and the runtime logged `Resolved adaptor @openfn/language-publicschema to version 0.0.1`.

## Changeset

`minor` bump for `@openfn/ws-worker`.

### DIFF
--- a/packages/ws-worker/src/util/convert-lightning-plan.ts
+++ b/packages/ws-worker/src/util/convert-lightning-plan.ts
@@ -1,4 +1,5 @@
 import crypto from 'node:crypto';
+import fs from 'node:fs';
 import path from 'node:path';
 import type {
   Step,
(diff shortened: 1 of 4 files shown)


packages/ws-worker/src/util/convert-lightning-plan.ts
```diff
@@ -50,18 +51,49 @@ export default (
 ): { plan: ExecutionPlan; options: WorkerRunOptions; input: Lazy<State> } => {
   const { collectionsVersion, monorepoPath } = options;
 
+  // monorepoPath is a colon-separated list of monorepo roots, mirroring how
```
josephjclark: The most obvious problem that strikes me here is that this will break on windows machines.

It's convenient that the colon approach mimics a unix path. And I'm not aware we have many windows users doing this stuff - although occasionally someone crops up.

Anyway I would prefer a comma or semi-colon separated list here so that it's safer

packages/ws-worker/src/util/convert-lightning-plan.ts
```diff
@@ -50,18 +51,49 @@ export default (
...
+  const resolveLocalAdaptorPath = (shortName: string): string | undefined => {
+    if (monorepoRoots.length === 0) return undefined;
+
+    for (const root of monorepoRoots) {
+      const candidate = path.resolve(root, 'packages', shortName);
+      if (fs.existsSync(path.join(candidate, 'package.json'))) {
+        return candidate;
+      }
+    }
+    // Fall back to the first root's resolved candidate path. The directory
```
josephjclark, line 75: Way too many comments  here, and in the wrong place. But I can tidy that up before merge to main. I presume these are AI generated?

packages/ws-worker/test/util/convert-lightning-plan.test.ts
```diff
@@ -7,6 +10,17 @@ import type {
 import convertPlan from '../../src/util/convert-lightning-plan';
 import { Job } from '@openfn/lexicon';
 
+// Builds a temporary monorepo root with package.json files in each named adaptor.
+const makeMonorepo = (adaptors: string[]) => {
```
josephjclark, line 14: Not wild about the file IO stuff in these tests to be honest. Mostly in this repo we use the `mock-fs` library to mock out the file system. If you're using an AI agent can you ask it to regenerate like that?

josephjclark (commented):
Hi @jeremi, just taking a closer look at this. The solution is basically fine but it strikes me that this approach won't work for windows users, who have paths like `c:/repo/adaptors`.

I think a comma separated list will do just as well, and be a bit more accessible to other users

jeremi:
Thanks @josephjclark — addressed in 66e889bb:

- Switched the split separator from `:` to `,` so Windows drive-letter paths like `c:/repo` stay usable; single-path callers (the common case) are unchanged.
- Updated the comment block, CLI help text, changeset blurb, and the dedicated multi-root tests (titles + literals).
- Companion change on the registry side: https://github.com/OpenFn/lightning/pull/4714

Ready for another look.

josephjclark:
Alright @jeremi this looks great.

One more issue I see: we'll also need to update the CLI so that it can support comma-separated paths on `OPENFN_ADAPTORS_REPO`.   That need to be done before I can release this.

Are you able to take a look at that, or shall try to resolve it for you?

josephjclark (approved):

jeremi:
Thanks @josephjclark !

PR #1404, opened 2026-05-11
feat: forward `state.webhookResponse` on `step:complete`

### LINKED ISSUE #1399: Add `webhook_reponse` in `StepCompletePayload`
### Implementation notes
- The worker reads the `event.state` object of `step:complete` in `src/events/step-complete`
- It looks for a `webhookResponse`
- If it IS set, update the `step:complete` payload we send to lightning to include `webhook_response`

For more info, look at: https://github.com/OpenFn/lightning/pull/4666#issuecomment-4388242230

midigofrank:
## Short Description

Job code sets state.webhookResponse = { status, body }; the worker forwards it on the step:complete payload as webhook_response. The key is stripped from the dataclip (worker) and from the next step's input state (runtime) so it cannot leak downstream.

Fixes #1399

## QA Notes

Can be tested via https://github.com/OpenFn/lightning/pull/4666

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Couple of notes from me:
1. Don't touch any of this stuff in the runtime
2. Don't delete the webhookResponse key from state (anywhere)
3. basically the only diff is on the worker's step complete payload assembly

josephjclark (approved):
Lovely and light and clean - nice work @doc-han 

I'll release this now

packages/ws-worker/src/events/step-complete.ts
```diff
@@ -42,6 +42,23 @@ export default async function onStepComplete(
     state.inputDataclips[nextJobId] = dataclipId;
   });
 
+  let webhook_response,
+    response_data = event.state?.webhookResponse;
+
+  // validating structure of response & add status if it's a number
+  if (
+    response_data &&
```
josephjclark: Just for the record @doc-han you can write this condition as 

```
if (typeof response-data?.body === 'object')
```

I'll do that before I push

josephjclark: Ok @doc-han I actually realised two things:

1. if the user sends a status but not a body (quite likely!) the response won't get send
2. The code is actually very over-complex - let's just write the user's `webhookResponse` straight to `webhook_response`

doc-han: Thanks. resolved

doc-han (approved):

packages/ws-worker/src/events/step-complete.ts
doc-han: Oh. that message came in late. the status thing isn't resolved

PR #1408, opened 2026-05-14
Types rethink

josephjclark:
Rethinking types a bit in the v1 deploy package, which I don't like to touch

But we have too much type duplication between lexicon and deploy. Claude has fixed it for me, I think

This will help ensure better type consistency

### DIFF
--- a/packages/deploy/src/types.ts
+++ b/packages/deploy/src/types.ts
@@ -1,13 +1,6 @@
 import { Provisioner } from '@openfn/lexicon/lightning';
 
-export type StateJob = {
-  id: string;
-  name: string;
-  adaptor: string;
-  body: string;
-  project_credential_id: string | null;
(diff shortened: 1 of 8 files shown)


packages/deploy/src/types.ts
```diff
@@ -14,138 +9,127 @@ export type SpecJobBody =
...
   description: string;
   workflows: Record<string | symbol, WorkflowState>;
   project_credentials: Record<string | symbol, CredentialState>;
   collections: Record<string | symbol, CollectionState>;
+  scheduled_deletion?: string | null;
+  history_retention_period?: string | null;
+  dataclip_retention_period?: string | null;
+}
+
+export interface ProjectSpec
```
josephjclark, line 114: this doens't work for me at all. It's too hard to reconcile these  types right now, sadly

PR #1416, opened 2026-05-19
Optional cross-worker lock for shared adaptor repo

### LINKED ISSUE #1414: Worker: allow multiple workers to safely share a repo directory
## Problem

If you set `WORKER_REPO_DIR` to a path on a shared volume (e.g. an NFS mount, k8s PVC, or a shared docker volume), the intent is obvious: persist the adaptor
cache so workers don't re-download everything on every restart, and share it
across multiple workers so we're not duplicating gigabytes per pod.

This *almost* works today. A single worker pointed at a persistent volume
behaves correctly — restarts are fast, no redundant installs. But the moment
you point more than one worker at the same volume, things go sideways:

- The in-memory install queue in engine-multi only serialises installs within
  one process. Two workers seeing the same uninstalled adaptor will both shell
  out to `npm install` against the same `node_modules/` and `package.json`.
- npm has no awareness of other npm processes writing the same tree. Result:
  corrupt `node_modules`, half-written `package.json`, runs failing with
  module resolution errors that are a pain to diagnose (see #503).

## What we want

Multiple workers should be able to share one repo directory safely. The common
case (adaptor already installed) should stay fast — no coordination overhead
when there's nothing to do. The common but only once case (something needs installing)
should end up with a single install across all the workers sharing the volume,
with the others waiting.

Crash safety matters: if a worker dies mid-install, the next one shouldn't be
stuck forever waiting on a ghost.

## Out of scope

- Anything that requires external coordination (Redis, etcd, etc.). Keep it
  filesystem-only so it works wherever a shared volume works.
- Changing the CLI's behaviour.
- Garbage-collecting old adaptor versions from the cache.

## Related

- #919 — ephemeral storage pressure in workers. A shared persistent cache is
  one way to take pressure off pod-local disk, so this work helps there too.
- #73 — closed, but the in-process precedent. This is essentially the
  multi-process / multi-host version of that.

stuartc:
## Short Description

Adds a filesystem lock so multiple ws-workers can safely share a single adaptor repo directory (e.g. an NFS mount or k8s PVC), without two workers racing on the same `npm install`.

Fixes #1414

## Implementation Details

When the repo directory is shared between worker pods, two workers can otherwise hit `install` for the same adaptor at the same time and corrupt each other's `node_modules`. The lock lives in `engine-multi`'s autoinstall flow — that's where the install actually happens, so that's where the coordination belongs.

How it works:

- A new `withInstallLock` helper in `packages/engine-multi/src/util/repo-lock.ts` acquires a per-adaptor lockfile under `<repoDir>/.locks/<alias>.lock` via `proper-lockfile`. Different adaptors don't block each other.
- Autoinstall wraps its existing `install` call with the lock. Inside the lock it re-checks `isInstalled` and skips the install if another worker has just finished it.
- The engine's `isInstalled` is now stricter: it ANDs the existing repo `package.json` dependency check with a stat of `node_modules/<alias>/package.json`. npm extracts each package atomically, so the presence of that file is itself proof of completion. This also fixes a latent bug where a worker killed mid-install could leave the repo `package.json` saying "installed" while `node_modules` is half-written, and the next worker would happily import broken code.
- The cache-hit path stays lock-free — `isInstalled` returns true without touching the lock directory.
- If `install` throws, the lock is released and the next worker will retry.
- Stale-lock recovery: `proper-lockfile`'s mtime heartbeat updates every 5s and another worker considers a lock stale after 5 min. The retry ceiling (6 min) sits above that, so when a pod dies mid-install its lock expires before waiting workers give up.
- When a worker has to wait for the lock, one info-level log line fires so the cause of a slow cold-start is visible without enabling debug logging.

On by default. Disable with `--no-repo-lock` / `WORKER_NO_REPO_LOCK=true`. Single-worker deployments don't need to do anything to opt out — they'll just never hit contention and so never log anything new.

New dep on `proper-lockfile@4.1.2` (with `graceful-fs`, `retry`, `signal-exit` transitives) in `engine-multi`. All checked for CVEs — clean.

## QA Notes

Interesting cases to exercise (with `--repo-dir` set):

- Two workers pointing at the same `--repo-dir` starting a run that needs the same adaptor at the same time — only one `npm install` should actually run; the other should see the install completed during its wait and skip.
- A worker killed mid-install (SIGKILL) — the next worker should recover after the stale window (5 min) rather than getting stuck forever.
- Different adaptors should install in parallel, not serialise.
- With `--no-repo-lock`, multi-worker behaviour should be unchanged from before this PR.
- Single-worker behaviour with the default `--repo-lock` should be indistinguishable from before — no contention, no extra logs.

There's a multi-process test suite using `child_process.fork` that exercises the cross-worker cases deterministically — `pnpm test test/util/repo-lock.test.ts` from `packages/engine-multi`.

K8s deployments using this need NTP/chrony across nodes — stale-lock detection is mtime-based.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI


packages/ws-worker/src/start.ts
```diff
@@ -119,6 +120,20 @@ if (args.mock) {
...
     profilePollInterval: args.profilePollIntervalMs,
   };
+
+  if (args.repoLock) {
+    if (args.repoDir) {
+      logger.info(
+        'Repo lock enabled: coordinating adaptor installs via filesystem lock at',
+        args.repoDir
+      );
+      engineOptions.autoinstall = createLockedHandlers(args.repoDir);
```
josephjclark: I'm surprised to see all this stuff in the worker - isn't it more a concern for the engine, which actually handles the auto install? Or even the runtime, which provides the actuall  npm install hook.

I've also just checked the runtime install code and it's doing quite a lot of stuff which is going to be duplicated or removed by the implementation here

packages/ws-worker/src/util/repo-lock.ts
```diff
@@ -0,0 +1,150 @@
...
+  await rename(tmp, sentinel);
+};
+
+export type InstallFn = (
+  specifier: string,
+  repoDir: string,
+  logger: Logger
+) => Promise<unknown>;
+
+const defaultInstall: InstallFn = (specifier, repoDir, logger) =>
```
josephjclark: Ah so THIS guy calls out to the runtime's install

So this file isn't really a repo lock at all - it's an "install adaptor WITH repo lock"

I still think this logic should go down into the runtime, but I also think the lock logic and the actual install should be better separated. Very misleading to have `repo-lock.ts` actually take responsibility for the install!

packages/ws-worker/src/start.ts
stuartc: Oh interesting, I actually intentionally stayed away from engine - after some consternation; thought that the idea of parallel installs happen only in the worker and thought to keep it out . Happy to redo it, or move the repoLock stuff into engine? Thoughts on where you see it fitting? (I totally get the knowledge leak in the lock wrapper, since it needs to know a fair bit about the package and path.

packages/ws-worker/src/util/repo-lock.ts
stuartc: You reckon I move the lock wrapper to runtime or engine? My gut says runtime, but now it feels like installing things should be in runtime? 😅

packages/ws-worker/src/start.ts
josephjclark: Well the parallelisation happens in the engine - worker requisitions work, engine parallelises it. So by that logic engine is the right place.

But given that the actual install function is in the runtime I still think it should go there. That or we intercept the install in the engine (kinda like we do now)

Let me think about it some more

packages/ws-worker/src/util/repo-lock.ts
```diff
@@ -0,0 +1,150 @@
...
+  runtimeInstall([specifier], repoDir, logger);
+
+export const createLockedHandlers = (
+  repoDir: string,
+  installFn: InstallFn = defaultInstall
+) => {
+  const locksDir = path.join(repoDir, '.locks');
+  const sentinelsDir = path.join(repoDir, '.sentinels');
+
+  const ensureDirs = (async () => {
```
josephjclark: This is weird. Why not remove the wrapper and just await the `mkdir` calls?

josephjclark:
ok @stuartc having looked a bit more closely  - this implementation MUST move into the engine

The engine already has an `autoinstall.ts`. You should just extend that to use the lockfile stuff before calling the runtime's install.  I think you probably want the lock and install logic in two different files, but that's up to you.

What is the sentinel stuff all about?

packages/ws-worker/src/util/repo-lock.ts
```diff
@@ -0,0 +1,150 @@
...
+        throw new Error(
+          `Lock acquisition timed out after ${waitSecs}s waiting for ${alias}; another worker likely still installing (lock: ${target})`
+        );
+      }
+      throw e;
+    }
+    logger.debug(`acquired install lock for ${specifier}`);
+
+    try {
+      const [hasSentinel, hasPkg] = await Promise.all([
```
josephjclark: As we're going to move this to the engine, the engine already has an `isInstalled` function, which determines whether an adaptor is installed or not.

Rather than implement that test twice - as we're doing here - I'd like tot ry and re-use that logic if possible.

stuartc:
Thanks @josephjclark - all fair points, addressed.

**Lock lives in the engine now.** New file `packages/engine-multi/src/util/repo-lock.ts` exports a single `withInstallLock(repoDir, alias, logger, fn)` helper - pure lock primitive, nothing else. Autoinstall wraps its existing `install` call with it. The worker doesn't touch lock files at all any more; it just flips `autoinstall.lockRepo` via a `--no-repo-lock` flag. Lock is on by default.

**Sentinels are gone.** Those were the `.sentinels/<alias>.done` marker files - written atomically once `npm install` returned successfully, so the install handler could check them as a "this finished cleanly" signal. They're redundant: npm extracts each package atomically, so the presence of `node_modules/<alias>/package.json` is enough. I've tightened the engine's existing `isInstalled` to AND its package.json dependency check with a stat of that file - so it's now a check of disk state rather than just npm's book-keeping. That catches manual `node_modules` cleanup, NFS visibility lag on shared volumes, and probably some mid-install crash cases too (haven't dug into npm's exact write ordering to be sure). Benefits every deployment, not just locked ones.

**No duplicated `isInstalled`.** Engine's `isInstalled` is the single source of truth now - initially this  ws-worker version had its own sentinel + node_modules check inline, which duplicated the engine's pkg-deps check. The in-lock double-check (re-run `isInstalled` after acquiring, so a worker that queued behind another one's install sees the install has completed during the wait and skips its own attempt) is preserved - just calling the engine's check rather than a parallel implementation.

**`ensureDirs` IIFE is gone** - replaced with a small `ensureLockTarget(target)` helper that `mkdir`s the parent and touches the lock file. No IIFE-on-first-call dance.

**Two files** as you suggested: `util/repo-lock.ts` (lock primitive) and `api/autoinstall.ts` (orchestration + install). Runtime is untouched - engine still calls the runtime's `install` exactly as before, just inside a lock.

A few smaller things came out of the review:

- Security/alias guard in `withInstallLock` - refuses aliases where `..` appears as a path segment, or that are absolute, on top of the upstream whitelist.
- Contention is logged at info level (one line, only when actually contended) so ops can see why a worker is sat waiting on cold start without needing debug logs on.
- Dropped a dead `pending` map in `autoinstall.ts` left behind by your 2023 single-queue refactor - declared but never written to, just a stray `delete pending[a]` in the error path.
- Caught a yargs auto-negation bug while wiring the flag up: `--no-repo-lock` was being parsed as `{ repoLock: false }` rather than `{ noRepoLock: true }`, so the original wiring was a no-op. Reversed the option and let yargs do the negation properly. Tests now cover all four combinations of CLI flag + env var.
- Pinned per-call options (`context`/`repoDir`/`logger`/`useLock`) on the autoinstall queue entry. They were captured in the *first* caller's closure under the original code, which is fine today (engine is a singleton with stable options) but would silently misbehave the moment anyone wires per-run autoinstall overrides or proxies `AUTOINSTALL_COMPLETE`/`ERROR` through `createWorkflowEvents`. Cheap to fix now, expensive to debug later.

Changeset updated to declare both `@openfn/engine-multi` and `@openfn/ws-worker` as minors.

.changeset/distributed-adaptor-install-lock.md
```diff
@@ -0,0 +1,6 @@
+---
+'@openfn/engine-multi': minor
+'@openfn/ws-worker': minor
+---
+
+Add a filesystem-based lock so multiple workers can safely share a single adaptor repo directory (e.g. an NFS mount or k8s PVC). The lock lives inside engine-multi's autoinstall and serialises installs across processes via a per-adaptor lockfile. It is enabled by default and can be disabled with `WORKER_NO_REPO_LOCK=true` (or `--no-repo-lock`) on ws-worker. The engine's `isInstalled` check has also been tightened to verify the adaptor's `node_modules` entry exists on disk, not just the repo `package.json` dependency entry — this catches half-installed adaptors left behind by a crashed worker.
```
josephjclark, line 6: I prefer the release notes to be way more terse. No explanation, no justification - just the headline. Maybe a migration guide if breaking but obviously that doesn't apply here

packages/ws-worker/src/util/cli.ts
```diff
@@ -306,6 +313,11 @@ export default function parseArgs(argv: string[]): Args {
       'ws://localhost:4000/worker'
     ),
     repoDir: setArg(args.repoDir, WORKER_REPO_DIR),
+    repoLock:
+      args.repoLock ??
```
josephjclark, line 317: This doesn't work with `setArg`?

I expect ` setArg(args.repoLock, WORKER_NO_REPO_LOCK, true)` to work

I suppose we might need a bit more logic to track the boolean but I doubt it, and we don't need to be strict or fussy

packages/ws-worker/src/start.ts
```diff
@@ -108,6 +108,13 @@ if (args.mock) {
     engineReady(engine);
   });
 } else {
+  const lockRepo = args.repoLock !== false;
+  if (lockRepo && !args.repoDir) {
+    logger.warn(
+      'WARNING: repo lock is enabled but --repo-dir is not set; lock will be a no-op'
```
josephjclark, line 114: repodir has a default value so I don't think this is true?

packages/engine-multi/src/api/autoinstall.ts
```diff
@@ -13,98 +15,102 @@ import type { Logger } from '@openfn/logger';
...
-    fn: string,
-    repoDir: string,
-    logger: Logger
-  ): Promise<boolean>;
+  lockRepo?: boolean;
   versionLookup?: (specifier: string) => Promise<string>;
 };
 
-const pending: Record<string, Promise<void>> = {};
+// Per-entry options are pinned at enqueue time so the worker that drains the
```
josephjclark, line 27: Is this comment valid? Looks random?

packages/engine-multi/src/api/autoinstall.ts
```diff
@@ -13,98 +15,102 @@ import type { Logger } from '@openfn/logger';
...
-          module: name,
-          version: version!,
-          duration,
-          message: e.message || e.toString(),
+    const startTime = Date.now();
+    try {
+      if (useLock) {
+        // Re-check inside the lock so we skip work a peer worker
+        // completed while we were waiting.
+        await withInstallLock(repoDir, getAliasedName(a), logger, async () => {
```
josephjclark, line 66: hmm. Why is the last argument a callback? We can't just let the promise return?

packages/engine-multi/src/util/repo-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const withInstallLock = async (
+  repoDir: string,
+  alias: string,
+  logger: Logger,
+  fn: () => Promise<void>
+): Promise<void> => {
+  // Defence-in-depth: refuse aliases that could escape the .locks directory.
+  // Upstream whitelist filtering should make this unreachable.
```
josephjclark, line 36: Should it? Where?

If this is true I expect to see some shared logic which sanitises aliases to be safe

packages/engine-multi/test/util/repo-lock.unit.test.ts
```diff
@@ -0,0 +1,43 @@
+/**
```
josephjclark, line 1: This is weird. Just stick the test in `repo-lock.test.ts` so that a) it can be seen and b) there isn't this weird "othering" of unit tests

josephjclark (commented):
Posting as far as I've got to but not done looking through this.

I like that things feel cleaner and less impactful here. You've got a lot of the logic and testing contained to `repo-lock.ts`

Worried that the test framework looks very heavy and intense. I suppose it has to be for something like this, which is very process-y. I'll take a closer look and have a think about it it.

I'd love to see some kind of integration test for this in `integration-tests/worker`. I don't know if we could build something reliably - but we do already have a bunch of autoinstalll tests. But maybe we need the process tests because actual autoinstall tests are always going to be flaky?

packages/engine-multi/src/util/repo-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+  if (alias.split(/[/\\]/).includes('..') || path.isAbsolute(alias)) {
+    throw new Error(`Invalid alias for install lock: ${alias}`);
+  }
+
+  const locksDir = path.join(repoDir, '.locks');
+  const target = path.join(locksDir, `${alias}.lock`);
```
josephjclark, line 42: since locksDir isn't used anywhere else you can simplify to
```
const target = path.join(repoDir, '.locks', `${alias}.lock`);
```

packages/engine-multi/src/util/repo-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  await ensureLockTarget(target);
```
josephjclark, line 44: Hold up, this function creates a lock file at the target path. But that's exactly what `lockfile.lock` does for us right?

So how does that work? Doesn't `lockfile` just assume the file is  always locked because we've just created the lock?

packages/engine-multi/src/util/repo-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+    release = await lockfile.lock(target, { ...lockOpts, retries: 0 });
+  } catch (e: any) {
+    if (e.code !== 'ELOCKED') throw e;
+
+    logger.info(
+      `waiting for install lock on \`${alias}\` (another worker is installing)`
+    );
+
+    try {
+      release = await lockfile.lock(target, {
```
josephjclark, line 61: So we call lockfile.lock twice here. Once I guess to see if we can aquire the lock - and if we can't, we print this log line and then try again with a number of retries?

First observation is that we're then in effect retrying the lock LOCK_RETRY_OPTIONS + 1 times

Second observation is - is this worth it? Are we better calling `lockfile.check()` first, and then doing the log-and-retry call if a lock already exists?

packages/engine-multi/src/util/repo-lock.ts
```diff
@@ -0,0 +1,88 @@
...
+import type { Logger } from '@openfn/logger';
+
+// k8s CPU throttling can delay the heartbeat setTimeout; 5 min avoids false-stale on tight clusters
+const STALE_MS = 5 * 60_000;
+// Retry ceiling exceeds STALE_MS so a dead lock-holder's stale window expires before we give up
+const MAX_LOCK_WAIT_MS = STALE_MS + 60_000;
+const LOCK_INTERVAL_MS = 2_000;
+const UPDATE_MS = 5_000;
+
+const LOCK_RETRY_OPTIONS = {
```
josephjclark, line 13: Surprises me to see this declared as a constant. I get the convention but an object isn't really a constant. Just feels wierd.

Also none of this is configurable? The lock uses hard-coded values and that's it? Not a blocker for me but we might want to open an issue to make this configurable later.

Not because I _want_ to configure it today - but because it's weird that I am unable to.

josephjclark (commented):
I've been looking through a bit more closely at the code and tests. I really don't love those tests: there's a lot of business logic in that worker and I'm a bit nervous that we're testing test code rather than actual logic. 

I'm taking a break for now but I think there are a few things here we should talk about before I look deeper

PR #1420, opened 2026-05-21
Don't commit credentials.yaml in gh sync

### LINKED ISSUE #1419: CLI: make credentials.yaml safer
we don't want users to accidentally check in credentials.yaml with real credentials

It is designed 

How do we best guard for this?

- if in a git repo, auto add it to gitignore
- don't generate it by default (prompt to generate at runtime maybe?)

josephjclark:
## Short Description

This PR ensures that the v2 sync proxy used by GH sync does NOT generate a credentials.yaml file.

Users are free to use this of course, but it should not be checked in to GH by default

Fixes #1419 

A lot of the work here has been in updating the pull and deploy tests so I can cover credentials.yaml usage

## QA Notes

Needs testing against prod one relesaed

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/packages/cli/src/pull/handler.ts
+++ b/packages/cli/src/pull/handler.ts
@@ -27,10 +27,10 @@ async function pullHandler(options: PullOptions, logger: Logger) {
       'openfn.yaml'
     );
     if (!process.env.PREFER_LEGACY_SYNC && (await fileExists(v2ConfigPath))) {
-      // override endpoint with one from openfn.yaml
-      const config = yamlToJson(await fs.readFile(v2ConfigPath, 'utf-8')) ?? {};
-      if (config?.project?.endpoint) {
-        config.endpoint = config.project.endpoint;
+      // default endpoint to one from openfn.yaml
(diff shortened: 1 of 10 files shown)


josephjclark:
Weird test fails here making me nervous. I'll pick this up in the morning with a clear head.

packages/cli/src/pull/handler.ts
```diff
@@ -42,6 +42,8 @@ async function pullHandler(options: PullOptions, logger: Logger) {
           project: options.projectId,
           force: true,
           endpoint: config.endpoint,
+          apiKey: config.apiKey,
```
josephjclark: I don't understand why I had to add this (it hasn't worked anyway), but I'm also a bit surprised it wasn't there any way. I  guess GH sync uses an env var so its not really needed.

But I think I want to leave it in

josephjclark:
Ok, works great pulling and pushing to staging.openfn.org using the v1 commands and an openfn.yaml file. Confident that this small change is very safe.

PR #1440, opened 2026-06-03
chore: ignore sentry errors by regexp

### LINKED ISSUE #1429: Worker: don't report oauth errors to sentry
A very commonly occurring error in the worker is `oauth token has expired` triggered by `fetch:credential`

This is totally harmless and not our fault. It's a userland issue. Nothing to see here officer.

But also this issue should not be reported to sentry. We should just gracefully, intentionally crash the run. No alerts, no warnings, just log to the user the reason we've crashed.

We could just archive this on sentry but I don't think I can filter the message very well there. If  a different error comes out of the same stack, I want to know about it.

[sentry reference](https://openfn.sentry.io/issues/6512942563/?environment=prod&project=4508936084848640&query=&referrer=issue-stream)

doc-han:
## Short Description

This PR allows us to ignore specific errors from going through sentry. 
We currently don't want Oauth error to pass through sentry. hence we use a beforeSend hook in sentry to filter the errors 

(in the image below, the error is intercepted and logged to not be reported to sentry)
ha

Fixes #1429 

## Implementation Details

## QA Notes

On Lighting there's a branch that simulates Oauth always failing
Branch: https://github.com/OpenFn/lightning/tree/ignore-sentry-errors (`ignore-sentry-errors`)

1. Run lighTning without a work (That's with RTM=false)
2. Build and run the worker pointing to that lightning server (remember to set the secret to match that of lightning)
3. To see Sentry debug logs, update the code to have `debug:true` at https://github.com/OpenFn/kit/blob/8e29d122daab29c07381d10d9d9fd777fcbbf3a8/packages/ws-worker/src/server.ts#L249-L262
4. Create a workflow and pick a credential for one of the nodes in lightning
5. Run the workflow and then check the worker logs
6. You should see the Oauth error captured in the logs but sentry debug logs should notify that this specific error isn't going to be reported.

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/.changeset/honest-maps-know.md
+++ b/.changeset/honest-maps-know.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Add the ability to ignore and override errors messsages before reporting to sentry
--- a/packages/ws-worker/src/events/run-error.ts
+++ b/packages/ws-worker/src/events/run-error.ts
@@ -6,6 +6,7 @@ import { RUN_COMPLETE } from '../events';
 import { Context, onJobError } from '../api/execute';
 import logFinalReason from '../util/log-final-reason';
 import { sendEvent } from '../util/send-event';
+import { getIgnoredErrorSeverity } from '../util/ignored-errors';
 
 export default async function onRunError(
   context: Context,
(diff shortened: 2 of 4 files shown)


josephjclark:
@doc-han this looks great, I love the approach

One adjustment I'd like (which might change the approach) is that ideally the workflow should exit with state Crashed

image

I'd also love to see this reported on the `status-codes` docs page

image

Can you have a think about what it would take to do that? I think it means you still might have to catch the exception earlier in the process

josephjclark (approved):

PR #1443, opened 2026-06-04
Openfn compile for unit tests

### LINKED ISSUE #1424: Compile a project for unit tests
Users would like to be able to unit test code in the workflows.

Testing adaptor logic is super hard because those require system integrations - those are integration tests.

But the pure functions within a workflow can and should be unit testable

# The Problem

Users cannot run unit tests on job code right now without effort. Job expressions are not executable javascript: try to import them into mocha or jest and they'll throw exceptions.

Compiling the code (using the CLI for example) gets you a lot closer: now you have import statements and the code can be safely imported into another JS file.

But you can only compile a single workflow or expression at a time, and the devx of this is generally not very good.

# Solutions

These can be done in any order I think?

## 1. Compile whole projects

The `openfn compile`  command can right now only compile a workflow or expression.

We want to be able to just run `openfn compile` and it'll automatically compile all workflows the current repo and write them to disk.

Compiled js files should be emitted to a folder like `compiled` or `dist` at the top of the project. Optionally uses can provide an output path (usually the CLI uses `-o` for output, so make sure we can accept a folder here. We might also add an option to `openfn.yaml` for the default compile path.

Refer to how `cli execute` works as a guide. The `projects` package knows how to parse and read a project file system and return the active workflows. Use that.

## 2. Add a compiler flag to emit only pure JS

It may be desirable to strip operation code, and maybe run other optimisations, while compiling code for testing. Perhaps we only want to compile exported  value. This would help skip large data structures from compilation and keep the compiled code generally tidier.

To do this, a new flag like `--test` (or something else, if that's confusing) needs to be added to the command. This should work when compiling steps, workflows or projects.

I would recommend that import statements are still added to the pure JS files. For example if using the exported `dateFns` function of adaptor, the compiled code should still `import { datefns } from '@openfn/language-common'`

## 3. Add `--watch` command to compile

Compiler should listen for changes on step and workflow files and re-compile when they are modified.

## 4. Document

Once verified this approach should be documented somewhere for other users to follow.

Consider adding to the project template.

mtuchi:
## Short Description

Update `openfn compile` to compiles workflows job expressions to standard JavaScript, writing output to `compiled/` by default.

Fixes #1424 

## Implementation Details

  - `openfn compile` now compiles all workflows in the project to `compiled/` by default (no flag needed)
  - `openfn compile <workflow-name>` looks up a workflow by name in the project and compiles it to `compiled/` use -`O` for stdout or `-o <dir>` to redirect output
  - `openfn compile <file.js>` and openfn compile `<workflow.json>` retain their existing behaviour (stdout by default)
 
  - Added new exports-only transformer (`src/transforms/exports-only.ts`) with a single responsibility: keep imports, named export declarations, and their transitive non-exported dependencies; drop everything else (operations, export default, bare declarations). Runs at order: 0 (before
  all other transformers). Is a no-op unless explicitly enabled.
  - Added `--exports-only` flag: strips adaptor operation calls and keeps only exported declarations — intended for unit testing job code

## QA Notes

**Setup**

Check out the `1424-unit-tests` branch and install `openfnx`:

```bash
pnpm install:openfnx
```

Verify the correct version is installed — `openfnx --version` should output `branch/1424-unit-tests`.

**Testing `openfn compile`**

Navigate to a local OpenFn project containing an `openfn.yaml` file and test the new `openfn compile` command. Available options:

```bash
# Compile all workflows in the project (full compilation, preserves operations)
openfn compile

# Compile all workflows, stripping operation calls (useful for unit testing)
openfn compile --exports-only

# Compile a single workflow by name
openfn compile my-workflow

# Compile a single workflow to a custom directory
openfn compile my-workflow -o tests/

# Compile a single job expression (prints to stdout)
openfn compile workflows/my-workflow/step-a.js --exports-only

# Watch mode: recompile whenever source files change
openfn compile --exports-only --watch
```
## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
@mtuchi I'm  not liking the `--tests` argument. What we really want here is just to easily compile some or all of a project to disk. Stripping the operations is an optional step. As mentioned in the issue I'd call that `compile --exports-only`.

When you do `openfn compile workflow.json` today, it'll compile that workflow and print the result to stdout.

When you do `openfn compile my-workflow`, what will happen? It _should_ lookup the workflow called `my-workflow` in the project, compile that, and print it to stdout. It doesn't, but that would be consistent.

When you do `openfn compile` it _should_ compile the whole project. It can't print that to stdout because it'll compiling multiple files. But it can compile to `dist` and write all the files there. That is what I expect the default behaviour to be.

I'd also suggest that `openfn compile my-workflow` should compile to disk by default (the user can pass -O to write to stdout). That's a little inconsistent with `openfn compile workflow.yaml` but it IS consistent with compiling through the project. So I'm OK with it. Maybe later, in CLI 2.0, we'll _always_ compile to disk for consistency. But since this stuff is very rarely used I wouldn't worry too much.

So drop the `--tests` flag please and just make `openfn compile` compile a project to a folder on disk.

packages/compiler/src/transforms/top-level-operations.ts
```diff
@@ -16,10 +16,70 @@ export type ExtendedProgram = NodePath<
 
 export type TopLevelOpsOptions = {
   // Wrap operations in a `(state) => op` wrapper
-  wrap: boolean; // TODO
+  wrap?: boolean; // TODO
+  // Strip operations instead of moving them into the export array (for test compilation)
+  strip?: boolean;
```
josephjclark: I am not sold that this is the correct implementation

This transformer's one job is to identify operations in the code.

What we need to add here is a different transformer which strips all non-exported lines from code. It's a totally unrelated thing really (OK, as it goes, all operations would be removed - but that's fine). The compiler is simply looking for export statements and removing everything else. The logic would be a lot simpler.

So this would be new transformer which runs probably before all the others

mtuchi: I have added new exports-only transformer (`src/transforms/exports-only.ts`) + unit tests

mtuchi:
QA walk through video for @hunterachieng 
https://www.loom.com/share/033dc7f75c5642a1b5e8edb8a9a35f46

packages/cli/src/compile/compile.ts
```diff
@@ -1,17 +1,25 @@
...
 import type { SourceMapWithOperations } from '@openfn/lexicon';
+import { Workspace } from '@openfn/project';
 
 import createLogger, { COMPILER, Logger } from '../util/logger';
 import abort from '../util/abort';
 import type { CompileOptions } from './command';
 
 export type CompiledJob = { code: string; map?: SourceMapWithOperations };
 
+export const hasExportableCode = (code: string): boolean =>
```
josephjclark: what is this here for?

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +187,105 @@ export const loadTransformOptions = async (
...
+      };
+
+      const { code } = await compileJob(
+        expression,
+        stepOpts,
+        log,
+        (step as any).name ?? step.id
+      );
+
+      if (opts.exportsOnly && !hasExportableCode(code)) {
```
josephjclark: The compiler should decide whether there's any exportable code - not the CLI. 

The CLI could choose to not write empty files, that would be fine. But we can't have it  do compiler stuff.

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +187,105 @@ export const loadTransformOptions = async (
...
+      process.exit(1);
+    }
+  }
+
+  const outPaths: string[] = [];
+  const stalePaths: string[] = [];
+
+  for (const workflow of workflows) {
+    for (const step of workflow.steps) {
+      const expression = (step as any).expression;
```
josephjclark: typing is very messy here. Can we `continue` if there's no expression in step, and otherwise cast step properly?

or, if we have to cast to `any`, let's just do it once at the top of the loop. But I don;t think that should be neccessary

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +187,105 @@ export const loadTransformOptions = async (
...
+      if (!expression || typeof expression !== 'string') continue;
+
+      const adaptor: string | undefined =
+        (step as any).adaptor ?? (step as any).adaptors?.[0];
+      const stepOpts: CompileOptions = {
+        ...opts,
+        adaptors: adaptor ? [adaptor] : opts.adaptors ?? [],
```
josephjclark: This is very complex! Can we simplify?

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +187,105 @@ export const loadTransformOptions = async (
...
+        stepOpts,
+        log,
+        (step as any).name ?? step.id
+      );
+
+      if (opts.exportsOnly && !hasExportableCode(code)) {
+        const stalePath = compiledDir
+          ? path.join(compiledDir, workflow.id, `${step.id}.js`)
+          : null;
+        log.info(
```
josephjclark: This is a debug log at most

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +187,105 @@ export const loadTransformOptions = async (
...
+        const outPath = path.join(compiledDir!, workflow.id, `${step.id}.js`);
+        await fs.mkdir(path.dirname(outPath), { recursive: true });
+        await fs.writeFile(outPath, code);
+        outPaths.push(outPath);
+        log.success(`  ${workflow.id}/${step.id} → ${outPath}`);
+      }
+    }
+  }
+
+  // Only deletes exact step paths — user-added files alongside them are untouched.
```
josephjclark: I'm not sure what's going on here - we should only write paths that have content. I don't think we need to track `stalePaths` explicitly like this.

Is there more to deletes than I realise? I wouldn't do any tidy up: I'd just write compiled code to the output folder. Maybe take a `--clean` parameter to remove old files.

We're treating the compilation output as a `dist` folder, totally automated and unused by humans. Unit tests should not be inside this folder.

packages/cli/src/compile/handler.ts
```diff
@@ -19,12 +29,57 @@ const compileHandler = async (options: CompileOptions, logger: Logger) => {
...
+
+  const watcher = chokidar.watch(watchTargets, {
+    ignoreInitial: true,
+    ignored: ['**/node_modules/**', '**/compiled/**'],
+  });
+
+  watcher.on('change', async (changedPath: string) => {
+    logger.info(`${changedPath} changed, recompiling...`);
+    try {
+      if (options.workflowName) {
```
josephjclark: This looks like a duplication of the first half of the function?

packages/cli/src/compile/handler.ts
```diff
@@ -19,12 +29,57 @@ const compileHandler = async (options: CompileOptions, logger: Logger) => {
...
+    assertPath(options.path);
+    await doCompile(options, logger);
   }
+
+  if (!options.watch) return;
+
+  const watchTargets = collectWatchTargets(options);
+  logger.info(`Watching for changes. Ctrl+C to stop.`);
+
+  const watcher = chokidar.watch(watchTargets, {
```
josephjclark: nodejs has a `--watch` argument now. I wonder if you could just that, alongside with `--watch-path` in the scripts to get the same effect more easily

packages/compiler/src/transforms/exports-only.ts
```diff
@@ -0,0 +1,106 @@
...
+  const exportSeeds = body.filter((node) =>
+    n.ExportNamedDeclaration.check(node)
+  ) as n.Statement[];
+  const needed = collectDeps(exportSeeds, declMap);
+
+  programPath.node.body = body.filter(
+    (node) =>
+      n.ImportDeclaration.check(node) ||
+      n.ExportNamedDeclaration.check(node) ||
+      needed.has(node as n.Statement)
```
josephjclark: I don't understand this `needed` bit - what is it trying to do?

I think this should be really simple: accept import and name export statements, and discard the rest.

If we're trying to keep non exported constants and functions and dependencies and stuff, well that's out of scope and more work than I was expecting. Is there a specific use  case for that you're looking at?

packages/compiler/src/transforms/top-level-operations.ts
```diff
@@ -5,8 +5,6 @@
 import { namedTypes as n, namedTypes } from 'ast-types';
 import type { NodePath } from 'ast-types/lib/node-path';
 import type { Transformer } from '../transform';
-// Note that the validator should complain if it see anything other than export default []
```
josephjclark, line 8: Why are there diffs in this file? This should be unchanged now right?

packages/compiler/test/transforms/exports-only.test.ts
```diff
@@ -0,0 +1,287 @@
...
+    ),
+    []
+  );
+  const result = collectDeps([seed], declMap);
+  t.true(result.has(dep));
+});
+
+// --- exports-only transformer ---
+
+test('is a no-op when options is not true', (t) => {
```
josephjclark: I would prefer if these tests tests source code before and after

So like
```
const before = `export const x = () => {}; fn()'`

const after = `fn()`
```
So much easier to read than the ASTs, and since these test aren't particularly intricate I think it'll work better

mtuchi: I have updated unit tests to have the before and after version

packages/cli/src/compile/compile.ts
mtuchi: for checking if we have exports functions or variables

packages/cli/src/compile/compile.ts
mtuchi: I have clean up this a bit

packages/cli/src/compile/compile.ts
mtuchi: moved the logic to compiler

packages/cli/src/compile/handler.ts
mtuchi: Yes i cleaned it up

packages/compiler/src/transforms/top-level-operations.ts
mtuchi: made sure no diff in top-level-operations

packages/cli/src/compile/handler.ts
mtuchi: `node --watch` must be passed at process launch time, so it can't be toggled internally by `openfn compile --watch`. Should i drop the `--watch` flag from the CLI in favour of a Node-native approach ?.

mtuchi:
Unit test are failing, even on main. I am not sure what's causing it, i will investigate more tomorrow

josephjclark:
@mtuchi the tests failing here in CI are probably just flaky ones. I'll re-run them now. Locally it's probably most important that your compiler tests are passing. Sometimes my CLI tests fail locally because of caches stuff (they take so long to test I don't tend to run the whole suite locally - just the bit I'm working on)

josephjclark:
sorry @mtuchi I don't think I'm going to get to this this week. I've not forgotten, just couldn't get to it yet. It'll be high priority early next week.

packages/cli/src/compile/command.ts
```diff
@@ -27,40 +30,64 @@ export type CompileOptions = Pick<
...
   o.logJson,
-  override(o.outputStdout, {
-    default: true,
-  }),
+  o.outputStdout,
   o.outputPath,
   o.repoDir,
   o.trace,
   o.useAdaptorsMonorepo,
+  o.watchFlag,
```
josephjclark, line 50: We need to add the `workspace` command now: this sets the working directory for the current  project. `--workspace` is used by `openfn execute` and `openfn project`.

Very very useful for local testing!

packages/cli/src/compile/handler.ts
```diff
@@ -20,13 +29,50 @@ const compileHandler = async (options: CompileOptions, logger: Logger) => {
...
+    await mkdir(path.dirname(options.outputPath), { recursive: true });
+    await writeFile(options.outputPath, result);
+    logger.success(`Compiled to ${options.outputPath}`);
+  } else {
     logger.success('Result:\n\n' + result);
+  }
+};
+
+const runCompile = async (options: CompileOptions, logger: Logger) => {
+  if (options.workflowName) {
```
josephjclark: The logic here feels weird. `workflowName` feels too important, and `compileProject(options, logger)` is duplicated, which is a smell.

A cleaner entrypoint for this stuff is good. Basically if the path does not end with a file extension, we should `compileProject`

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +183,89 @@ export const loadTransformOptions = async (
...
+    process.exit(1);
+  }
+
+  const wsConfig = workspace.getConfig() as any;
+
+  const compiledDir = opts.outputStdout
+    ? null
+    : path.resolve(
+        cwd,
+        opts.outputPath ?? wsConfig.dirs?.compiled ?? 'compiled'
```
josephjclark: I would rather compile to `dist` than `compiled`. That's more likely to be tracked by a gitignore file too.

josephjclark: also `wsConfig.dirs.compiled` feels like a new key in workspace config. It should be defined in `lexicon/core.d.ts:WorkspaceConfig`

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +183,89 @@ export const loadTransformOptions = async (
...
+
+    if (opts.exportsOnly && !hasExportableCode(code)) {
+      log.debug(`  ${stepId} — skipped (no exportable code)`);
+      continue;
+    }
+
+    if (opts.outputStdout) {
+      log.success(`// ${stepId}\n\n` + code);
+    } else {
+      const outPath = path.join(compiledDir!, workflow.id, `${step.id}.js`);
```
josephjclark: It's always bothered me that `job.js` and `job.compiled.js` have the same file extension. For years I'd wondered if we should be using `job.ojs` for job expressions - but that ship has long since sailed.

As a compromise, can we output compiled files to `job.mjs`? That's more standard for modern Javascript esm modules, and it helps to distunguish a bit the file types

packages/cli/src/compile/command.ts
```diff
@@ -27,40 +30,64 @@ export type CompileOptions = Pick<
 const options = [
```
josephjclark, line 34: Can we add a `--clean` option which cleans the output folder before compiling? I think this is often going to be useful when doing `openfn compile`, but I don't think I want to default it to true

packages/cli/src/compile/compile.ts
```diff
@@ -179,3 +183,89 @@ export const loadTransformOptions = async (
...
+      await fs.mkdir(path.dirname(outPath), { recursive: true });
+      await fs.writeFile(outPath, code);
+      outPaths.push(outPath);
+      log.success(`  ${stepId} → ${outPath}`);
```
josephjclark: I think I would log this to `info`, not success, because it's very verbose

packages/cli/src/compile/handler.ts
```diff
@@ -1,16 +1,25 @@
...
-const compileHandler = async (options: CompileOptions, logger: Logger) => {
-  assertPath(options.path);
+const collectWatchTargets = (options: CompileOptions): string[] => {
+  if (options.expressionPath) {
+    return [path.resolve(options.expressionPath)];
+  }
+  if (options.path) {
+    return [path.resolve(options.path)];
+  }
+  return [path.join(process.cwd(), 'workflows', '**', '*.js')];
```
josephjclark: The `workflows` part of this path is config driven, so the watch would break for user setting a different path (which we generally allow).

Also `process.cwd()` doesnt' work if you're passing the `--workspace` argument (which we should support)

We should be able to feed in the correct path to the workflows folder from the associated Workspace.

I also wonder generally if `collectWatchTargets` should just be re-using the same logic as resolving the input paths.

packages/cli/src/compile/handler.ts
```diff
@@ -20,13 +29,50 @@ const compileHandler = async (options: CompileOptions, logger: Logger) => {
...
+    await writeFile(options.outputPath, result);
+    logger.success(`Compiled to ${options.outputPath}`);
+  } else {
     logger.success('Result:\n\n' + result);
+  }
+};
+
+const runCompile = async (options: CompileOptions, logger: Logger) => {
+  if (options.workflowName) {
+    await compileProject(options, logger, process.cwd(), options.workflowName);
```
josephjclark: Don't pass `process.cwd()` here. if you add the `workspace` option you should be able to just use `options.workspace` (which would default to process.cwd())

packages/cli/test/compile/compile.test.ts
```diff
@@ -342,3 +343,135 @@ test.serial('loadTransformOptions: ignore some imports', async (t) => {
...
+  'loadTransformOptions: --exports-only does not affect other options',
+  async (t) => {
+    const opts = { exportsOnly: true } as CompileOptions;
+    const result = await loadTransformOptions(opts, mockLog);
+    t.falsy(result['add-imports']);
+  }
+);
+
+test.serial(
+  'compileProject: compiles all workflow steps and writes output files',
```
josephjclark: These tests shouldn't be prefixed with `compileProject:` and appended to the end of the file. They should be like `compile from project` and be peers of the first `compile from source string` test.

packages/cli/test/compile/compile.test.ts
```diff
@@ -342,3 +343,135 @@ test.serial('loadTransformOptions: ignore some imports', async (t) => {
...
+  async (t) => {
+    const pnpm = path.resolve('../../node_modules/.pnpm');
```
josephjclark: This pnpm mock shoudn't be repeated across all tests like this. Create a mock wrapper at the top of the file which captures the boilerplate

packages/cli/test/compile/compile.test.ts
```diff
@@ -342,3 +343,135 @@ test.serial('loadTransformOptions: ignore some imports', async (t) => {
...
+    const outPaths = await compileProject(
+      {} as CompileOptions,
+      mockLog,
+      '/proj'
+    );
+
+    t.is(outPaths.length, 1);
+    t.true(outPaths[0].endsWith('compiled/wf1/step-a.js'));
+
+    const { default: nodeFsPromises } = await import('node:fs/promises');
```
josephjclark: This import is incredibly wierd. `fs` promises is imported at the top of the file already - no need for this stuff

packages/compiler/src/transforms/exports-only.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * Strips all non-exported top-level code (operations, export default, bare declarations).
+ */
+
+import { namedTypes as n } from 'ast-types';
+import type { NodePath } from 'ast-types/lib/node-path';
+import type { Transformer } from '../transform';
+
+// Returns true if compiled output contains any declarations worth importing in tests.
+export const hasExportableCode = (code: string): boolean =>
```
josephjclark: This would break with `export { x, y, z }` no?

Not a fan of this. What purpose does it solve? Feels like we're over-optimising

packages/compiler/test/transforms/exports-only.test.ts
```diff
@@ -0,0 +1,142 @@
+import test from 'ava';
+import { print } from 'recast';
+import parse from '../../src/parse';
+import transform from '../../src/transform';
+import visitors, { hasExportableCode } from '../../src/transforms/exports-only';
+
+const compile = (source: string, options = {}) =>
+  print(transform(parse(source), [visitors], options)).code.trim();
+
+// --- hasExportableCode ---
```
josephjclark: This comment has no value. AI is terrible fof this sort of thing - please try and be strict on it. Only keep stuff with value

packages/compiler/test/transforms/exports-only.test.ts
```diff
@@ -0,0 +1,142 @@
...
+});
+
+test('hasExportableCode: false for operations only', (t) => {
+  t.false(hasExportableCode('fn();\nget();'));
+});
+
+test('is a no-op when options is not true', (t) => {
+  const before = `fn();
+export default [];`;
+  t.is(compile(before), before);
```
josephjclark: I don't like logic inside an assertion. A better pattern is:
```
const before = `fn();
export default [];`;
const after = compile(before)
  t.is(after, before);
```

packages/compiler/test/transforms/exports-only.test.ts
```diff
@@ -0,0 +1,142 @@
...
+
+test('hasExportableCode: false for empty string', (t) => {
+  t.false(hasExportableCode(''));
+});
+
+test('hasExportableCode: false for operations only', (t) => {
+  t.false(hasExportableCode('fn();\nget();'));
+});
+
+test('is a no-op when options is not true', (t) => {
```
josephjclark: In terms of storytelling through tests, rather than testing the inverse case first, I'd rather test basic functionality:

```
strips a non exported value
is a no-op if exports-only is false
```

Where the "before" code in both tests is exactly the same

packages/compiler/test/transforms/exports-only.test.ts
```diff
@@ -0,0 +1,142 @@
...
+  const before = `fn();
+export default [];`;
+  t.is(compile(before), before);
+});
+
+test('is a no-op when options is false', (t) => {
+  const before = `fn();`;
+  t.is(compile(before, { 'exports-only': false }), before);
```
josephjclark: Actually come to think of it, I don't like the use of `compile` here

The other transformer functions just test the transform functoin with visitors passed in.

This calls the top level `compile` function, so it's testing all sorts of logic before even getting to the transformer.

Ask the AI to rewrite these tests against the `transform` function, not the `compile` function. use the `parse` function with source code code, like in `lazy-state.test.ts`, rather than the the AST builders in `add-imports.test.ts`

josephjclark (changes requested):
Hi @mtuchi, I'm so sorry for the slow reaction on this.

I've been through and tested and it all works great. I've picked out a few things that need addressing - mostly style refactors but also a couple of important details.

If you don't have time to resolve this I may be able to pick it up at the end of the week.

packages/cli/src/compile/compile.ts
mtuchi: done 👍🏽

packages/compiler/test/transforms/exports-only.test.ts
mtuchi: thanks for the reminder joe, cleaned up

packages/compiler/test/transforms/exports-only.test.ts
mtuchi: I have update the test to use `transform` function

packages/cli/src/compile/handler.ts
mtuchi: update my implementation approach by checkign if it's file input first

packages/compiler/src/transforms/exports-only.ts
mtuchi: dropped `hasExportableCode` function

packages/cli/src/compile/command.ts
mtuchi: I have include `po.workspace` option

packages/compiler/src/transform.ts
```diff
@@ -60,6 +63,7 @@ export default function transform(
 
   if (!transformers) {
     transformers = [
+      exportsOnly,
```
josephjclark: I've just seen this and I'm super woirred about it. It's _critical_ that exportsOnly is off by default.

Tests are passing on this branch so I think I'm just being paranoid: but I want to look at this really closely before I merge anything.

Which won;t be today.

josephjclark: Ok: so this paniced me because if this transformer runs, all jobs break. Which obviously we're not seeing in unit tests.

Unlike every other transformer, this one will only run in the visitor if `options.exportsOnly`is truthy.

I would much rather that this transfomer was added to the list under certain conditions - rather than always active but sometimes disabled.

But I wonder what the engineering cost of that is...

PR #1444, opened 2026-06-05
fix(cli): fix cache .gitignore written to fs root and undefined workflow name

### LINKED ISSUE #669: CLI cache bug
### User story

I tried running this job `cursor.js` 👇🏽 
```
cursor($.manualCursor, { defaultValue: 'now' });
```
I am getting this error 👇🏽 
```
openfn jobs/uk/tmp/cursor.js -ma common -o jobs/uk/tmp/cursor.json
[CLI] ♦ Versions:
         ▸ cli                        1.2.0
         ▸ @openfn/language-common    monorepo
[CLI] ✔ Loading adaptors from monorepo at /Users/openfn/Workspace/adaptors
[CLI] ⚠ Skipping auto-install as monorepo is being used
Setting cursor "now" to: 11:04 17 Apr 2024 (GMT+3)
[R/T] ✔ Completed step job-1 in 262ms
[CLI] ✔ Cached output written to ./cli-cache (see info logs for details)
node:internal/errors:484
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
    at new NodeError (node:internal/errors:393:5)
    at validateString (node:internal/validators:163:11)
    at Object.resolve (node:path:1098:7)
    at ensureGitIgnore (file:///Users/openfn/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/dist/process/runner.js:49:29)
    at saveToCache (file:///Users/openfn/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/dist/process/runner.js:66:5)
    at async notify (file:///Users/openfn/.asdf/installs/nodejs/18.12.1/lib/node_modules/@openfn/cli/dist/process/runner.js:100:13) {
  code: 'ERR_INVALID_ARG_TYPE'
}

Node.js v18.12.1
```

### Additional Context 
I have setup the always cache in my `.zshrc` file  so that i don't have to specify `--cache-steps` every-time 
```
export OPENFN_ALWAYS_CACHE_STEPS=true
```

lamine2000:
## Summary

Fixes the CLI cache bug reported in #669 where running a `.js` expression with caching enabled (via `OPENFN_ALWAYS_CACHE_STEPS=true` or `--cache-steps`) caused a `TypeError [ERR_INVALID_ARG_TYPE]` crash.

- **`ensureGitIgnore` rewritten**: The old implementation walked up the cache file path looking for a segment ending in `.cli-cache`. If the path did not contain that segment (e.g. a custom `cachePath`), the loop traversed all the way to the filesystem root and wrote `/.gitignore`. Replaced with a direct call to `getCachePath(options)` (no `workflowName`/`stepId`), which naturally returns the cache root, so no traversal is needed.
- **Undefined workflow name crash fixed**: `path.resolve` throws `ERR_INVALID_ARG_TYPE` when passed `undefined`. `getCachePath` now uses `.filter(Boolean)` before spreading into `path.resolve`, so an absent `plan.workflow.name` is simply omitted from the path rather than crashing.
- **Eliminated `getCacheRoot` helper**: After the above two fixes, `getCachePath(options)` with no extra arguments already returns the cache root directly — no separate helper needed.
- **Type fix for `saveToCache` and `clearCache`**: Both were typed as `Pick<Opts, 'baseDir' | 'cacheSteps'>`, missing `cachePath`, so the `cachePath` branch inside `getCachePath` was unreachable from these functions. Added `cachePath` to both Pick types.

## Test plan

- [x] Added `cache steps when running a .js expression` — runs an expression with `cacheSteps: true` and verifies the cache file is written under `.cli-cache/<name>/` (this is the exact scenario from #669, which had zero test coverage)
- [x] Added `.cli-cache has a gitignore when caching a .js expression` — verifies `.gitignore` is correctly placed in `.cli-cache/` and not at the filesystem root
- [x] All existing cache tests continue to pass

Closes #669

### DIFF
--- a/packages/cli/src/util/cache.ts
+++ b/packages/cli/src/util/cache.ts
@@ -23,12 +23,10 @@ export const getCachePath = (
     return path.resolve(cachePath);
   }
 
-  // path.resolve throws on undefined, so filter it out; when workflowName
-  // is absent the path resolves to the bare CACHE_DIR root.
   const basePath = path.resolve(
-    ...([baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
-      Boolean
-    ) as string[])
+    baseDir ?? process.cwd(),
+    CACHE_DIR,
+    workflowName ?? ''
   );
 
   if (stepId) {
--- a/packages/cli/src/util/cache.ts
+++ b/packages/cli/src/util/cache.ts
@@ -13,7 +13,7 @@ export const CACHE_DIR = '.cli-cache';
 export const getCachePath = (
   options: Pick<Opts, 'baseDir' | 'cachePath'>,
   workflowName?: string,
-  stepId?: string,
+  stepId?: string
 ) => {
   const { baseDir, cachePath } = options;
   if (cachePath) {
@@ -26,9 +26,9 @@ export const getCachePath = (
   // path.resolve throws on undefined, so filter it out; when workflowName
   // is absent the path resolves to the bare CACHE_DIR root.
   const basePath = path.resolve(
-    ...([baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
-      Boolean,
-    ) as string[]),
+    ...[baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
+      Boolean
+    ) as string[]
   );
 
   if (stepId) {
@@ -55,7 +55,7 @@ export const saveToCache = async (
   stepId: string,
   output: any,
   options: Pick<Opts, 'baseDir' | 'cachePath' | 'cacheSteps'>,
-  logger: Logger,
+  logger: Logger
 ) => {
   if (options.cacheSteps) {
     const cachePath = getCachePath(options, plan.workflow.name, stepId);
@@ -73,7 +73,7 @@ export const saveToCache = async (
 export const clearCache = async (
   plan: ExecutionPlan,
   options: Pick<Opts, 'baseDir' | 'cachePath'>,
-  logger: Logger,
+  logger: Logger
 ) => {
   const cacheDir = getCachePath(options, plan.workflow?.name);
 
--- a/packages/cli/src/util/cache.ts
+++ b/packages/cli/src/util/cache.ts
@@ -26,9 +26,9 @@ export const getCachePath = (
   // path.resolve throws on undefined, so filter it out; when workflowName
   // is absent the path resolves to the bare CACHE_DIR root.
   const basePath = path.resolve(
-    ...[baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
+    ...([baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
       Boolean
-    ) as string[]
+    ) as string[])
   );
 
   if (stepId) {
--- a/packages/cli/src/util/cache.ts
+++ b/packages/cli/src/util/cache.ts
@@ -8,7 +8,6 @@ import type { Logger } from './logger';
 
 export const CACHE_DIR = '.cli-cache';
 
-// TODO this is all a bit over complicated tbh
 export const getCachePath = (
   options: Pick<Opts, 'baseDir' | 'cachePath'>,
   workflowName?: string,
@@ -24,7 +23,7 @@ export const getCachePath = (
 
   const basePath = path.resolve(
     baseDir ?? process.cwd(),
-    `${CACHE_DIR}/${workflowName}`
+    `${CACHE_DIR}/${workflowName ?? 'workflow'}`
   );
 
   if (stepId) {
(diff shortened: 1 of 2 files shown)


packages/cli/src/util/cache.ts
```diff
@@ -24,7 +23,7 @@ export const getCachePath = (
 
   const basePath = path.resolve(
     baseDir ?? process.cwd(),
-    `${CACHE_DIR}/${workflowName}`
+    `${CACHE_DIR}/${workflowName ?? 'workflow'}`
```
josephjclark: Not sure how often this problem occurs but I don't want to default to `workflow` here

A better solution would be 
```
 const basePath = path.resolve(
    baseDir ?? process.cwd(),
    CACHE_DIR,
    workflowName,
  );
```

Where the undefined workflow name will simply be ignored by path.resolve (I think!)

packages/cli/src/util/cache.ts
```diff
@@ -33,38 +32,41 @@ export const getCachePath = (
   return basePath;
 };
 
-const ensureGitIgnore = (options: any, cachePath: string) => {
+// Returns the root cache directory where .gitignore should live
+const getCacheRoot = (options: Pick<Opts, 'baseDir' | 'cachePath'>): string => {
```
josephjclark: What is the different between `getCachePath` and `getCacheRoot`?

packages/cli/src/util/cache.ts
```diff
@@ -33,38 +32,41 @@ export const getCachePath = (
...
+  const { baseDir, cachePath } = options;
+  if (cachePath) {
+    return path.resolve(cachePath);
+  }
+  return path.resolve(baseDir ?? process.cwd(), CACHE_DIR);
+};
+
+const ensureGitIgnore = (options: any, cacheRoot: string) => {
```
josephjclark, line 38: The old logic here, the reason we walk up through the folder structure, is to see if there's a parenting .gitignore file which we can edit. This would live at the root of the github repo, which we may or may not find.

This PR changes the approach to force a .gitignore file as a child of the cache folder. 

I'm not against this. It might be a bit messy but I suppose it's not too big a deal. And I can see the original issue of writing a `.gitignore` file to root (although I can't reproduce this locally but that might be something in my local setup)

So Ok, let's do it!

packages/cli/test/execute/execute.test.ts
```diff
@@ -418,6 +418,53 @@ test.serial('.cli-cache has a gitignore', async (t) => {
   t.is(gitignore, '*');
 });
 
+// Regression test for https://github.com/OpenFn/kit/issues/669
```
josephjclark, line 421: Why is this test making such a distinction between workflow.js and expression.js? I don't think  it materially affects the cache behaviour?

packages/cli/test/execute/execute.test.ts
```diff
@@ -418,6 +418,53 @@ test.serial('.cli-cache has a gitignore', async (t) => {
...
+// Running a .js expression (not a workflow JSON) with caching enabled must not crash.
+test.serial('cache steps when running a .js expression', async (t) => {
```
josephjclark, line 422: This test just ensures that the cache is written, not read. Which is fine - I'd just like the test name to be accurate.

josephjclark (changes requested):
Hi @lamine2000, thank you for raising this. I think there's just one  necessary change but there are a few discussion points I'd like addressed.

I am not able to reproduce Mtuchi's original issue. It is quite old. Were you able to reproduce it with the CLI?

I appreciate the type fix on `saveToCache`. It is correct, but the PR comments are not. This untrue: `This meant custom cache paths set by workspace projects were silently ignored when calling getCachePath from these functions`.  The option is carried through regardless of the typings.

Please check that AI generated PR descriptions are accurate! Misleading descriptions take more time to check

lamine2000 (commented):
Not with the CLI directly, no. But I traced both failure paths in the code and confirmed them.

For the \`path.resolve\` crash, this reproduces it:

\`\`\`
node -e "const path = require('path'); path.resolve('/base', '.cli-cache', undefined)"
# TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received undefined
\`\`\`

\`plan.workflow.name\` is undefined when running a bare \`.js\` file, so that's exactly what was happening.

For the \`.gitignore\` at \`/\`, I confirmed it by reading the code. The old loop walked up the path one segment at a time looking for something ending in \`.cli-cache\`. If the cache path doesn't contain that string, it reaches root and writes there.

The two new tests cover both cases and fail against the original code, which felt like the most solid confirmation I could give.

On the description — you're right, I corrected it. The type fix doesn't affect runtime behaviour at all, the option was always passed through. I've updated the description to reflect that accurately.

And noted on AI-generated descriptions — I should have caught that before opening the PR. I'll make sure to verify every claim manually going forward.

packages/cli/src/util/cache.ts
josephjclark: Ah I see. Following your comment and checking the node docs, this doesn't actually work. A non string causes the throw. one comment coming in on the refactor

packages/cli/src/util/cache.ts
```diff
@@ -22,9 +23,12 @@ export const getCachePath = (
     return path.resolve(cachePath);
   }
 
+  // path.resolve throws on undefined, so filter it out; when workflowName
+  // is absent the path resolves to the bare CACHE_DIR root.
   const basePath = path.resolve(
-    baseDir ?? process.cwd(),
-    `${CACHE_DIR}/${workflowName}`
+    ...[baseDir ?? process.cwd(), CACHE_DIR, workflowName].filter(
```
josephjclark: This is still quite ungainly, not really an improvement.

Since the path is finite and short, we can do this instead:
```
 const basePath = path.resolve(
    baseDir ?? process.cwd(),
    CACHE_DIR,
    workflowName ?? '',
  );
```

lamine2000: great catch ! I applied that too. Thanks

josephjclark (approved):
Thanks @lamine2000! I'll merge and run some local tests before releasing

PR #1445, opened 2026-06-05
fix(cli): create output directory if it doesn't exist

### LINKED ISSUE #859: CLI: don't throw if output path doesn't exist
Currently if you run the CLI and ask for output in a path that doesn't exist (ie, a folder hasn't been created), the CLI will throw after execution.

This isn't very good - we should `rm.mkrDir(path.dirname(opts.outputPath), { recursive: true })` to ensure the output is created safely.

lamine2000:
## Summary

Fixes #859. The `compile` and `collections get` commands threw `ENOENT` when `--output-path` pointed to a directory that hadn't been created yet, forcing users to manually `mkdir` before running the CLI.

- **`compile/handler.ts`**: added `await mkdir(dirname(outputPath), { recursive: true })` before `writeFile`
- **`collections/handler.ts`**: same fix in the `get` command's output branch

The `execute` command's `serialize-output.ts` already had this pattern — this PR brings the other two commands in line with it.

## Test plan

- [ ] `openfn compile job.js -o new/nested/dir/out.js` — should create `new/nested/dir/` and write the file
- [ ] `openfn collections get myCollection myKey -o new/nested/dir/out.json` — same
- [ ] Both commands still work normally when the output directory already exists

Closes #859

### DIFF
--- a/packages/cli/src/collections/handler.ts
+++ b/packages/cli/src/collections/handler.ts
@@ -1,5 +1,5 @@
 import path from 'node:path';
-import { readFile, writeFile } from 'node:fs/promises';
+import { mkdir, readFile, writeFile } from 'node:fs/promises';
 
 import { Logger } from '../util/logger';
 import request from './request';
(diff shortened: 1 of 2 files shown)


josephjclark:
Thanks @lamine2000 . There's a failing test here that needs resolving. Also this branch includes your other work. Please rebase against main so that this PR is nice and clean. Thank you!

josephjclark (approved):
Thanks @lamine2000 !

PR #1446, opened 2026-06-08
Fix incorrect divergence warnings on checkout

### LINKED ISSUE #1433: merge: investigate surprising divergence warnings
When testing merging locally last week, I experienced surprising divergence warnings on merge and checkout.

Is this because version hashes are out of date? Is the divergence warning just wrong (surely not?). Am I just losing the plot?

josephjclark:
## Short Description

The checkout code  is using the wrong algorithm to detect whether there are "unstaged" changes that could be lost on checkout. I think this was probably broken when updating the algorithm for other reasons, which subtly broke checkout.

Checkout didn't have good divergence tests so the changes went unnoticed.

This PR introduces bespoke algorithm which does the job a lot better.

Fixes #1433

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/hungry-lights-love.md
+++ b/.changeset/hungry-lights-love.md
@@ -1,0 +1,5 @@
+---
+'@openfn/lexicon': patch
+---
+
+Allow project.alias to be null
--- a/.changeset/late-needles-scream.md
+++ b/.changeset/late-needles-scream.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Fix an issue on checkout where incorrect divergence warnings can be shown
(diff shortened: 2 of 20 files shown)


josephjclark:
While testing here I keep hitting issues because the workflow is out of date somehow with the state file, and so the versions are different (because the state is different). Why is this?

Is it because I'm just doing things that don't make sense locally? Did I forget to pull?

Or is it something deeper. Like the checked out workflow is simply behind the statefile? Oh that could well be something:
- I pull the project
- I change remotely
- I pull again - local conflict right, because local and remote are different (even though remote did not change)

So I need to be comparing the local history to the app's history. Bit of a wrinkle there is that the app might have squashed versions - the client kinda needs to maintain versions. This is getting hairy.

Well, one thing about that is that it's live in production now. Not new to this branch. I think I'm at least solving some things here.

PR #1453, opened 2026-06-09
Fix gitignore cache

josephjclark:
merge branch for https://github.com/OpenFn/kit/pull/1444


josephjclark:
made a tweak to log output while testing because some of the logging isn't actually super clear.

But the fix looks great. The gitignores themselves get ignored so creating new ones doesn't create lots of clutter

Issue #1456, opened 2026-06-10
Sync: ensure workflows can be removed


josephjclark:
- In v1 deploy workflows cannot be removed because of the error generated here: https://github.com/OpenFn/kit/blob/main/packages/deploy/src/stateTransform.ts#L464
- Not sure how this works in v2 deploy

The provisioner will remove workflows, steps and edges if delete: true is passed. It's the CLI's job to detect the removal and set this flag.

Ensure it works in v1 and v2 and is well tested.

josephjclark:
State of v2 is:
- removal is handled in merge
- removed items are just removed, they;re not flagged with delete (we need an option for this behaviour)
- If I remove a step in v2 it seems deploy will just throw an error  `Edge references unknown target "undefined". seems a node's id changed without edges connecting to it being updated.`
- If I remove a workflow in v2 I get a different error `sourceWorkflow.canMergeInto is not a function`

So basically this is completely untested and supported in v2 :( Feels like a pretty big oversight

PR #1457, opened 2026-06-10
Allow deploy v1 to remove workflows

josephjclark:
## Short Description

Remove error on mismatching workflows and replace with a delete operation

Partial fix for #1456  (but doesn't address v2)

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Is there a better  way to tell the difference between "the spec and state don't match" and "the spec removed a workflow"?

Want to think about this some more

PR #1458, opened 2026-06-18
feat: ensure max state memory size

### LINKED ISSUE #1459: Allow setting max state memory via environment variable
`stateLimitMb` which controls the maximum size of state being passed between nodes in the runtime is currently not configurable via a worker env or lightning configuration.

doc-han:
## Short Description

This PR completes the plumbing required to get state limit set for a worker.

#### 2 ways to go about it
1. `state_limit_mb` from Lightning
2. `--state-memory` / `WORKER_MAX_STATE_MEMORY_MB`

Fixes #1459 

## An update from Joe

Turns out that there's already a reasonable limit set by the engine.

So the original PR just exposes options to properly control it.

I've added a couple of things that I think are more valuable:
1. Logging the size of state objects after serialization
2. Reduce the default a little bit to try and prevent blowups and lost runs

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/packages/ws-worker/src/util/cli.ts
+++ b/packages/ws-worker/src/util/cli.ts
@@ -38,6 +38,7 @@ type Args = {
   profilePollIntervalMs?: number;
   repoDir?: string;
   runMemory?: number;
+  stateMemory?: number;
   secret?: string;
   sentryDsn?: string;
   sentryEnv?: string;
@@ -91,6 +92,7 @@ export default function parseArgs(argv: string[]): Args {
     WORKER_MAX_LOG_PAYLOAD_MB,
     WORKER_MAX_RUN_DURATION_SECONDS,
     WORKER_MAX_RUN_MEMORY_MB,
+    WORKER_MAX_STATE_MEMORY_MB,
     WORKER_MESSAGE_TIMEOUT_SECONDS,
     WORKER_PORT,
     WORKER_PROFILE_POLL_INTERVAL_MS,
@@ -219,6 +221,11 @@ export default function parseArgs(argv: string[]): Args {
         'Maximum memory allocated to a single run, in mb. Env: WORKER_MAX_RUN_MEMORY_MB',
       type: 'number',
     })
+    .option('state-memory', {
+      description:
+        'Maximum size of the state object returned by each step, in mb. Defaults to 25% of run-memory. Env: WORKER_MAX_STATE_MEMORY_MB',
+      type: 'number',
+    })
     .option('payload-memory', {
       description:
         'Maximum memory allocated to a single run, in mb. Env: WORKER_MAX_PAYLOAD_MB',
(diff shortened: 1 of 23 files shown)


josephjclark:
@doc-han this looks great but I think we can add a better test to make sure the limit is set and working. Probably in integration tests? Just set the limit quite low and build a state object big enough to trigger it. Search for Array.fill for patterns of this sort of thing

doc-han:
There was a flaky test on dataclip loading which I've resolved on this PR

packages/ws-worker/test/api/execute.test.ts
```diff
@@ -301,8 +301,12 @@ test('loadDataclip report to sentry on fail', async (t) => {
   } catch (e) {}
 
   const reports = await waitForSentryReport(testkit);
-  t.is(reports.length, 1);
-  t.is(reports[0].error.name, 'LightningSocketError');
+  const dataclip_report = reports.find((r: any) =>
+    /fetch\:dataclip.+not_found/i.test(r.error.message)
+  );
+  console.log(dataclip_report);
```
josephjclark: fantastic, thank you for the fix! Can you remove this log line for me?

josephjclark:
@doc-han just a changeset and that log line needed and we can release this. Thank you, nice work!

packages/ws-worker/test/api/execute.test.ts
doc-han: removed

doc-han:
Ready!

josephjclark (approved):

packages/ws-worker/src/util/cli.ts
```diff
@@ -219,6 +221,11 @@ export default function parseArgs(argv: string[]): Args {
         'Maximum memory allocated to a single run, in mb. Env: WORKER_MAX_RUN_MEMORY_MB',
       type: 'number',
     })
+    .option('state-memory', {
+      description:
+        'Maximum size of the state object returned by each step, in mb. Defaults to 25% of run-memory. Env: WORKER_MAX_STATE_MEMORY_MB',
```
josephjclark, line 226: I want to think about this before merging

I'd missed/forgotten that we're doing this already. Which means the limit in prod right now isn't 500mb (as is the default in the function), it's more like 250mb. And that might be big enough to cause problems still.

Should we reduce the default to 10% of working memory? Is that too strict and small?

Also should we be adding some logging around this? For users or GCP?

I do see a `logger.debug` line which prints the state size limit but I can't see that being logged in production - what's that all about?

josephjclark (changes requested):
Picked up a couple of questions in a final review pass before merging. I need to think about these before going ahead.

josephjclark:
I've run some tests, added some logging and reduced the default max size. That might help a bit.

I'm just looking at one more thing that might help process large state objects

josephjclark (approved):

PR #1462, opened 2026-06-23
Use prlimit

josephjclark:
## Implementation Details

Spun out of  https://github.com/OpenFn/kit/pull/1370

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Fantastic. Wasted a whole day on getting this ready and it turns out it's just not going to work.

This won't help us. The address space limit we set has to be siginificantly larger than the space required by a run. I'll post more detai in #1370.

tl;dr: If I set AS to 1gb using rlimit, thenthe node process won't even run. We need to let it allocate way more memory (the original PR uses 10x the limit). That means kubernetes will kill the pod way before the rlimit kicks in.

PR #1465, opened 2026-06-25
CLI: Deploy through v1 API with v2 spec

### LINKED ISSUE #1464: Enable deploy v1 to deploy a v2 spec file
We're changing the exported project.yaml file which lighting generates to use the new v2 spec format.

This means that anyone using CLI deploy v1 and exported project.yamls will find their deploys start breaking. This affects eg ethiopia but maybe others.

To preserve back compatibility, it would be neat for the v1 deploy to:
1) detect if an incoming yaml is in v2 format
2) convert it to v1 format (project has APIs for this)
3) carry on as normal

No need to itervene in v1 pull.

josephjclark:
## Short Description

When using the v1 CLI deploy commands, add support for a v2 project file by converting it into a v1 spec file.

Fixes #1464

This includes a rather janky fix - because I thought a v1 spec file was just state minus UUIDS. But there's a bit more to it than that, so the code which converts a Project to v1 state need smore logic to generate v1 spec instead. Not nice. But I hope this is a temporary fix and it won't be around for long

## QA Notes

I have run the following manual tests

* [x] Can deploy a v1 spec to lightning on main
* [x] Can deploy a v1 spec to lightning on portability branch
* [x] Can deploy a v2 spec to lightning on portability branch
* [x] Handles credentials

All for new projects and with the v1 command

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/forty-areas-remain.md
+++ b/.changeset/forty-areas-remain.md
@@ -1,0 +1,5 @@
+---
+'@openfn/lightning-mock': patch
+---
+
+Add validation to the provisioner endpoint
--- a/.claude/yaml-formats.md
+++ b/.claude/yaml-formats.md
@@ -82,10 +82,11 @@ workflows:
 
 ## Detection logic
 
-Mirrors `packages/project/src/parse/from-project.ts`:
+Use `detectVersion(data)` from `@openfn/project` — returns `1` or `2`. Accepts YAML/JSON string or pre-parsed object.
 
 ```typescript
-const isV2 = json.schema_version || json.cli?.version === 2 || json.version;
+import { detectVersion } from '@openfn/project';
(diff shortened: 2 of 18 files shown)


brandonjackson (commented):
🎉

josephjclark:
Ok, hit a problem which will need to be properly reflected in unit and integration tests. Because these should not be passing.

When converting the v2 project down to a v1 spec, we MUST exclude UUIDs. The Project converter will add them somewhere down the line but this invalidates assumptions in the v1 sync.

This should have been caught in a integration test: I'm going to add some validation to the mock API. So the plan is:
1. Update lightning mock
2. Integration test should fail
3. Export a v2 project without UUIDs
4. Add the no-uuid flag to the CLI

PR #1466, opened 2026-06-28
portability test: state to spec

josephjclark:
Using claude to look at converting a Project to a v1 spec file.

This is actualy more involved than I thought. I thought it was just stripping the uuids but actually there are some different key names too.

Should I stick with this approach or should we have a `serialize('spec', opts)` function? Maybe it's not worth it?

Do we need a v2 spec serializer?


josephjclark:
this does seem to work so maybe it's good enough :thinking:

josephjclark:
Issue: credentials are not converted

josephjclark:
Think I've pushed a credentials fix. It's really ugly because of subtle differences between spec and state. Probably I should create a `to-spec` serializer but the differences aren't quite pronounced enough that I want  to do it. Also this is hopefully not code that's going to be exercised very much, so it an afford to be ugly.

PR #1477, opened 2026-07-05
Add node 18 legacy test

josephjclark:
Restore integration tests for node 18

this is needed because the GH sync action relies on node 18, and sometimes I forget

Maybe we don't need the whole suite though. Just needed for legacy deploy.


josephjclark:
Ok, part of the problem is that lightning mock won't work in node18

I'm going to just do a basic HTTP mock for the provisioner API call instead

PR #1481, opened 2026-07-15
CLI: Deploy straight from yaml

### LINKED ISSUE #1264: Deploy: allow deploy straight from project.yaml
Speak to frank about this

Frank says it's easier to use v1 to deploy new projects because he can deploy a single file.

But we should be able to do this with v2 by not checking out the project. Fetch and then deploy with --new should do it? We may need a flag to skip the working dir and deploy straight from a file.

josephjclark:
## Short Description

This PR enables users to run a deploy from a file: either an exported project.yaml (from the app) or from A single v2 state file (the thing created when fetching a project).

```
openfn project deploy .projects/main@app.openfn.org.yaml --name my-new-project
```

Note that if you try and deploy a project to the same instance and the same UUID, it'll just override. In this case you have to pass `--new` or set a new name.

A side-effect of this work is that you can now edit a local project.yaml  file and deploy it directly to lightning - previously the CLI didn't allow this because a deploy would re-build the project.yaml from the local filesystem. Being able to modify the yaml directly is probably a useful escape hatch. But one I'll leave undocumented for now.

Fixes #1264

## QA Notes

To test this ( I recommend running against local or at least staging)

You can set `OPENFN_ENDPOINT` and `OPENFN_API_KEY` env vars to configure deploy targets and access and stuff. `.env` files work.

### Scenario 1 - deploy an exported project

- Go to any project in the app and export it as yaml
- copy the path to the file
- With a superuser API token, deploy that file to an instance: `openfn project deploy downlaoded-project.yaml --name my-duplicate`
- The project should exist on the target server
- Should work with whatever name you pass, or no name
- Should work on the same instance as you exported from, or a different one

### Scenario 2 - re-deploy an existing project

- Fetch a project: `openfn project pull <uuid> --alias og`
- Note that you'll now have a file like  `og@app.openfn.org.yaml`
- Now, with a superuser API token, deploy that file back: `openfn project deploy ./projects/og@app.openfn.org.yaml --name my-duplicate`
- Should work with whatever name you pass, or no name
- Should work on the same instance as you exported from, or a different one

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have not used AI


josephjclark:
Right something I need to think about here

if you post a yaml directly, how do we know it's a new project and you don't just want to force push an update?

Presumanly you should be able to manuaally edit the yaml file and force push it from the file. A regular deploy will update from the file system but why not just edit the yaml?

You can pass --new to force it but eh it should be obvious right? The project is new if either
- You are posting with a different name
- You are posting to a different domain

So let's add a little automation around that

josephjclark:
@midigofrank just a headsup that I'm close to getting this closed out - expecting to release tomorrow

josephjclark:
Rebuilt as https://github.com/OpenFn/kit/pull/1533 because the base got messed up (idk even what's happened here)

PR #1485, opened 2026-07-16
Release worker 1.29.0: cgroups memory enforcement

josephjclark:
# Enforce run memory limits with cgroups (opt-in)

Adds cgroup-enforced memory limit for each child process in the engine.

This means the kernel can kill an over-budget run directly instead of the memory pressure spilling over and taking down the whole worker container.

## Background

A cgroup ("control group") is a Linux kernel feature that puts a hard ceiling on the resources a process (or group of processes) can use. Unlike our existing memory limit — which sets Node/V8's `--max-old-space-size` and only bounds the JS heap — a cgroup limit is enforced by the kernel and covers _all_ memory the process touches, including native allocations and buffers.
https://github.com/OpenFn/lightning/pull/5043
Today, a run that allocates heavily outside the JS heap can slip past our heap limit, balloon past what it's meant to be allowed, and get killed by the surrounding container platform (e.g. Kubernetes OOM-killing the whole worker pod) instead of just that one run. cgroup enforcement fixes this by having the kernel kill only the offending run's process.

## Implementation details

- Enforcement lives in `engine-multi`'s worker pool, keyed off a new `memoryEnforcement: { oldspace?, cgroup? }` option (`oldspace` is today's heap limit, on by default; `cgroup` is the new one, off by default). Both can run at once.
- On startup, the pool resolves the cgroup it's currently running in (`/proc/self/cgroup`) and uses that as the parent for per-child leaves — the engine never provisions the cgroup hierarchy itself, only leaves underneath whatever it's handed.
- Each pooled child process gets its own leaf cgroup, with `memory.max` set to the run's memory limit plus a fixed overhead. The child's pid is moved into the leaf once the limit is set.
- When a worker exits unexpectedly, the pool checks the leaf's `memory.events` for an OOM kill before falling back to the existing heuristic (scraping stderr for V8's "heap out of memory" message). This distinguishes a kernel-level cgroup kill from a heap-level one.
- The distinction is surfaced end-to-end as `OOMError.source` (`'heap' | 'cgroup'`), threaded through IPC and the engine's error events so callers (e.g. `ws-worker`) can tell which limit was breached.
- `ws-worker` exposes the `cgroup` option via `--cgroups` / `WORKER_ENABLE_CGROUP_ENFORCEMENT`.

## Basic docs

Cgroups are off by default because they require support from the environment.

Enable with either:

```bash
openfn worker --cgroups
```

or

```bash
WORKER_ENABLE_CGROUP_ENFORCEMENT=true
```

## Container environment considerations

The worker doesn't create its own cgroup hierarchy — it enforces limits using whatever cgroup it's already running in, so the platform has to hand it something usable:

- **Write access.** A plain `docker run` or default Kubernetes pod won't grant this out of the box; it needs configuring in your container runtime/orchestration setup (e.g. `systemd-run ... -p Delegate=yes` locally). If it's missing entirely, enforcement silently falls back to heap-only — safe, just without the extra protection.
- **Exclusivity.** Write access alone isn't enough — the cgroup also needs to be exclusive to the worker. If it's shared with unrelated processes (e.g. a bare dev terminal, where it's shared with your shell, other tools, sometimes your whole desktop session), an OOM kill can take out more than the offending run. See `packages/ws-worker/README.md` for the local dev workaround.
- For local Docker testing, write access can be granted with `--privileged`, but **this isn't appropriate for production** — it grants far more than cgroup access (full device access, all Linux capabilities, disabled seccomp/AppArmor). Production deployments should work out the narrowest cgroup-write grant with whoever owns the container/orchestration setup.

## Docs

See `packages/engine-multi/README.md` and `packages/ws-worker/README.md` for full details.


packages/engine-multi/src/worker/cgroup.ts
```diff
@@ -0,0 +1,287 @@
...
+ * systemd unit with Delegate=, by a container runtime, or by a launcher script
+ * (see docs/cgroup-spec.md). When the contract isn't met, every function here
+ * degrades to a no-op and the caller falls back to heap-limit-only behaviour.
+ */
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import type { Logger } from '@openfn/logger';
+
+const CGROUP_ROOT = '/sys/fs/cgroup';
```
josephjclark, line 15: Let's take an env var to drive this

packages/engine-multi/src/worker/cgroup.ts
```diff
@@ -0,0 +1,287 @@
...
+// lists the given whitespace-separated token.
+function fileLists(file: string, token: string) {
+  return fs.readFileSync(file, 'utf8').split(/\s+/).includes(token);
+}
+
+// Relocate every process in `parent` into a leader leaf, so `parent` itself
+// becomes empty and its controllers can be delegated. Within a delegated
+// subtree the only member processes are our own (typically just the worker,
+// when it was started inside `parent`)
+function moveProcsToLeader(parent: string, logger: Logger) {
```
josephjclark, line 205: Rename procs to Processes

packages/engine-multi/src/worker/pool.ts
```diff
@@ -13,13 +13,29 @@ import {
...
   env?: Record<string, string>; // default environment for workers
   memoryLimitMb?: number; // --max-old-space-size for child processes
 
+  // Hard memory.max ceiling (mb) applied to each child via a cgroup v2 leaf.
+  // Best-effort: ignored on hosts without a writable cgroup v2 hierarchy.
+  cgroupMemoryLimitMb?: number;
+  // Parent cgroup under which per-child leaf cgroups are created. Defaults to
+  // the cgroup this process was started in; overriding it generally requires
+  // root (the kernel checks the common ancestor's cgroup.procs on migration)
+  cgroupParent?: string;
```
josephjclark: Is this really something we need to customize?

packages/engine-multi/src/worker/pool.ts
```diff
@@ -13,13 +13,29 @@ import {
...
 
 export type PoolOptions = {
   capacity?: number; // defaults to 5
   maxWorkers?: number; // alias for capacity. Which is best?
   env?: Record<string, string>; // default environment for workers
   memoryLimitMb?: number; // --max-old-space-size for child processes
 
+  // Hard memory.max ceiling (mb) applied to each child via a cgroup v2 leaf.
+  // Best-effort: ignored on hosts without a writable cgroup v2 hierarchy.
+  cgroupMemoryLimitMb?: number;
```
josephjclark: The cgroup limit should be derived from the memoryLimit, no? Maybe we accept a multiplier or overhead as an argument, but I think this should be automated

packages/engine-multi/src/worker/pool.ts
```diff
@@ -13,13 +13,29 @@ import {
...
+  resolveSelfCgroup,
+} from './cgroup';
 
 export type PoolOptions = {
   capacity?: number; // defaults to 5
   maxWorkers?: number; // alias for capacity. Which is best?
   env?: Record<string, string>; // default environment for workers
   memoryLimitMb?: number; // --max-old-space-size for child processes
 
+  // Hard memory.max ceiling (mb) applied to each child via a cgroup v2 leaf.
```
josephjclark: Maybe add a default `use-cgroup-memory-enforcement`, aliased to `cgroups` and disabled with `--no-cgroup`s.

Now if cgroups re available, they'll be used as a backstop unless explicitly disabled

packages/engine-multi/src/worker/pool.ts
```diff
@@ -107,6 +137,17 @@ function createPool(script: string, options: PoolOptions = {}, logger: Logger) {
...
       logger.debug('pool: Created new child process', child.pid);
       allWorkers[child.pid!] = child;
+
+      // Place the child in its own cgroup with a hard memory ceiling. The leaf
+      // lives for the lifetime of the (reused) child and is removed when it dies.
+      if (cgroupEnabled && child.pid) {
+        cgroups[child.pid] = createChildCgroup(
+          cgroupParent!,
+          child.pid,
+          options.cgroupMemoryLimitMb! * 1024 * 1024,
```
josephjclark: The cgroup limit right now is set when the process is created, and it looks like a created 1 cgroup file per pid. This is fine, and inline with max-old-space-size.

if we later want to configure this limit per run (rather than per worker), we'll have to consider other architectures. But that's a wider story, it's fine, lets just get this working for now

josephjclark:
@doc-han I'd appreciate you spending an hour with this and letting me know what you think. There are pretty good docs (I hope!) in the worker and engine readmes

The outstanding gotcha is permissions. This will only work if the docker image runs with `--privileged` right now (I presume this is the same change we need in your original PR?). I've opened a thread in `deployment` to try and work out a solution: https://github.com/OpenFn/deployment/issues/482

josephjclark:
I'll have to redraft this PR if we go ahead to that its clean (right now is describes the diff from the other cgroup PR, not from main). It looks like we can switch on some GKE settings to allow the cgroups to be writable, which would allow us to set `WORKER_ENABLE_CGROUP_ENFORCEMENT`

josephjclark:
Rebased, versioned, and ready to ship

josephjclark:
We've put this out onto staging and the good news is that runs are getting oomkilled before taking the whole worker down. [For example](https://app.staging.openfn.org/projects/b387f5a4-a3bb-4bb8-8093-f99436d09d92/w/587e8d05-3a20-4968-8968-49d34d66af70?job=fe51c10c-3199-428f-81ef-49819523af68&panel=editor&run=bffa9bb5-6d05-4971-b819-a06bc0d0566e&step=f9cc5d8a-b31d-467d-92ec-790f2c33a2c8)

The bad news is that the OOMkill came from a heap error - from the native node stack. [gcp logs](https://console.cloud.google.com/logs/query;cursorTimestamp=2026-08-06T14:02:28.060346077Z;duration=PT6H;pinnedLogId=2026-08-06T14:02:15.365682924Z%2Fy7r26x467xipjz8n;query=resource.type%3D%22k8s_container%22%0Aresource.labels.namespace_name%3D%22staging%22%0Aresource.labels.project_id%3D%22platform-test-267207%22%0Aresource.labels.cluster_name%3D%22swiss-standard-1%22%0Aresource.labels.container_name%3D%22staging-worker-fast-lane%22%0Aresource.labels.pod_name%3D%22staging-worker-fast-lane-7458d47c7f-wr29z%22%0Aresource.labels.location%3D%22europe-west6-a%22%0Atimestamp%3D%222026-08-06T14:02:28.060280043Z%22%0AinsertId%3D%22zr0w597b1nsozud7%22?project=platform-test-267207)The cgroup enforcement didn't kick in. I _think_ it would have done, but it did not. The recent node update to 24 may be helping us here.

It's possible that the 128mb overhead we've set is too generous.

Not too worried about this but want to keep an eye on it.

PR #1495, opened 2026-08-04
Worker: add metadata to runs

josephjclark:
## Short Description

This PR adds a `meta` object to the global scope of every step in a workflow.

* Two keys: `runId` and `startTime`
* The meta key is immutable and cannot tbe edited in job code
* The meta key is created before the run starts and does not change per step

workflow id is not currently available.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/late-games-deny.md
+++ b/.changeset/late-games-deny.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': minor
+---
+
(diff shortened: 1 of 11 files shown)


josephjclark:
The meta object is mutable I think. What happens if you try and write to it? Should we block this? Is that object shared across steps?

PR #1502, opened 2026-08-14
Fix the flaky worker integration tests

### LINKED ISSUE #1501: Worker integration tests fail at random
### User story

As someone reviewing or merging a PR, I want the worker integration tests to pass or fail based on the code, so that a red build means something.

### Details

The worker integration job fails intermittently, and which of the three Node versions fails moves around from run to run. On PR #1500 we saw it go 22.20 fail / 24.18 pass / 26.5 pass, then the reverse, then back again - with an empty commit between the last two, so the code was identical. The failure looks like `Error: listen EADDRINUSE: address already in use :::2492`, followed by a test file whose tests all sit pending until ava's two minute timeout kills the job.

The cause is that the tests pick a port at random between 2000 and 3000 and start a server on it, without checking whether that number is already in use. `integration-tests/worker/src/init.ts` does this once for the mock Lightning server and again for every worker, and one test file starts a worker a dozen times. With that many draws from a thousand numbers, two of them landing on the same port is likely enough to happen every few runs. When the second server can't bind, the setup hook never finishes, so nothing in that file ever runs.

It's worth noting that three test files also hardcode the same Lightning port (4321). That's harmless today because ava runs the files one at a time, but it's a trap for anyone who later turns on file concurrency.

### Implementation notes

Because the files run serially and each one is its own process, handing out ports from a counter instead of guessing is enough to remove the collision - something like starting at 4400 and incrementing. The random helper is only called in two places, so changing it in one spot covers both. Worth picking a base that doesn't overlap the 3000s, which the server tests already use.

The more thorough option is to bind to port 0 and let the operating system pick a free one, then read back what it assigned. That's a bigger change, because the mock Lightning server has to be listening before a worker can be told where to connect, which means touching the setup in all six test files. Probably not worth it unless we want file concurrency later.

While in there, the duplicated 4321 could be made distinct per file so the trap goes away.

### Release notes

None - test infrastructure only.

### Tests

Running the worker integration suite repeatedly should stop producing the occasional hung file. A quick sanity check is that no two servers in a single run are asked for the same port.

elias-ba:
Makes the worker integration tests pass reliably. Closes #1501.

That job was failing at random - a different Node version each run, and the same tree giving different results across reruns. It turned out to be four separate bugs, all in the tests rather than the worker:

**Ports were guessed.** Servers were started on a random number between 2000 and 3000 with no check that it was free. A run makes around a dozen of those picks, so collisions happened every few runs. That was the `EADDRINUSE :::2492` we kept seeing. Ports now come from a counter.

**The worker wasn't connected yet.** `initWorker` handed the server back before it had connected to Lightning - it fetches the collections version over http first, then opens the socket. Seven tests swap the worker out and queue a run straight afterwards, so the run could sit in the queue with nothing listening, hanging the file until ava's timeout. It now waits for the socket.

**The sigterm test raced its own signal.** It failed if a claim already in flight arrived after we sent the signal, which the worker can't help. It now keys off the worker reporting that it received it.

**A lost listener in the exit reason tests.** The helper every test there goes through listened with a plain `once` and checked the run id inside the callback, so a completion for any other run consumed the listener without resolving - hanging that promise for good. It now uses the mock's `onSocketEvent`, which only releases on an id match.

The four fixes are separate commits, worth reading individually.

## Validation steps

1. The three integration jobs should pass, and keep passing on reruns.
2. `cd integration-tests/worker && pnpm test` locally.

## AI Usage

Please disclose whether you've used AI anywhere in this PR (it's cool, we just want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/integration-tests/worker/src/init.ts
+++ b/integration-tests/worker/src/init.ts
@@ -3,7 +3,7 @@ import crypto from 'node:crypto';
 
 import createLightningServer, { toBase64 } from '@openfn/lightning-mock';
 import createEngine from '@openfn/engine-multi';
-import createWorkerServer from '@openfn/ws-worker';
+import createWorkerServer, { INTERNAL_SOCKET_READY } from '@openfn/ws-worker';
 import { createMockLogger } from '@openfn/logger';
 import createLogger from '@openfn/logger';
 
@@ -65,6 +65,24 @@ export const initWorker = async (
     batchLogs: true,
     ...workerArgs,
   });
+
+  // The server is returned before it has connected to Lightning: it fetches the
+  // collections version over http first and only then opens the socket. Tests
+  // which swap the worker out and queue a run straight afterwards could leave
+  // that run sitting in the queue with nothing listening for it, and the whole
+  // file would hang until ava's timeout. Waiting for the socket closes that.
+  if (!worker.socket) {
+    await new Promise<void>((resolve, reject) => {
+      const timeout = setTimeout(() => {
+        reject(new Error('worker did not connect to lightning within 10s'));
+      }, 10_000);
+
+      worker.events.once(INTERNAL_SOCKET_READY, () => {
+        clearTimeout(timeout);
+        resolve();
+      });
+    });
+  }
 
   return { engine, engineLogger, worker };
 };
--- a/integration-tests/worker/src/init.ts
+++ b/integration-tests/worker/src/init.ts
@@ -10,7 +10,17 @@ import createLogger from '@openfn/logger';
 const debugWorker = process.env.OPENFN_DEBUG_WORKER;
 const debugLightning = process.env.OPENFN_DEBUG_LIGHTNING;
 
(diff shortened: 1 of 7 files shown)


doc-han (approved):
looks just fine

integration-tests/worker/src/init.ts
```diff
@@ -10,7 +10,17 @@ import createLogger from '@openfn/logger';
 const debugWorker = process.env.OPENFN_DEBUG_WORKER;
 const debugLightning = process.env.OPENFN_DEBUG_LIGHTNING;
 
-export const randomPort = () => Math.round(2000 + Math.random() * 1000);
+// Hand out a fresh port for every server we start. This used to pick a random
```
josephjclark: Please remove all these comments - we don't need a history of every change in the code. The incrementing port number is intuitive and doesn't need explaining

elias-ba: Fair point, removed them all. The detail lives in the commit messages instead.

PR #1506, opened 2026-08-25
tests: add lazy-state integration tests

### LINKED ISSUE #840: Add more lazy state tests
We have a set of integration tests under `execute`, which compile and run code without the overhead of the CLI and Worker. It's a really nice way to focus tests on code without worrying about infrastructure.

We don't have many tests at the moment and I'm keen to add more

I'd be really keen to add a while of lazy state tests to the `lazy-state` file. Simple stuff, advanced stuff, even illegal stuff. I'd like this to be a really comprehensive and definitive listing of lazy state usage.

We should start with very basic tests, which could even be copied from the examples on the docsite. Tests should build up methodically and go onto more complicated stuff.

The examples used here could go on to be used to train AI.

Resources:

* [Lazy State Operator](https://docs.openfn.org/documentation/jobs/job-writing-guide#the-lazy-state-operator)
* [Community: Intro to Lazy State](https://community.openfn.org/t/write-better-jobs-with-the-lazy-state-operator/432)

AchrafReyani:
## Short Description

Adds 21 lazy-state (`$`) tests to `integration-tests/execute/test/lazy-state.test.ts`, building up from simple reads to expressions, mapping/iteration, scoping rules and illegal usage.

Fixes #840

## Implementation Details

The new tests are grouped and ordered methodically, following the examples on the [Lazy State Operator docs page](https://docs.openfn.org/documentation/jobs/lazy-state-operator):

- **Basic reads** – top-level value, nested path, array index, optional chaining on a missing path, and `fn($.data)` returning a value from state as the next state.
- **Expressions** – arithmetic, `&&` / `||` / `!`, ternary, and `$` as a dynamic property key (`codes[$.location.country]`).
- **Objects, mapping and iteration** – mapping state into an object argument (the docs' `create('agent', { name: $.patient.name })` shape, via a tiny in-job `upsert` helper that resolves its argument the way `expandReferences` does), `each($.data.items, …)`, reading the current item with `$.data` inside `each()`, and `group($.data.rows, $.keyPath)`.
- **Scoping** – `$` declared as a parameter or a local `const` is left alone; `$` inside a string literal is not converted.
- **Illegal usage** – every ❌ example from the docs asserts the `LazyStateError` message: assigning `$` to a variable, using it inside a nullary arrow, inside an arrow whose parameter isn't named `state`, and writing to `$` at the top level or inside an operation.

No source changes; only the test file is touched. No changeset added since `@openfn/integration-tests-execute` is private (same as previous test-only changes here).

One thing I noticed while writing these and deliberately left out: a method call on a *nested* state value, e.g. `fnIf($.name.toUpperCase() === 'JOHN', …)`, currently compiles to `(state => state.name.toUpperCase)() === 'JOHN'` – the arrow wraps the callee rather than the call (the special case in `ensureParentArrow` only handles `$.method()` directly on state). Happy to open a separate issue for that if useful.

## QA Notes

Ran in a clean `node:22` Linux container (fresh `pnpm install --frozen-lockfile` + `pnpm -r --filter=./packages/* run build`):

```
integration-tests/execute $ pnpm ava test/lazy-state.test.ts --verbose
  ✔ … (23 tests)
  23 tests passed

integration-tests/execute $ pnpm ava
  35 tests passed
  1 test skipped
```

Also `tsc --noEmit -p integration-tests/execute/tsconfig.json` and `prettier --check` on the file pass.

(Side note, not addressed here: on Windows the whole `execute` integration suite fails before any test runs because the runtime linker builds `file://C:\…` instead of a `file:///C:/…` URL; CI on Linux is unaffected.)

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/integration-tests/execute/test/lazy-state.test.ts
+++ b/integration-tests/execute/test/lazy-state.test.ts
@@ -1,6 +1,15 @@
 import test from 'ava';
 import execute from '../src/execute';
 
+const helpers = `
+const resolve = (state, value) => typeof value === 'function' ? value(state) : value;
+const upsert = (name, obj) => state => {
+  state.upserts ??= [];
+  state.upserts.push({ name, ...resolve(state, obj) });
+  return state;
+};
+`;
+
 test.serial(
   'use lazy-state in template literal & property access',
   async (t) => {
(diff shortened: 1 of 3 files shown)


integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
 
   t.deepEqual(result, { data: { name: 'John', greetings: 'Hello John' } });
 });
+
+// Note: language-common's fn() passes the resolved argument through as the
+// next state, so fn($.x) can be used to return state.x directly.
+// This helper is a minimal "adaptor operation" which resolves a lazy-state
+// argument the same way real adaptor operations do (via expandReferences).
+const helpers = `
```
josephjclark: can we put this helper at the top of the file?

I don't think the comment is neccessary

integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
...
+};
+`;
+
+/*
+ * Basic reads
+ */
+
+test.serial('read a top-level boolean in a condition', async (t) => {
+  const state = { ready: true };
+  const job = `fnIf($.ready, fn(state => {
```
josephjclark: common does actually have an `assert` function: https://docs.openfn.org/adaptors/packages/common-docs#assert

That was probably added after these execute tests were first created, so you might have to update the local adaptor, but these tests would be way neater if they used `assert`

The error message is optional and lazy state is supported in the condition. I just tested this on the app:
```
fn(() => ({
  x: 22
}))

assert($.x === 22)
```

integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
...
+  const job = `fnIf($.data.missing?.id, fn(state => {
+  state.ran = true;
+  return state;
+}))`;
+
+  const result = await execute(job, state);
+  t.is(result.ran, undefined);
+});
+
+test.serial('return a value read from state as the next state', async (t) => {
```
josephjclark: eh this isn't really a lazy state test, it's an fn test. From a lazy state perspective it's exactly the same as `read a top level boolean`.

integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
...
+  state.upserts.push({ name, ...resolve(state, obj) });
+  return state;
+};
+`;
+
+/*
+ * Basic reads
+ */
+
+test.serial('read a top-level boolean in a condition', async (t) => {
```
josephjclark: The top level boolean thing is immaterial here. We're not testing top level numbers and strings etc - and nor should we.

We should rename this test to something like `read a top level path`.

Because it's the evaluation of the expression that's important, not the value of the property.

integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
...
+
+test.serial('return a value read from state as the next state', async (t) => {
+  const state = { data: { x: 1 }, other: 'stuff' };
+  const job = `fn($.data)`;
+
+  const result = await execute(job, state);
+  t.deepEqual(result, { x: 1 });
+});
+
+/*
```
josephjclark: I would remove these labels because the previous tests are expression tests too. The labels don't add any value.

integration-tests/execute/test/lazy-state.test.ts
```diff
@@ -49,3 +49,280 @@ fn($.callMeMaybe($.data.name))`;
...
+  const job = `fnIf($.report.revenue - $.report.expenses === 60, fn(state => {
+  state.profit = true;
+  return state;
+}))`;
+
+  const result = await execute(job, state);
+  t.true(result.profit);
+});
+
+test.serial('use lazy-state in logical expressions', async (t) => {
```
josephjclark, line 104: this is a great little test!

josephjclark (commented):
This is really good, thank you @AchrafReyani 

I've made some suggestions if  you have time to make them. If not I'm happy enough to merge as-is.

Thanks!

integration-tests/execute/test/lazy-state.test.ts
AchrafReyani: Done — moved to the top of the file and dropped the comment.

integration-tests/execute/test/lazy-state.test.ts
AchrafReyani: Thanks, that's much neater. Switched the read/expression tests to `assert(...)` and bumped `@openfn/language-common` to 3.3.4 in the execute package (lockfile updated). One note: a failed `assert` doesn't reject the promise here — the runtime puts the AdaptorError on `state.errors` — so the tests check `result.errors` rather than expecting a throw, and the first test also covers the failing case.

integration-tests/execute/test/lazy-state.test.ts
AchrafReyani: Fair point — removed.

integration-tests/execute/test/lazy-state.test.ts
AchrafReyani: Renamed to `read a top level path`.

integration-tests/execute/test/lazy-state.test.ts
AchrafReyani: Removed the section labels.

josephjclark (approved):
Thank you @AchrafReyani !

It's a shame the asserts aren't quite as neat as I hoped, but I still think it's a nicer pattern than before.

Thanks for taking a second look at this!

PR #1507, opened 2026-08-26
Worker: Fix dataclip serialization

### LINKED ISSUE #1504: Large dataclips can still cause runs to be lost
Despite a lot of improvements to memory handling and payload processing, I think we're still losing lost runs with large dataclips.

I'm looking at [impact water](https://app.openfn.org/projects/8b67c0ed-40a2-4537-9a02-b8da905e704b/runs/a387e375-138a-4781-a526-992fa6fe097c) which is regularly building 9.8mb dataclips and lost runs.

I'm not sure sure exactly where the error lies - but I think when processing this 10mb dataclip to send to lightning, either the connection is actually timing out or some process (lightning side or worker side) is exceeded a memory limit and giving in.

The worst thing is that this is dying silently and causing the websocket to disconnect, resulting in a lost run

A few things jump to mind:
- Claude points out the state objects are double encoded in to JSOn strings. bloating the size A 9mb dataclip in memory might acutally serialize to 10mb or more in the event sent to lightning because of the nested JSON escape characters
- We are not getting good sentry reports when these events timeout - that needs fixing

This is the underlying cause behind https://github.com/OpenFn/kit/issues/1072 and https://github.com/OpenFn/kit/issues/1179. Both those issues describe symptoms of this one underlying thing. They both describe a case where an event fails to send and then bad things happen. This issue is the reason teh events fail.

josephjclark:
## Short Description

This PR adds an option to remove the stringifying of dataclips before they're sent to Lightning.

I'm certain this was added in error in the very first implementation and never looked at again.

But it's causing dataclips send to lightning to bloat over the allowed frame limit, causing large dataclips to trigger lost runs.

See matching Lightning PR for more details: https://github.com/OpenFn/lightning/pull/5098
A one or two-sentence description of what this PR does.

Fixes https://github.com/OpenFn/kit/issues/1504

This is branched on top of some general diagnostics and sentry improvements.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/tame-hats-drum.md
+++ b/.changeset/tame-hats-drum.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Capture more diagnostic detail when the connection to Lightning drops unexpectedly
--- a/.changeset/tidy-plums-obey.md
+++ b/.changeset/tidy-plums-obey.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Reduce dataclip bloat when sending step results to Lightning
--- a/packages/engine-multi/src/worker/thread/runtime.ts
+++ b/packages/engine-multi/src/worker/thread/runtime.ts
@@ -7,7 +7,6 @@ import {
   type WorkerEvents,
 } from '../events';
 import ensurePayloadSize from '../../util/ensure-payload-size';
-import type { ExternalEvent } from '../../events';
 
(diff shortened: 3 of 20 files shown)


josephjclark:
I'd like to confirm lightning  version before merging

josephjclark:
Looks good. This produces correct dataclip with:
- this branch and main lightning
- this branch and lightning branch with default settings (double encoding) - works great
- this branch and lightning branch with encoding off (the fix) - works great

PR #1511, opened 2026-08-30
CLI: support custom webhook paths in project.yaml

### LINKED ISSUE #1516: Support custom webhook paths in project.yaml
### User story

As someone deploying the same workflows across many facilities, I want to name a webhook's URL in `project.yaml`, so I know every endpoint before I deploy anything rather than opening each workflow afterwards to copy a generated id.

### Details

Lightning is gaining a custom path on webhook triggers in OpenFn/lightning#5105. A trigger named `facility-001` in a project answers at `/i/<project-id>/facility-001` as well as its usual `/i/<trigger-id>` URL.

The CLI cannot send that yet. Deploy builds what it sends for a trigger from a fixed list of fields, and the path is not on it, so a name written in `project.yaml` is quietly dropped on the way up. Pull is fine, since it takes the yaml straight from Lightning.

There is a second half to this. The CLI decides which workflows to send by comparing a fingerprint of each one, and that fingerprint ignores the path. So a workflow whose only change is its endpoint name looks unchanged and gets skipped. Lightning builds the same fingerprint the same way, so both sides have to start counting the path in the same release or they stop agreeing.

### Implementation notes

A path should only travel when the spec mentions it. Leaving the key out has to mean "keep whatever the server has", or deploying a spec written before any of this existed would wipe a name somebody set in the app. Writing it blank clears it.

Worth being careful with the fingerprint. Adding a field to it can move every workflow's version, which would make everything look changed at once. It does not have to: a workflow with no path can keep hashing exactly as it does today.

### Release notes

You can now name a webhook's URL in `project.yaml`, so its address is known before you deploy.

### Tests

Round trip a project with a named webhook: deploy it, post to the named URL, pull it back and check the name comes down. Deploy a spec that never mentions the path and check nothing is wiped. Check a blank value clears it. Check a workflow whose only change is its path is actually sent.

elias-ba:
## Short Description

Adds `custom_path` to the webhook trigger payload `@openfn/deploy` sends, so a webhook endpoint can be named in `project.yaml`. This is the CLI half of OpenFn/lightning#5105.

Fixes #1516

## Implementation Details

`@openfn/deploy` never sent `custom_path`, so a name written in `project.yaml` was dropped on the way up. It carries it now.

A path only travels when the spec mentions it. Leaving the key out means the server keeps whatever it has, so deploying a spec written before this existed will not wipe a path somebody set in the app. Writing it blank clears it, either as `''` or as a bare `custom_path:`, which YAML reads as null. That is the rule the provisioner applies on the Lightning side.

Two other things had to move or the feature does not work. The workflow version hash ignored the path, so a workflow whose only change was its endpoint name looked unchanged and `openfn project deploy` skipped it; both sides hash it now, and a workflow without a path hashes exactly as it did. And `deploy` showed a removal for every trigger field the payload does not carry, which it was never going to remove.

The v2 project format swept a pulled path under `openfn:` instead of keeping it beside `webhook_reply`, so the two formats disagreed about where a webhook's path lives.

## QA Notes

Nothing here works until OpenFn/lightning#5105 ships. An older CLI against a newer Lightning is fine, and a newer CLI against an older Lightning sends a key the server ignores.

Once both are up: put `custom_path: facility-001` on a webhook trigger in `project.yaml`, run `openfn deploy`, and `POST /i/<project-id>/facility-001` should create a work order. Then `openfn pull` and the path should come back down.

## AI Usage

Please disclose whether you've used AI anywhere in this PR (it's cool, we just want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


packages/deploy/src/stateTransform.ts
```diff
@@ -205,6 +213,16 @@ function mergeTriggers(
           enabled: pickValue(specTrigger!, stateTrigger!, 'enabled', true),
         };
 
+        // Only when the spec says something. An absent key means "leave it
+        // alone", so deploying a spec written before custom paths existed does
+        // not wipe a path set through the app. A blank one clears it.
+        if (
+          trigger.type === 'webhook' &&
+          specTrigger!.custom_path !== undefined
```
doc-han:
```suggestion
          specTrigger && specTrigger.custom_path !== undefined
```

This prevents use from the error `trying to access key custom_path from undefined`

packages/deploy/src/stateTransform.ts
```diff
@@ -205,6 +213,16 @@ function mergeTriggers(
...
+        ) {
+          trigger.custom_path = specTrigger!.custom_path;
```
doc-han: If we add the suggestion above. we wouldn't need `!` on `specTrigger`

doc-han (approved):

packages/deploy/src/stateTransform.ts
elias-ba: Done, thanks. Left the same assertion on the lines below, they were there before this and I didn't want to widen the diff.

packages/deploy/src/stateTransform.ts
elias-ba: Gone with the same change.

doc-han (changes requested):
The changeset looks weird to me. I mostly expect a single one if it's all patch or all minor. I see there's a bit of a mix here. but I don't get how a single PR has a patch and at the same time a minor for deploy.

elias-ba (commented):
@doc-han good catch, thanks. Deploy has one entry now.

josephjclark (changes requested):
Please bump versions before merging to main

elias-ba (commented):
@josephjclark versions are bumped, and the changelog notes are much shorter now.

josephjclark (commented):
Added a new test for v2 deploy, otherwise looks good.

I wanna hold release by a couple of hours while I look into somehing for Tuch

PR #1520, opened 2026-09-01
Support collections in v2 deploy

### LINKED ISSUE #1418: Sync: adding collections and channels locally
How do we add collections and channels to a local project with v2 sync?

Collections and channels both live at the project level

We don't really represent the project level stuff on disk. You've got your workflow.yaml but that doesn't hold collections or channels

The project yaml file contains both keys at the top level. But you can't edit that locally (when you try and  deploy it, we fetch the latest version and drop your local edits)

Probably the only place we currently do hold this is the `project` key in `openfn.yaml`. We can pull those details into the metadata, and users can edit it freely to add more.

That's got to be the answer, but it does slightly change my perception on what the openfn.yaml file is

josephjclark:
First pass at adding collections metadata to openfn.yaml. Seems like the only place to track project-wide metadata like this.

Fixes #1418 

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
This deliberately doesn't do channels because I think it's a little more complex and no-one has asked yet. I might sneak it in - just want to understand the impact of collections first.

.claude/plans/1418-project-collections.md
```diff
@@ -0,0 +1,250 @@
+# Implementation plan: editable collections in openfn.yaml
```
josephjclark: TODO remove this

josephjclark:
This has ended up being complex enough so I am absolutely not going to do Channels too. I'll close the issue and create a new one.

josephjclark:
Ready to go here

PR #1521, opened 2026-09-02
Worker: don't hang when fetch:plan replies with an error

stuartc:
## Short Description

Fixes a bug where the worker would hang forever (and quietly leak a capacity slot) if Lightning rejected a `fetch:plan` request after a run had already been claimed.

## Implementation Details

Lightning is going to start being able to reply with an error to `fetch:plan` (e.g. when it can't resolve the job's adaptor — `adaptor_not_ready`, `adaptor_not_found`, etc).

Found a loose unresolved promise when the run channel is joined and `sendEvent` rejects when fetching the plan. Turns out we were leaving it stuck (not catching the rejection). 

What happens is that `joinRunChannel` just hung forever, the upstream try/catch in `server.ts` (which does the clean up) doesn't get the exception, and the worker's slot for that run was occupied forever. The lost runs janitor does mark it as lost, so it's not completely "silent" but it doesn't look like we'd get the slot back.

So wrapping the call in a try/catch makes a failed call do the same as a failed or timed out join on the run channel.

## QA Notes

Theres a test that forces `fetch:plan` to error and confirms `joinRunChannel` rejects instead of hanging.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI

### DIFF
--- a/.changeset/khaki-jars-relax.md
+++ b/.changeset/khaki-jars-relax.md
@@ -1,0 +1,5 @@
+---
+'@openfn/ws-worker': patch
+---
+
+Fail a run cleanly instead of hanging when Lightning rejects the plan fetch after a claim
--- a/packages/ws-worker/src/channels/run.ts
+++ b/packages/ws-worker/src/channels/run.ts
@@ -37,11 +37,16 @@ const joinRunChannel = (
         if (!didReceiveOk) {
           didReceiveOk = true;
           logger.success(`connected to ${channelName}`, e);
-          const run = await sendEvent<GetPlanReply>(
-            { channel, logger, id: runId, options: {} },
-            GET_PLAN
-          );
-          resolve({ channel, run });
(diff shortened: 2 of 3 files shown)


elias-ba (approved):
@josephjclark I looked at this and it looks good to me. Do you wanna have a look too before we merge ? @stuartc and the connections team will need it in prep for their work on the new adaptor registry.

josephjclark:
Taking a look

josephjclark (approved):
Yep looks good. Do you want this released now?

stuartc:
> Yep looks good. Do you want this released now?

Ah thanks for the offer, can go out on the next one - or I'll shout. The code that would hit this is still in a branch.

josephjclark:
This is merged in https://github.com/OpenFn/kit/pull/1532 and will be released soon

Issue #1523, opened 2026-09-03
New projects are not given a unique alias


josephjclark:
When you do `project deploy --new --name`, we do not seem to create a new alias. So basically the new project will overwrite the original oroject

issue as bookmark to investigate

- checkout a project (make sure it's on the default alias)
- re-deploy it `openfn deploy --new --name copy`
- your checked-out state is now the new copy (is this right? idk tbh - probably?)
- there is only one local project@domain.yaml file and it tracks the clone, not the original

This isn't a problem when deploying to a different endpoint: but if posting to the same domain it's not good.

PR #1533, opened 2026-09-08
CLI: checkout direct from project.yaml

### LINKED ISSUE #1264: Deploy: allow deploy straight from project.yaml

josephjclark:
## Short Description

This PR enables users to run a deploy from a file: either an exported project.yaml (from the app) or from A single v2 state file (the thing created when fetching a project).

```
openfn project deploy .projects/main@app.openfn.org.yaml --name my-new-project
```

Note that if you try and deploy a project to the same instance and the same UUID, it'll just override. In this case you have to pass `--new` or set a new name.

A side-effect of this work is that you can now edit a local project.yaml  file and deploy it directly to lightning - previously the CLI didn't allow this because a deploy would re-build the project.yaml from the local filesystem. Being able to modify the yaml directly is probably a useful escape hatch. But one I'll leave undocumented for now.

Fixes #1264

Bonus fix for https://github.com/OpenFn/kit/issues/1523

## QA Notes

To test this ( I recommend running against local or at least staging)

You can set `OPENFN_ENDPOINT` and `OPENFN_API_KEY` env vars to configure deploy targets and access and stuff. `.env` files work.

### Scenario 1 - deploy an exported project

- Go to any project in the app and export it as yaml
- copy the path to the file
- With a superuser API token, deploy that file to an instance: `openfn project deploy downlaoded-project.yaml --name my-duplicate`
- The project should exist on the target server
- Should work with whatever name you pass, or no name
- Should work on the same instance as you exported from, or a different one
- Should work with a new project, or an existing one

### Scenario 2 - re-deploy an existing project

- Fetch a project: `openfn project fetch <uuid> --alias og`
- Note that you'll now have a file like  `og@app.openfn.org.yaml`
- Now, with a superuser API token, deploy that file back: `openfn project deploy ./projects/og@app.openfn.org.yaml --name my-duplicate`
- Should work with whatever name you pass, or no name
- Should work on the same instance as you exported from, or a different one
- - Should work with a new project, or an existing one

## Test Cases

This PR has crept into some sensitive areas so I want to run a really thorough suite of QA against staging:

- [x] Do a regular change and deploy
- [x] deploy checked out as new 
- [x] deploy checked as an update to an existing project
- [ ] ~export a project spec (v1 and v2?) and deploy as new~
- [ ] ~export a project spec (v1 and v2?) and deploy to existing~
- [x] Take a project state and re-deploy as new
- [x] Take a project state and re-deploy to existing

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [x] I have not used AI

### DIFF
--- a/.changeset/fast-boats-grin.md
+++ b/.changeset/fast-boats-grin.md
@@ -1,0 +1,5 @@
+---
+'@openfn/project': patch
+---
+
+Fix an issue where serializing a project without credential uuids to app state causes credentials on job bodies to be lost
--- a/.changeset/fast-cobras-throw.md
+++ b/.changeset/fast-cobras-throw.md
@@ -1,0 +1,5 @@
+---
+'@openfn/lexicon': patch
+---
+
+Project credentials can optionally have a UUID
--- a/.changeset/shaky-papayas-change.md
+++ b/.changeset/shaky-papayas-change.md
@@ -1,0 +1,5 @@
+---
+'@openfn/project': patch
+---
+
+Enable projects loaded from state files to be stateless
--- a/.changeset/slow-seals-type.md
+++ b/.changeset/slow-seals-type.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Ensure that new projects don't overwrite existing aliases
--- a/.changeset/tidy-hoops-jam.md
+++ b/.changeset/tidy-hoops-jam.md
@@ -1,0 +1,6 @@
+---
+'@openfn/project': patch
+'@openfn/cli': patch
+---
+
+Fix an issue where --name doens't apply to a project loaded from file
--- a/.changeset/young-brooms-drum.md
+++ b/.changeset/young-brooms-drum.md
@@ -1,0 +1,5 @@
+---
+'@openfn/cli': patch
+---
+
+Enable a v2 project yaml file to be deployed directly, eg, `openfn project deploy .projects/main@app.openfn.org.yaml
--- a/integration-tests/cli/test/deploy.v2.test.ts
+++ b/integration-tests/cli/test/deploy.v2.test.ts
@@ -427,6 +427,91 @@ test.serial('warn when local and remote workflows have diverged', async (t) => {
 });
 
 test.serial(
+  'deploy a pulled v2 state file as a new project',
+  async (t) => {
+    const projectId = 'iiiiiiii';
+    server.addProject(makeProject(projectId) as any);
+
+    const before = Object.keys(server.state.projects);
+
+    // pull with an alias, producing a fetched v2 state file that has a uuid
+    const pullResult = await run(
+      `openfn project pull ${projectId} --alias og --log-json -l debug`
+    );
+    t.falsy(pullResult.stderr);
+
+    const pulledPath = path.join(tmpDir, '.projects', 'og@localhost.yaml');
+
+    // deploy that exact file back as a duplicate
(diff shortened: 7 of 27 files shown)


josephjclark:
ok, this is merged and caught up. Just need the integration tests and a round of manual testing and I think we're good.

josephjclark:
It looks like if a workflow is removed, and you try to deploy to a different project, we get a "nothing to deploy" error. That might need raising as a new bug because it's too far removed from this work

Ie:
- Checkout a project
- Post it as new 
- Checkout the OG
- Delete a workflow
- Post it to the new `openfn project deploy new`
- should be told nothing has changed

btw deleting a workflow on the OG project DID work, so not sure why this isn't stable

josephjclark:
Something is wrong when cross posting a new project straight from state:

```
 pnpm openfn project deploy .projects/jam3@localhost.yaml --apikey $LOCAL_LIGHTNING_SUPER_PAT --new --name jam4 --endpoint http://localhost:4000
```

Looks like some uuid wasn't stripped

```
Request: POST /api/provision/
** (exit) an exception was raised:
    ** (Ecto.ConstraintError) constraint error when attempting to insert struct:

    * "workflows_pkey" (unique_constraint)

If you would like to stop this constraint violation from raising an
exception and instead add it as an error to your changeset, please
call `unique_constraint/3` on your changeset with the constraint
`:name` as an option.

The changeset defined the following constraints:

    * "workflows_name_project_id_index" (unique_constraint)
    * "workflows_project_id_fkey" (foreign_key_constraint)
```


Edit: yes, I can see that the state wasn't converted down to a spec. All the old UUIDs are there. Wasn't this the whole fix on this branch?

josephjclark:
Ok I think this might finally be done- it raised some really tricky questions which I hope I've navigated reasonably well.

Tomorrow I'll re-do the manual test suite again before merging

josephjclark:
Just hit some drama trying to deploy an exported project.yaml:

1) "extraneous parameters:  credentials". The provisioner is rejecting any credentials being uploaded. That doens't normally happen does it? Oh - the `credential` key needed to be converted into `project_credential`. I won't why that didn't happen?

2) One edge is like `x->x` - it's an edge pointing to itself. That doesn't exist in the base project yaml. How did that get created?


The more changes I have to make on this PR the more nervous I get. It's been way, way more work than I antipicated.


EDIT: Hang on hang on, that's because the exported project is using the v1 format. All bets are off.

That's because the portability PR hasn't been merged in lightning. And is full of conflicts. Damn.  Now I have a conundrum: do I fix support for v1 exports (which tbh shoulldn't be super hard) or do I merge and run? I think at this stage I'll raise an issue, merge and run.

Opened as https://github.com/OpenFn/kit/issues/1540. I'll skip the export testing for now. I might come back.

josephjclark:
Yet another problem. If you try and deploy a project file directly without `--new`, the cli will try and fetch the latest version. Which means if you edit the project file locally and try to deploy, changes will be overwritten.

We need to disable that extra fetch in load-from-file mode.

josephjclark:
Ok I think this is ready - I'm going to merge it to the release and give the release a test on Monday.

I'm thinking about some quick fixes for credentials too, because if you have a credential this basically completely breaks

PR #1534, opened 2026-09-08
fix(ws-worker): suppress credential environment mismatch from sentry

### LINKED ISSUE #1460: Worker: ignore sentry errors `Credential environment mismatch`
We're seeing another user error on sentry:

```
[fetch:credential] Credential environment mismatch.
This project is using 'geda' but credential 'Pius Git Token' does not have a matching environment. To fix:
  1. Go to https://app.openfn.org/projects/a36ac16e-3b46-4bfa-a9b8-4ad5975bafb0/settings#credentials
  2. Find 'Pius Git Token'
  3. Click 'Edit' and add an environment named 'geda'
Or change your project environment to match an existing credential environment.
```

This error is thrown in a projeect when you select a credential that does not have a value for a given environment. For example, you have a sandbox in `dev` but your credential only has a `main`

This is a perfectly OK and expected user error. It should be supressed from sentry. See #1440 for a reference fix

[example error](https://openfn.sentry.io/issues/6512942563/?environment=prod&project=4508936084848640&query=is%3Aunresolved&referrer=issue-stream)

CaptainAni187:
Closes #1460

Adds `Credential environment mismatch` to `IGNORED_ERROR_PATTERNS`, following the `OAuth token has expired` entry added in #1440.

```js
export const IGNORED_ERROR_PATTERNS: IgnoredError[] = [
  { pattern: /OAuth token has expired/i, severity: 'crash' },
  { pattern: /Credential environment mismatch/i },
];
```

## On the missing severity

The OAuth entry carries `severity: 'crash'`, which `onRunError` uses to override the exit reason. I deliberately left it off here.

The issue asks for this to be suppressed from sentry, and `matchesIgnoredError` alone does that — `getIgnoredErrorSeverity` is a separate lookup, and `onRunError` only overrides when it returns something. So omitting it keeps the change to exactly what was asked and leaves the reason Lightning receives untouched. If you do want an override, say which reason and it's a one-line addition; I did not want to change what the user sees reported without being asked.

## On the wider point in the thread

@josephjclark noted that the more important problem is differentiating these in sentry, since `LightningSocketError` is generic and sentry lumps everything together. That is a bigger change than this and I have not attempted it — this PR only stops this particular message from being reported.

## Tests

`src/util/ignored-errors.ts` had no tests, so this adds the first ones: both patterns, the case-insensitive match, an unrelated message, empty and nullish input, and the severity lookup for both entries — including that a credential mismatch returns no override, which pins the decision above.

- `test/util`: **75 passing**, up from 67. The 4 failures, 1 skip and 3 uncaught exceptions in that directory are identical before and after this change.
- `tsc --noEmit`: output identical before and after (the existing errors are unbuilt workspace deps and some `any` parameters in `wake-up.test.ts`).
- Reverting only the new pattern fails exactly the two credential tests.

### DIFF
--- a/packages/ws-worker/src/util/ignored-errors.ts
+++ b/packages/ws-worker/src/util/ignored-errors.ts
@@ -8,6 +8,10 @@ type IgnoredError = {
 // list of errors here!
 export const IGNORED_ERROR_PATTERNS: IgnoredError[] = [
   { pattern: /OAuth token has expired/i, severity: 'crash' },
+  // Raised when a project's environment has no matching credential
+  // environment. A configuration mistake for the user to fix, not a worker
+  // fault, and the message already tells them how to fix it.
+  { pattern: /Credential environment mismatch/i },
 ];
 
 const findIgnoredError = (message?: string | null) => {
--- a/packages/ws-worker/test/util/ignored-errors.test.ts
+++ b/packages/ws-worker/test/util/ignored-errors.test.ts
@@ -1,0 +1,43 @@
+import test from 'ava';
+
+import {
+  getIgnoredErrorSeverity,
+  matchesIgnoredError,
+} from '../../src/util/ignored-errors';
+
+test('matches an expired OAuth token', (t) => {
+  t.true(matchesIgnoredError('OAuth token has expired'));
+});
+
+test('matches a credential environment mismatch', (t) => {
+  const message = `[fetch:credential] Credential environment mismatch.
+This project is using 'geda' but credential 'Pius Git Token' does not have a matching environment.`;
+
+  t.true(matchesIgnoredError(message));
+});
+
+test('matches regardless of case', (t) => {
+  t.true(matchesIgnoredError('credential ENVIRONMENT Mismatch'));
+});
+
+test('does not match an unrelated error', (t) => {
+  t.false(matchesIgnoredError('Something else went wrong'));
+});
+
+test('does not match empty or nullish input', (t) => {
+  t.false(matchesIgnoredError(''));
+  t.false(matchesIgnoredError(null));
+  t.false(matchesIgnoredError(undefined));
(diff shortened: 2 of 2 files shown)


CaptainAni187:
Note on the red `integration_test` job — it doesn't look related to this change.

The worker integration suite hangs rather than fails:

```
48 tests passed
4 tests skipped
3 tests remained pending after a timeout
Failed
```

Nothing errors; three tests just never resolve and ava times out. That's the same open-handle symptom described in #1321 ("some open loop is keeping the process alive").

It also isn't specific to forks or to these PRs — #1527 and #1530, both branched inside this repo, fail the same job, while #1528, #1529, #1532 and #1533 pass it. Looks intermittent.

Unit tests and type-checks for the packages touched here are green locally, and identical to a clean checkout.

josephjclark:
Thanks for this @CaptainAni187 🎉

I'm a bit torn on the tests. They're not very valuable because they're mostly just asserting on basic regex patterns. It's 43 lines of bloat.

I'm planning a release today so I'll probably just merge this in. I might remove the tests, maybe it's better than nothing 🤔

For the record, pasting test suite results into the PR body is pointless, misleading, potentialy dangerous, and kinda annoying. We have automated test suites which tell is if tests are passing. I'll take those over plain text any day.

Please carefully review AI generated PR summaries and make sure to remove irrelevant or misleading information. Humans have to read this stuff and make decisions based on it.

CaptainAni187:
Both fair. Please drop the tests, they assert the regex engine more than the behaviour and aren't worth 43 lines to carry.

Point taken on the pasted test output too. I'll stop putting it in PR bodies and trim it from my open ones.

josephjclark:
Closing in favour of https://github.com/OpenFn/kit/pull/1550

PR #1535, opened 2026-09-09
Keep the step ids a project already has

elias-ba:
## Short Description

Two steps called `step 👍` and `step 👎` both slugify to `step`, so on pull one overwrites the other and a step's code is gone. Two workflows named that way both become `wf`, and the pull crashes with `step not found`. This keeps the ids a project already has instead of working them out again.

Part of #1388, which asks for this directly: stop using slugify for identity.

## Implementation Details

The file already stores each step's id, and the reader already takes it as written. Only the pull side ignores it. So a pull now keeps the ids the project has, and only works one out for a step it has not seen.

Nobody can hit the step case today because Lightning restricts step names to ASCII, but OpenFn/lightning#5106 lifts that. Workflow names have never been restricted, so that one is reachable now.

Ids are matched by name, since the files carry no uuids, so a rename still moves a step. There is a test for that.

## QA Notes

Name two steps `step 👍` and `step 👎` and pull. You should get two `.js` files with the right code in each. On main you get one.

Same with two workflows. On main the pull throws.

## AI Usage

- [x] I have used Claude Code
- [ ] I have not used AI


packages/project/src/parse/from-app-state.ts
```diff
@@ -77,8 +81,18 @@ export default (
     };
   }
 
-  proj.workflows = Object.values(stateJson.workflows).map((w) =>
-    mapWorkflow(w, proj.credentials)
+  const allWorkflows = Object.values(stateJson.workflows);
+  const workflowIds = resolveWorkflowIds(
+    allWorkflows,
+    recordedIds?.workflows ?? {}
```
josephjclark: Just make the inner signature robust and you don't need all this defaulting. Same above

josephjclark:
But this is circular no?

The step ids that were previously generated came on fetch from slugify. Preserving them is meaningless if they're not safe. And if they don't exist we still slugify them. So this just seems to be pushing the problem around and complicating the code.

Also I  think this is just locking in local IDs (even if step names change).

There should be ways we can smartly and safely slugify to resolve conflict, I'd rather look at that.

I was also hoping for this validation to come from lightning to avoid getting in this situation in the first place. Otherwise Lightning will also have to resolve this if eg we put step names in URLs.

josephjclark:
Closing as the linked issue https://github.com/OpenFn/lightning/issues/5156 is a wontfix

packages/project/src/parse/from-app-state.ts
```diff
@@ -109,22 +123,120 @@ export const mapEdge = (edge: Provisioner.Edge) => {
...
+    }
+    steps[workflow.name] = forWorkflow;
+  }
+
+  return { workflows, steps };
+};
+
+// Both passes run in uuid order so the result does not depend on how the server
+// listed things.
+const assignIds = (
```
josephjclark: There is some value in the unique id generation enabled by this function. I might come back and borrow this later.

Frustrating that the most important and useful function in the PR is undocumented, apart from one comment which might as well be from a random word generator 🤦

PR #1542, opened 2026-09-10
attempt to fix a deploy to an existing project

josephjclark:
Using claude to try and fix deploying from a spec file to an existing project.

```
pnpm openfn project deploy ./.projects/jam3@localhost.yaml  --endpoint http://localhost:4000 --apikey $LOCAL_LIGHTNING_SUPER_PAT --name foo --log debug -f

```

Two problems here: a dubious divergence test (maybe we should ignore divergence when posting cross-domain?),  and then the acutal diff doesn't seem to apply

### DIFF
--- a/packages/cli/src/projects/deploy.ts
+++ b/packages/cli/src/projects/deploy.ts
@@ -176,13 +176,17 @@ export type SyncResult = {
 // This function is responsible for syncing changes in the user's local project
 // with the remote app version
 // It returns a merged state object
+//
+// skipLocallyChangedCheck: merge every workflow in the source, not just those
+// edited since its own last snapshot (that check is meaningless for a file)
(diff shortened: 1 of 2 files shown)


josephjclark:
This was breaking locally but I think actually it's just strugglign to dpeloy to the right place

The real problem here might be that  Idon;t have the target remove available locally. If I fix that I think it'll work

josephjclark:
And just seen a problem where I've fetched a project but it overwrite the main alias. I thought I'd fixed that.

Never ending pain this project, swear to god

josephjclark:
Ok actually my local setup was just broke - it's working fine.

One trick: when we do this, the update isn't written to the right local file (the alias is wrong)

PR #1546, opened 2026-09-11
Checkout project v1

josephjclark:
## Short Description

This PR adds CLI support to deploy a v1 project.

The base PR adds support for project v2 only, on the assumption that the app is standardising on exporting projects in the v2 style.

But that assumption doesn't work:  production still uses v1 formats. So it's probably worth adding this support in, otherwise the feature is kind of usable for one of its major use-cases

## Implementation Details

This is a bigger  PR than I expected because it makes a change that was probably inevitable.

At the moment, there's a buch of condition logic in `to-app-state` used to serialize a statefile as as v1 spec file. This was pretty horrible really.

This PR splits that logic out in to `to-app-spec` and `from-app-spec`.

That's how we support the 1 file on deploy: we load a v1 spec as a Project, and then simply deploy it with the regular means.

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


josephjclark:
Well it seems to work :)

PR #1547, opened 2026-09-11
feat: add lightning contract action

doc-han:
## Short Description

A one or two-sentence description of what this PR does.

Fixes #

## Implementation Details

## QA Notes

## AI Usage

want to know!):

- [ ] I have used Claude Code
- [ ] I have not used AI

## Release branch checklist

If this IS a release branch:

- [ ] Run `pnpm install`

### DIFF
--- a/.github/workflows/lightning-contract.yaml
+++ b/.github/workflows/lightning-contract.yaml
@@ -1,0 +1,21 @@
+name: Lightning contract tests
+on:
+  pull_request:
+    types: [opened, synchronize, labeled]
+jobs:
+  contract:
+    if: contains(github.event.pull_request.labels.*.name, 'run_contract_tests')
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    services:
+      postgres:
+        image: postgres:17
+        env: { POSTGRES_USER: postgres, POSTGRES_PASSWORD: postgres }
+        ports: ["5432:5432"]
+        options: --health-cmd "pg_isready -U postgres" --health-interval 5s --health-retries 20
+    steps:
+      - uses: actions/checkout@v6
+      - uses: OpenFn/kit-lightning-integration@v1
+        with:
+          worker: ${{ github.workspace }}
+          lightning: main


josephjclark:
@doc-han the action warns of a deprecated node version - can we get that fixed before merging?

.github/workflows/lightning-contract.yaml
```diff
@@ -0,0 +1,22 @@
+name: Integration Testing
+on:
+  pull_request:
+    types: [opened, synchronize, labeled]
+jobs:
+  contract:
+    name: test build against lightning
+    if: contains(github.event.pull_request.labels.*.name, 'run_contract_tests')
```
josephjclark: So the tests only run on PRs with the label right now

Worried about that - if users forget to add the label, the tests will never run. In practice that means they'll never run.

I'd rather have tests run on all Ready PRs with a reviewer, so that hopefully we always run integration tests just before merging

Same idea on lightning too

.github/workflows/integration-testing.yaml
```diff
@@ -0,0 +1,22 @@
...
+  pull_request:
+    types: [opened, synchronize, labeled]
+jobs:
+  contract:
+    name: test build against lightning
+    if: contains(github.event.pull_request.labels.*.name, 'run_contract_tests')
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    services:
+      postgres:
```
josephjclark, line 11: can we not default all this stuff? Does it really need to be declared in the action?

.github/workflows/integration-testing.yaml
```diff
@@ -0,0 +1,22 @@
+name: Integration Testing
```
josephjclark, line 1: I'd like to update the filename here too so that it maps to the actual action name. Otherwise it's hard to find

doc-han: resolved

.github/workflows/lightning-contract.yaml
doc-han: resolved. label dependency removed

josephjclark (approved):

.github/workflows/lightning-contract.yaml
josephjclark: Ah but it'll still run on draft PRs and those without review?

PR #1548, opened 2026-09-14
Improvements for credential  sync

### LINKED ISSUE #1537: Allow deploy without credentials
Simple ideas for syncing credentials

```
openfn deploy --strip-creds --map-creds
```

- `--strip-creds` will remove all credentials before deploying
- `--strip-creds a,b,c` will just remove credentials a, b and c
- `-- map creds a@x:b@y` will map the name and owner of credential `a@x` to `b@y`. Maybe optional either side of the `@`

Also, as Frank has previously requested, drop unused creds (maybe allow `--no-prune-creds` to disable this)

josephjclark:
## Short Description

With deploy v2 this PR lets you:
- omit credentials
- force include all credentials
- map credentials new names/owners

Fixes #1537

This also makes a refactor to the core merge code, allowing keys to be removed if not present. I might need to think about this a little more.

To make the feature fit nicely there's a little refactoring of some utils

## Examples

Deploy a project spec with only the credentials it actually uses (this is actually the default)
```
openfn project deploy spec.yaml --creds prune
```

Deploy a project spec, stripping all credentials
```
openfn project deploy spec.yaml --creds none
```

Deploy a project spec, mapping `a@gmail.org:cred` to `service@openfn.org:cred`
```
openfn project deploy spec.yaml --creds cred=service@openfn.org|cred
```

Deploy a project spec, with a credentials.yaml with aliases
```
// credentials.yaml
a@gmail.org|cred:
  alias: service@openfn.org|cred

openfn project deploy spec.yaml -c credentials.yaml
```

Neat!

## AI Usage

want to know!):

- [x] I have used Claude Code
- [ ] I have not used AI


packages/project/test/merge/merge-project.test.ts
```diff
@@ -189,33 +247,6 @@ test('replace mode: source channels override target channels', (t) => {
   t.deepEqual(result.channels, sourceChannels);
 });
 
-test('replace mode: target channels preserved when source has none', (t) => {
```
josephjclark, line 192: This might be a big deal. A change in how we merge projects means that this behaviour is no longer true.

I _think_ this is correct! 

It means if I remove channels locally and push, they'll be removed from the remote.

The prior behaviour does something which looks wierd to me: it says if channels aren't delcared in the source project, go ahead and add the ones in the target. This is inconsistent with other keys.

packages/project/src/util/base-merge.ts
```diff
@@ -12,5 +12,15 @@ export default function baseMerge<T>(
   assigns: Record<PropsOnly<T>, unknown> = {}
 ) {
   const pickedSource = sourceKeys ? pick(source, sourceKeys) : source;
-  return assign({}, target, { ...pickedSource, ...assigns });
+  const overrides: any = { ...pickedSource, ...assigns };
+  const merged: any = assign({}, target, overrides);
+
+  // null in an override (from source or assigns) is an explicit removal
```
josephjclark: This is an importand diff as it allows us to remove keys in the mege (like next, if removing edges, or credentials, which is a bug I've fixed here)

packages/project/src/util/base-merge.ts
```diff
@@ -12,5 +12,15 @@ export default function baseMerge<T>(
...
+  for (const key of Object.keys(overrides)) {
```
josephjclark: Actually if I remove this code (because no-oneis passing in nulls anyway) everything seems to pass, apart from the bogus test

I think this is the AI leading me around the horn

packages/project/test/merge/merge-project.test.ts
josephjclark: Ok: so this is in place because the provisioner API doens't support deletions.

But it's dodgy - the CLI will silently do the wrong thing. I'd prefer it errors.

See https://github.com/OpenFn/kit/issues/1549

josephjclark:
Ok I think this is alright. I'll run some manual tests in the morning

josephjclark:
Ah one thing I noticed: in the workflow yaml a credential is sttyled as `owner:name`. I need to update the CLI to use the same format.

I should probably also ensure that in the credential map passed in, you can pass a full owner:name id. And that if an id is ambiguous we throw. I might factor that out into an issue because I really need to move on

